PR #14, opened 2016-07-28
Add check for invalid domain and throw better exception

nikolaseu:


auth0/src/main/java/com/auth0/android/Auth0.java
```diff
@@ -139,6 +139,9 @@ private String resolveConfiguration(String configurationDomain, String domainUrl
         String url = ensureUrlString(configurationDomain);
         if (configurationDomain == null && domainUrl != null) {
             final HttpUrl domainUri = HttpUrl.parse(domainUrl);
+            if (domainUri == null) {
```
nikolaseu: maybe it's better to check at ensureUrlString? also here the domainUrl was already modified by ensureUrlString called at the constructor

nikolaseu: also we can parse only once and save the HttpUrl directly?

lbalmaceda: Let's use `HttpUrl` for domainUrl and configurationUrl

nikolaseu:
@lbalmaceda updated

lbalmaceda:
LGTM

PR #15, opened 2016-07-29
Add Accept-Language Header

lbalmaceda:
Applies to all AuthenticationAPIClient calls. If the language is not found/available, then it will default to "en".

### DIFF
--- a/auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
+++ b/auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
@@ -38,18 +38,21 @@ import com.squareup.okhttp.HttpUrl;
 import com.squareup.okhttp.OkHttpClient;
 
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 public class RequestFactory {
 
     static final String AUTHORIZATION_HEADER = "Authorization";
     static final String USER_AGENT_HEADER = "User-Agent";
+    static final String ACCEPT_LANGUAGE_HEADER = "Accept-Language";
     static final String CLIENT_INFO_HEADER = Telemetry.HEADER_NAME;
 
     private final HashMap<String, String> headers;
 
     public RequestFactory() {
         headers = new HashMap<>();
+        headers.put(ACCEPT_LANGUAGE_HEADER, getDefaultLanguage());
     }
 
     public RequestFactory(@NonNull String bearerToken) {
@@ -68,7 +71,7 @@ public class RequestFactory {
 
     public AuthenticationRequest authenticationPOST(HttpUrl url, OkHttpClient client, Gson gson) {
         final AuthenticationRequest request = createAuthenticationRequest(url, client, gson, "POST");
-        addMetrics((ParameterizableRequest)request);
+        addMetrics((ParameterizableRequest) request);
         return request;
     }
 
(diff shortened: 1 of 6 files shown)


auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
```diff
@@ -137,4 +140,9 @@ AuthenticationRequest createAuthenticationRequest(HttpUrl url, OkHttpClient clie
     Map<String, String> getHeaders() {
         return headers;
     }
+
+    static String getDefaultLocale() {
+        String language = Locale.getDefault().toString();
+        return language != null ? language : "en";
```
nikolaseu: I think we should check for empty string, according to the docs it says an empty string is returned in case there's no lang or country.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
```diff
@@ -1283,4 +1339,9 @@ public void shouldParseUnauthorizedPKCEError() throws Exception {
         }.getType();
         return gson.fromJson(request.getBody().readUtf8(), mapType);
     }
+
+    private String getDefaultLocale() {
```
nikolaseu: wouldn't it be better if we add a test for the actual function  https://github.com/auth0/Auth0.Android/pull/15/files#diff-f8d8baa1565ca2a3ec49d45e48f3febdR144
and then use that same one?

auth0/src/test/java/com/auth0/android/request/internal/RequestFactoryTest.java
```diff
@@ -58,11 +58,19 @@ public void shouldHaveNonNullHeaders() throws Exception {
     }
 
     @Test
+    public void shouldHaveAcceptLanguageHeader() throws Exception {
```
nikolaseu: is this used?

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
lbalmaceda: If you mean to test `getDefaultLocale()` then that's not possible with Mockito, as the `Locale` class has the _static_ & _final_ modifiers and can't be mocked.

nikolaseu:
LGTM

auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
```diff
@@ -137,4 +140,9 @@ AuthenticationRequest createAuthenticationRequest(HttpUrl url, OkHttpClient clie
     Map<String, String> getHeaders() {
         return headers;
     }
+
+    static String getDefaultLocale() {
+        String language = Locale.getDefault().toString();
+        return !language.isEmpty() ? language : "en";
```
hzalaz: `en` is not a Locale

lbalmaceda: I took it from here https://gist.github.com/jasef/337431c43c3addb2cbd5eb215b376179 but I guess I can add the country. Defaulting to US seems fine, anyway this is an edge case. The phone should have a valid locale.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
lbalmaceda: done !

PR #16, opened 2016-08-03
Remove unlink method from the AuthenticationAPIClient

lbalmaceda:
This method is currently implemented in the `UsersAPIClient` class


nikolaseu:
LGTM

PR #20, opened 2016-08-09
Fix callback URI bug that included double backslash

lbalmaceda:
**beta.4** was generating Callback URIs like `https://domain.auth0.com//android/package.name/callback` (note the `//`)

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
@@ -28,13 +28,14 @@ import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.util.Log;
 
+import com.squareup.okhttp.HttpUrl;
+
 import java.util.HashMap;
 import java.util.Map;
 
 class CallbackHelper {
 
     private static final String TAG = CallbackHelper.class.getSimpleName();
-    private static final String REDIRECT_URI_FORMAT = "%s/android/%s/callback";
     private final String packageName;
 
     public CallbackHelper(@NonNull String packageName) {
@@ -47,9 +48,19 @@ class CallbackHelper {
      * @return the callback URI.
      */
     public String getCallbackURI(@NonNull String domain) {
-        String uri = String.format(REDIRECT_URI_FORMAT, domain, packageName);
-        Log.v(TAG, "The Callback URI is: " + uri);
-        return uri;
+        HttpUrl url = HttpUrl.parse(domain);
+        if (url == null) {
+            Log.e(TAG, "The Domain is invalid and the Callback URI will not be set. You used: " + domain);
+            return null;
+        }
+        url = url.newBuilder()
+                .addPathSegment("android")
+                .addPathSegment(packageName)
+                .addPathSegment("callback")
+                .build();
+
+        Log.v(TAG, "The Callback URI is: " + url);
+        return url.toString();
     }
 
     public Map<String, String> getValuesFromUri(@NonNull Uri uri) {
--- a/auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
+++ b/auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
@@ -24,6 +24,8 @@
 
 package com.auth0.android.request.internal;
 
+import android.support.annotation.VisibleForTesting;
+
 import com.auth0.android.Auth0Exception;
 import com.auth0.android.RequestBodyBuildException;
 import com.auth0.android.authentication.ParameterBuilder;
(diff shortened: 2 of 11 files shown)


auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.java
```diff
@@ -124,4 +176,74 @@ public void shouldPostOnFailure() throws Exception {
...
+
+    @Test
+    public void shouldParseUnsuccessfulInvalidResponse() throws Exception {
+        byte[] invalidBytes = new byte[]{12,23,2,1,23,3,21,3,12};
+        final Response response = createBytesResponse(invalidBytes, 401);
+        response.body().string();    //force a IOException the next time this gets called
+        baseRequest.parseUnsuccessfulResponse(response);
+
+        ArgumentCaptor<String> stringCaptor = ArgumentCaptor.forClass(String.class);
+        verify(errorBuilder).from(stringCaptor.capture(), any(Auth0Exception.class));
```
nikolaseu: the captor is not needed
`verify(errorBuilder).from(eq("Request to https://auth0.com/ failed"), any(Auth0Exception.class));`

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.java
```diff
@@ -124,4 +176,74 @@ public void shouldPostOnFailure() throws Exception {
...
+        verify(errorBuilder).from(stringCaptor.capture(), integerCaptor.capture());
+        assertThat(integerCaptor.getValue(), is(401));
+        assertThat(stringCaptor.getValue(), is("n=ot_a valid json {{]"));
+    }
+
+    @Test
+    public void shouldParseUnsuccessfulInvalidResponse() throws Exception {
+        byte[] invalidBytes = new byte[]{12,23,2,1,23,3,21,3,12};
+        final Response response = createBytesResponse(invalidBytes, 401);
+        response.body().string();    //force a IOException the next time this gets called
```
nikolaseu: this depends on the implementation details of something external ?

auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
```diff
@@ -47,9 +48,19 @@ public CallbackHelper(@NonNull String packageName) {
      * @return the callback URI.
      */
     public String getCallbackURI(@NonNull String domain) {
-        String uri = String.format(REDIRECT_URI_FORMAT, domain, packageName);
-        Log.v(TAG, "The Callback URI is: " + uri);
-        return uri;
+        HttpUrl url = HttpUrl.parse(domain);
+        if (url == null) {
```
nikolaseu: what kind of error will the user get when the request is made with a null callback URI?
maybe throw an exception is better? it will not work anyway, and the error will be easier to notice

nikolaseu:
not part of this review but at https://github.com/auth0/Auth0.Android/pull/20/files#diff-63e038cf526e31f64772e39db9ed4eb4R71
maybe its better to just use `HttpUrl.queryParameterNames()` ?

auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
lbalmaceda: The error message will be generated by the webpage. I don't know what's better, either crash the app when he presses the Social Button or log the error but keep the app running.

lbalmaceda:
@nikolaseu we might but that method receive the `Uri` from the android Intent. We would need to parse it first, and then get either the query values or the fragment values.

auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
```diff
@@ -103,6 +110,11 @@ protected final void postOnFailure(final U error) {
         return errorBuilder;
     }
 
+    @VisibleForTesting
+    BaseCallback<T, U> getCallback() {
```
nikolaseu: we can remove this now I guess

auth0/src/test/java/com/auth0/android/request/internal/AuthenticationErrorBuilderTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    public void shouldCreateFromMessage() throws Exception {
+        final AuthenticationException ex = builder.from("message");
+
+        assertThat(ex.getCause(), is(nullValue()));
+        assertThat(ex.getCode(), is("a0.sdk.internal_error.unknown"));
+        assertThat(ex.getStatusCode(), is(0));
+        assertThat(ex.getDescription(), is(any(String.class)));
```
nikolaseu: `is(equalTo("message"))` might be better so we test that the `.from(message)` actually sets/uses the correct message?
also the other test does this for the `from(string, int)` version

PR #31, opened 2016-10-13
Add connection_scope setter

lbalmaceda:


auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -158,16 +162,29 @@ public Builder withState(@NonNull String state) {
...
             return this;
         }
 
         /**
+         * Give a connection scope for this request.
+         *
+         * @param connectionScope to request.
+         * @return the current builder instance
+         */
+        public Builder withConnectionScope(@NonNull String connectionScope) {
```
hzalaz: Let's use variable arguments or a list and join them when sending them to the server

PR #40, opened 2016-11-23
Add Logging for Requests/Responses and Uri's.

lbalmaceda:
Logging is disabled by default. Usage:

```java
WebAuthProvider.init(account)
    .enableLogging()
    .start(activity, callback);
```

or in the API Clients

```java
AuthenticationAPIClient authClient = new AuthenticationAPIClient(account);
authClient.enableLogging();

UsersAPIClient usersClient = new UsersAPIClient(account);
usersClient.enableLogging();
```

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
+++ b/auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
@@ -531,9 +531,7 @@ public class WebAuthProvider {
     private PKCE createPKCE(String redirectUri) {
         if (pkce == null) {
             final AuthenticationAPIClient client = new AuthenticationAPIClient(account);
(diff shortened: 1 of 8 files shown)


auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -341,6 +353,9 @@ private boolean authorize(@NonNull AuthorizeResult data) {
             Log.w(TAG, "The response didn't contain any of these values: code, state, id_token, access_token, token_type, refresh_token");
             return false;
         }
+        if (loggingEnabled) {
```
hzalaz: I think we can do better than copy-pasting an if

PR #44, opened 2016-11-24
Add userInfo method

lbalmaceda:
Use `/userInfo` endpoint to get the user profile associated with a given Auth0 access token.

Usage:

```java
AuthenticationAPIClient client = new AuthenticationAPIClient(getAccount())
   .userInfo("access_token")
   .start(callback); //Callback receives the Credentials.
```

**Important**
The `getProfileAfter()` method is now using the `userInfo` method.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
@@ -359,7 +359,7 @@ public class AuthenticationAPIClient {
     }
 
     /**
-     * Fetch the user profile associated with the given Auth0 access token.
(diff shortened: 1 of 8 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -327,6 +330,29 @@ public AuthenticationRequest loginWithEmail(@NonNull String email, @NonNull Stri
     }
 
     /**
+     * Fetch the user profile associated with the given Auth0 access token.
```
hzalaz: `Returns the information of the user associated with the given access_token`

Issue #52, opened 2016-12-08
WebAuthProvider's callback URI uses https


collinsmwade:
[After following following this WebAuthProvider guide](https://auth0.com/docs/libraries/auth0-android#implementing-web-based-auth), the browser authenticates and closes, but instead of going back to the app an intent picker is shown, prompting the user to pick whether they'll open the callback in our app or in the browser:
![device-2016-12-08-141135](https://cloud.githubusercontent.com/assets/3690795/21026068/a1fbba02-bd50-11e6-9a44-0f513b73e268.png)

This is an issue because the average user is likely to get confused by this dialog, and if the browser is set as the default handler for the callback this way, there's no easy way for them to reset the default.

 Currently when using a WebAuthProvider, the callback URI is of the form: `https://{AUTH0_DOMAIN}/android/{BUNDLE_NAME}/callback`. 
However, for Android to know to open our app on the first try, the callback URI should be of the form: `{BUNDLE_NAME}://{AUTH0_DOMAIN}/android/{BUNDLE_NAME}/callback`.

Currently, it's working on iOS, with the bundle identifier as the URI protocol.

hzalaz:
> Currently, it's working on iOS, with the bundle identifier as the URI protocol.

It works since it's the only way it could work without using Universal Links, which is hard to setup.

We are considering adding the option to allow to choose the scheme and document better the use of Android AppLinks that will mitigate this issue

lbalmaceda:
Merged. Will be available on the next release.

PR #53, opened 2016-12-14
Using package name as scheme in WebAuth callback

collinsmwade:
This is a fix for https://github.com/auth0/Auth0.Android/issues/52


lbalmaceda:
Hi @collinsmwade. Thanks for the PR! 
We can't merge it as this is a breaking change, and also tests are failing. We're currently working on adding a method to configure the `scheme` value used in the callback Uri. 

Cheers.

PR #56, opened 2016-12-21
Use password-realm grant for /oauth/token endpoint

lbalmaceda:
Login using `/oauth/token` endpoint was sending `grant_type = 'password'` which was wrong. This PR fixes that and also allows the `AuthenticationRequest` to specify a realm value.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.java
@@ -106,7 +106,7 @@ public class ParameterBuilder {
     }
 
     /**
-     * Sets the 'realm' parameter
+     * Sets the 'realm' parameter. A realm identifies the host against which the authentication will be made, and usually helps to know which username and password to use.
      *
      * @param realm name of the realm
      * @return itself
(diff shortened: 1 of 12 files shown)


auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.java
```diff
@@ -104,6 +106,16 @@ public ParameterBuilder setConnection(String connection) {
     }
 
     /**
+     * Sets the 'realm' parameter
```
hzalaz: Let's explain better what the realm is

auth0/src/main/java/com/auth0/android/authentication/request/SignUpRequest.java
```diff
@@ -118,6 +118,13 @@ public SignUpRequest setConnection(String connection) {
         return this;
     }
 
+    @Override
+    public SignUpRequest setRealm(String realm) {
+        signUpRequest.setConnection(realm);
+        authenticationRequest.setRealm(realm);
```
hzalaz: And for connection? also if we are not using the proper grant (e.g. when working with the legacy endpoints) what will happen?. If we quietly ignore them we should at least issue a warning in the logs

PR #57, opened 2016-12-21
Add flag to authenticate with OIDC mode

lbalmaceda:
Make legacy authentication methods use `/oauth/token` endpoint if `client.setOAuth2Preferred(true)` is called. The methods that change it's behavior with the flag are:

* `login(String, String, String)` --> calls `login(String, String)` and sets the connection as realm.
* ~`tokenInfo(String)` --> calls `userInfo(String)`~ 
* `signUp(String, String, String)`  --> replaces the login call with `login(String, String)` and sets the connection as realm.
* `signUp(String, String, String, String)`  --> replaces the login call with `login(String, String)` and sets the connection as realm.

**This PR depends on https://github.com/auth0/Auth0.Android/pull/56** as there was no way to set a `realm` value before that.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
@@ -97,7 +97,7 @@ public class AuthenticationAPIClient {
     private final Gson gson;
     private final com.auth0.android.request.internal.RequestFactory factory;
     private final ErrorBuilder<AuthenticationException> authErrorBuilder;
-    private boolean useOAuth2;
+    private boolean oauth2Preferred;
 
 
     /**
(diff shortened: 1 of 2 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -421,6 +450,10 @@ public AuthenticationRequest loginWithEmail(@NonNull String email, @NonNull Stri
     @SuppressWarnings("WeakerAccess")
     @Deprecated
     public Request<UserProfile, AuthenticationException> tokenInfo(@NonNull String idToken) {
```
hzalaz: This is wrong, userInfo is meant for `access_token`

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -191,6 +212,12 @@ public void setUserAgent(String userAgent) {
      */
     @SuppressWarnings("WeakerAccess")
     public AuthenticationRequest login(@NonNull String usernameOrEmail, @NonNull String password, @NonNull String connection) {
```
hzalaz: What I'd do is to chose the grant type and legacy endpoint based on the pipeline flag. If it's off, ro and connection is used otherwise token and realm

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -223,7 +250,7 @@ public AuthenticationRequest login(@NonNull String usernameOrEmail, @NonNull Str
         Map<String, Object> requestParameters = ParameterBuilder.newBuilder()
```
hzalaz: I'd keep this as it was, using password grant, since realm grant without a realm is pretty useless. We can recommend using the other method and even deprecating this or explaining the default directory restriction of password grant

PR #61, opened 2016-12-28
Separate the WebAuthProvider logic into a new class.

lbalmaceda:


auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+            public void onSuccess(@NonNull Credentials codeCredentials) {
+                callback.onSuccess(mergeCredentials(urlCredentials, codeCredentials));
+            }
+        });
+        return true;
+    }
+
+    //Helper Methods
+
+    private void checkErrorValue(String errorValue) throws AuthenticationException {
```
hzalaz: `assertNoError(String errorValue)`

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        if (errorValue == null) {
+            return;
+        }
+        Log.e(TAG, "Error, access denied. Check that the required Permissions are granted and that the Application has this Connection configured in Auth0 Dashboard.");
+        if (ERROR_VALUE_ACCESS_DENIED.equalsIgnoreCase(errorValue)) {
+            throw new AuthenticationException(ERROR_VALUE_ACCESS_DENIED, "Permissions were not granted. Try again.");
```
hzalaz: Can't remember but if we localize other error messages we should do it with these too

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            throw new AuthenticationException("a0.invalid_configuration", "The application isn't configured properly for the social connection. Please check your Auth0's application configuration");
+        }
+    }
+
+    private void checkStateValue(String receivedState) throws AuthenticationException {
```
hzalaz: `assertValidState(String requestState, String responseState)`

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        String sentState = parameters.get(KEY_STATE);
+        if (receivedState != null && sentState != null && !receivedState.equals(sentState)) {
```
hzalaz: We should only deem the response `state` as valid if it's value matches the one in the request, either if both are `null` or the same `String`

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+            Log.e(TAG, String.format("Received state doesn't match. Received %s but expected %s", receivedState, sentState));
+            throw new AuthenticationException(ERROR_VALUE_ACCESS_DENIED, "The received state is invalid. Try again.");
+        }
+    }
+
+    private void checkNonceValue(String receivedNonce, String idToken) throws AuthenticationException {
```
hzalaz: `assertValidNonce(String requestNonce, String idToken)`

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        String sentNonce = parameters.get(KEY_NONCE);
+        if (parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_ID_TOKEN) && (idToken == null || !hasValidNonce(sentNonce, idToken))) {
```
hzalaz: Same as `state`, to be ok both `null` or the same `String`

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+    private void addClientParameters(Map<String, String> parameters, String redirectUri) {
+        if (account.getTelemetry() != null) {
+            parameters.put(KEY_TELEMETRY, account.getTelemetry().getValue());
+        }
+        parameters.put(KEY_CLIENT_ID, account.getClientId());
+        parameters.put(KEY_REDIRECT_URI, redirectUri);
+    }
+
+    private PKCE createPKCE(String redirectUri) {
+        if (pkce != null) {
```
hzalaz: where does this come from?
why is not set in this method?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    @VisibleForTesting
+    static boolean hasValidNonce(@Nullable String nonce, @NonNull String token) {
+        try {
+            final JWT idToken = new JWT(token);
+            final Claim nonceClaim = idToken.getClaim(KEY_NONCE);
+            return nonce != null && nonceClaim != null && nonce.equals(nonceClaim.asString());
+        } catch (DecodeException e) {
+            return false;
```
hzalaz: Let's add a Log

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        try {
+            checkErrorValue(values.get(KEY_ERROR));
+            checkStateValue(values.get(KEY_STATE));
+            checkNonceValue(values.get(KEY_NONCE), values.get(KEY_ID_TOKEN));
+        } catch (AuthenticationException e) {
+            callback.onFailure(e);
+            return true;
+        }
+
+        Log.d(TAG, "Authenticated using web flow");
```
hzalaz: We can move all this code inside the try-catch to avoid using it for control flow which is a bad practice

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -0,0 +1,295 @@
...
+        final Credentials urlCredentials = new Credentials(values.get(KEY_ID_TOKEN), values.get(KEY_ACCESS_TOKEN), values.get(KEY_TOKEN_TYPE), values.get(KEY_REFRESH_TOKEN));
+        if (!shouldUsePKCE()) {
```
hzalaz: There is little gain here to early exit since is just a method invocation so we can use a simple if-else

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
lbalmaceda: It's just for testing purposes. https://github.com/auth0/Auth0.Android/pull/61/files/c4947e1f4f2cc68686a1479e7f3ec6d8b97f7696#diff-fba4899390d3e4efe09155da3e2a8396R80

PR #68, opened 2017-01-09
Refresh auth using /oauth/token refresh_token grant (OIDC mode)

lbalmaceda:
Use either `/delegation` or `/oauth/token` endpoints depending on the OIDC flag.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
@@ -592,7 +592,9 @@ public class AuthenticationAPIClient {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -592,7 +592,8 @@ public SignUpRequest signUp(@NonNull String email, @NonNull String password, @No
     }
 
     /**
-     * Requests new Credentials using a valid Refresh Token.
+     * Requests new Credentials using a valid Refresh Token. How the new Credentials are requested depends on the {@link Auth0#isOIDCConformant()} flag.
```
hzalaz: Let's explain this a bit better, at least talk about the two endpoints involved.

PR #70, opened 2017-01-24
Log a warning message when using non-OIDC endpoints in OIDC mode

lbalmaceda:
If the account used to create the `AuthenticationAPIClient` has the OIDCConformant flag disabled and the client (dashboard) has the flag enabled, the requests to `/oauth/ro` and `/oauth/access_token` will fail with `invalid_request`. This PR check's that error and logs a warning message so the user can properly setup the android client and authenticate.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
@@ -89,7 +89,7 @@ public class AuthenticationException extends Auth0Exception {
     }
 
     private void warnIfOIDCError() {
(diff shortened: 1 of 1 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
```diff
@@ -85,6 +88,12 @@ public AuthenticationException(Map<String, Object> values) {
         }
     }
 
+    private void warnIfOIDCError() {
+        if ("invalid_request".equals(getCode()) && getDescription().startsWith("OIDC conformant clients cannot use")) {
```
hzalaz: Let's mot ever match description values of errors let's print the error and the recommendation if it's invalid_request

PR #79, opened 2017-02-14
Ensure closing the response body after it was parsed

lbalmaceda:
Body should be closed after we finish parsing the response. There are some methods like `body.string()` that close the body internally, but if the call throws an exception we would need to call `body.close()` manually.

Related: https://github.com/square/okhttp/issues/2311

### DIFF
--- a/auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
+++ b/auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
@@ -131,7 +131,7 @@ abstract class BaseRequest<T, U extends Auth0Exception> implements Parameterizab
             stringPayload = body.string();
             Type mapType = new TypeToken<Map<String, Object>>() {
             }.getType();
-            Map<String, Object> mapPayload = gson.fromJson(body.charStream(), mapType);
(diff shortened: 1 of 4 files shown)


auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
```diff
@@ -136,6 +138,11 @@ protected U parseUnsuccessfulResponse(Response response) {
         } catch (IOException e) {
             final Auth0Exception auth0Exception = new Auth0Exception("Error parsing the server response", e);
             return errorBuilder.from("Request to " + url.toString() + " failed", auth0Exception);
+        } finally {
+            try {
+                body.close();
```
hzalaz: Can body be `null`?

auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
```diff
@@ -136,6 +138,11 @@ protected U parseUnsuccessfulResponse(Response response) {
...
+            } catch (Exception ignored) {
```
hzalaz: Should we only catch `IOException` instead all of them so we know when we fail to proper manage the errors?

PR #83, opened 2017-03-14
Okhttp3

nekdenis:
Updated dependencies for okhttp3


auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -63,7 +65,7 @@ public SimpleRequest(HttpUrl url, OkHttpClient client, Gson gson, String httpMet
     }
 
     @Override
-    public void onResponse(Response response) throws IOException {
+    public void onResponse(Call call, Response response) throws IOException {
```
lbalmaceda: Why do we need the `Call` instance here?

auth0/src/main/java/com/auth0/android/request/internal/VoidRequest.java
```diff
@@ -46,7 +48,7 @@ public VoidRequest(HttpUrl url, OkHttpClient client, Gson gson, String httpMetho
     }
 
     @Override
-    public void onResponse(Response response) throws IOException {
+    public void onResponse(Call call, Response response) throws IOException {
```
lbalmaceda: Why do we need the `Call` instance here?

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.java
```diff
@@ -91,13 +92,14 @@ public void setUp() throws Exception {
...
         baseRequest = new BaseRequest<String, Auth0Exception>(url, client, new Gson(), adapter, errorBuilder, callback, headers, parameterBuilder) {
+
             @Override
             public String execute() throws Auth0Exception {
                 return null;
             }
 
             @Override
-            public void onResponse(Response response) throws IOException {
+            public void onResponse(Call call, Response response) throws IOException {
```
lbalmaceda: Why do we need the `Call` instance here?

auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
```diff
@@ -146,7 +148,7 @@ protected U parseUnsuccessfulResponse(Response response) {
     }
 
     @Override
-    public void onFailure(Request request, IOException e) {
+    public void onFailure(Call call, IOException e) {
```
lbalmaceda: Why do we need the `Call` instance here? If we already know it failed, just pass `Request` as `call.request()`.

nikolaseu: @lbalmaceda it is the new `Callback` interface of okhttp3

lbalmaceda:
Hi @nekdenis 
Thanks for the PR. I have a few concerns:
* How will this affect active `OkHttp2` users? AFAIK having different `Okio` dependencies will break things for them.
* I don't see the `build.gradle` file changed. Does this code compile?

Cheers,

nekdenis:
@lbalmaceda Thanks! you're right I've forgotten to add build.gradle. 
Just pushed with updated libraries. 

>How will this affect active OkHttp2 users?
For me (OkHttp3 user) dependency of okhttp2 is adding more than 2k methods that I don't need.
Oldtimers can use old version Auth0.Android with outdated dependencies

lbalmaceda:
@nekdenis Please rollback the `appcompat-v7` version update as it's breaking our CI script. I'll tweak them later in another PR.

nekdenis:
@lbalmaceda done

nekdenis:
@lbalmaceda any updates when you will add okhttp3 support?

lbalmaceda:
Hi @nekdenis 
CI needs to pass. I've found that the dependencies in the `build.gradle` file are wrong. Please correct the issues and make sure the tests pass by running `./gradlew clean test` before next push.
- okhttp3 should be `com.squareup.okhttp3:okhttp:3.6.0`
- logging-interceptor should be `com.squareup.okhttp3:logging-interceptor:3.6.0`

Keep in mind you might need to correct some imports as the package name had changed from v2 to v3.
Cheers.

chrisjenx:
You might also want to update to 3.8.0 as well.

twaddington:
Any updates on this PR?

donatienthorez:
I wanted to update and make a PR also for this but for now I am struggling with the tests which are failing. You can see the pull request here : https://github.com/getmimo/Auth0.Android/pull/2 .

The test method shouldSendSMSLinkWithCustomConnection in the class AuthenticationAPIClientTest and more precisely line 1415 never get called for some reason so the test makes an endless loop. If someone has time to look into it or help me with that I could make a PR with the last OkHttp version (v3.9.1)

PR #86, opened 2017-04-03
Add method to revoke a refresh_token

lbalmaceda:


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -618,6 +620,43 @@ public SignUpRequest signUp(@NonNull String email, @NonNull String password, @No
...
+     *          public void onFailure(AuthenticationException error) {}
+     *      });
+     * }
+     * </pre>
+     *
+     * @param refreshToken the token to revoke
+     * @return a request to configure and start
+     */
+    @SuppressWarnings("WeakerAccess")
+    public ParameterizableRequest<Void, AuthenticationException> revokeToken(@NonNull String refreshToken) {
```
nikolaseu: return a plain request, nobody should change anything anyway

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
```diff
@@ -1573,6 +1573,47 @@ public void shouldFetchProfileAfterLoginRequest() throws Exception {
...
+        assertThat(body, hasEntry("client_id", CLIENT_ID));
+        assertThat(body, hasEntry("token", "refreshToken"));
+
+        assertThat(callback, hasNoError());
+    }
+
+    @Test
+    public void shouldRevokeTokenSync() throws Exception {
+        Auth0 auth0 = new Auth0(CLIENT_ID, mockAPI.getDomain(), mockAPI.getDomain());
+        auth0.setOIDCConformant(true);
```
nikolaseu: why this test sets this flag and the async one not?

auth0/src/test/java/com/auth0/android/util/AuthenticationAPI.java
```diff
@@ -106,6 +106,12 @@ public AuthenticationAPI willReturnSuccessfulSignUp() {
         return this;
     }
 
+    public AuthenticationAPI willReturnSuccessfulEmptyBody() {
+        String json = "{}";
+        server.enqueue(responseWithJSON(json, 200));
```
nikolaseu: from the docs:

> If the request is valid, the refresh token is revoked and the response is HTTP 200, with an empty response body

empty response body is not the same as an empty object in the body.
just to be sure that when the test passes there are no issues when deserializing the response, let's make this return an actual empty body.

lbalmaceda: Good catch.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
lbalmaceda: I think I did this to prove that the flag wasn't considered here. Anyway, the method only works for OIDC clients so I'm considering throwing if the method is called with this flag disabled. What do you think @nikolaseu ?

lbalmaceda: Disregard my comment.. The oidc flag won't make a change on this endpoint since it should be callable anyway..

PR #89, opened 2017-04-04
Fix WebAuthProvider custom Rule error message parsing

### LINKED ISSUE #87: Custom Rule errors are not available in the exception
When using `WebAuthProvider` to log in and a Rule I've created on my dashboard fails to process the request, the custom error I return it's not available in the `AuthenticationException ` description. 

I would expect that the exception returns `true` when calling `exception.isRuleError()` and also that the `exception.getDescription()` in that case returns the message that I define in my Rule script.

The line where this decision is made: https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java#L165

lbalmaceda:
Support custom Auth0 Rules error messages and construct a proper exception with the `unauthorized` code and the given user's description. You might need to update your Rule's script as the error must be an instance of `UnauthorizedError`. Check [this](https://auth0.com/docs/rules#deny-access-based-on-a-condition) article for more context. 

Fixes #87.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
@@ -34,6 +34,7 @@ class OAuthManager {
     static final String RESPONSE_TYPE_CODE = "code";
 
     private static final String ERROR_VALUE_ACCESS_DENIED = "access_denied";
+    private static final String ERROR_VALUE_UNAUTHORIZED = "unauthorized";
     private static final String METHOD_SHA_256 = "S256";
     private static final String KEY_CODE_CHALLENGE = "code_challenge";
     private static final String KEY_CODE_CHALLENGE_METHOD = "code_challenge_method";
@@ -41,6 +42,7 @@ class OAuthManager {
     private static final String KEY_REDIRECT_URI = "redirect_uri";
     private static final String KEY_TELEMETRY = "auth0Client";
     private static final String KEY_ERROR = "error";
+    private static final String KEY_ERROR_DESCRIPTION = "error_description";
     private static final String KEY_ID_TOKEN = "id_token";
     private static final String KEY_ACCESS_TOKEN = "access_token";
     private static final String KEY_TOKEN_TYPE = "token_type";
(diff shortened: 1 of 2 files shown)


auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -1550,6 +1584,12 @@ private String createHash(@Nullable String idToken, @Nullable String accessToken
             }
             hash = hash.concat("error=" + error);
         }
+        if (errorDescription != null) {
+            if (!hash.endsWith("&")) {
```
nikolaseu: this is working, but I don't think it can ever be `false` (except if the values end with a `&` which is wrong).
all these checks should be `endsWith("#")` just in case all previous values where empty/null.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -154,13 +156,15 @@ public void onSuccess(@NonNull Credentials codeCredentials) {
...
-    private void assertNoError(String errorValue) throws AuthenticationException {
+    private void assertNoError(String errorValue, String errorDescription) throws AuthenticationException {
         if (errorValue == null) {
             return;
         }
         Log.e(TAG, "Error, access denied. Check that the required Permissions are granted and that the Application has this Connection configured in Auth0 Dashboard.");
         if (ERROR_VALUE_ACCESS_DENIED.equalsIgnoreCase(errorValue)) {
             throw new AuthenticationException(ERROR_VALUE_ACCESS_DENIED, "Permissions were not granted. Try again.");
+        } else if (ERROR_VALUE_UNAUTHORIZED.equalsIgnoreCase(errorValue)) {
+            throw new AuthenticationException(ERROR_VALUE_UNAUTHORIZED, errorDescription);
```
nikolaseu: can `errorDescription` be null/empty? in the case of a rule error, this is something we don't control, right? it's user defined.
maybe we can add a generic message "Unknown error" in case it's null, just to avoid a null description?

lbalmaceda: Yes! If the key doesn't exist on the map, then the returned value will be `null`. This case is covered by the behavior of the `AuthenticationException` class. getDescription will [always](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java#L131) return a valid message. Maybe instead of checking for null I can use `TextUtils.isEmptyString(description)` [here](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java#L125) so it checks for null and empty strings. What do you think?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
lbalmaceda: I guess i wanted to do that check but got lost and used the same `&` for comparison. Good catch 👍

PR #95, opened 2017-06-13
Use Chrome Custom Tabs when possible

lbalmaceda:
This PR adds support for **Chrome Custom Tabs** to use with the `WebAuthProvider` class when not launching a WebView. The user API remains the same, but now they need a different configuration and they can skip the call to `WebAuthProvider.resume(intent)` as it's handled internally by the library. This also means that they don't need to capture the `onNewIntent` or `onActivityResult` method calls.

### Proposed configuration changes

1. Remove the custom Intent-Filter that captures the Callback URI from the `AndroidManifest.xml` file.

2a. Add **both** `manifestPlaceholders` to the `app/build.gradle` file so that the library knows which domain and scheme to use. The host is set automatically using the `applicationId` placeholder:

```groovy
apply plugin: 'com.android.application'

android {
    compileSdkVersion 25
    buildToolsVersion "25.0.3"
    defaultConfig {
        applicationId "com.auth0.samples"
        minSdkVersion 15
        targetSdkVersion 25
        manifestPlaceholders = [auth0Domain: "@string/auth0_domain", auth0Scheme: "https"]
    }
//...
}
```

2b. Alternatively, don't set the `manifestPlaceholders` but replace the `RedirectActivity` definition in the `AndroidManifest.xml` file with one that defines the custom Intent-Filter.

```xml
<activity android:name="com.auth0.android.provider.RedirectActivity">
<intent-filter>
                <action android:name="android.intent.action.VIEW" />
                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="android.intent.category.BROWSABLE" />

                <data
                    android:host="@string/auth0_domain"
                    android:pathPrefix="/android/{YOUR_APP_PACKAGE_NAME}/callback"
                    android:scheme="https" />
            </intent-filter>
</activity>
```

> Not setting the placeholders like in (2a) or not overriding the Intent-Filter declaration like in (2b) will result in the library's defined Intent-Filter to not find the expected placeholder values and the code to not compile. 

### Discussion

The trade-off is between setting a default `https` scheme in the library's Intent-Filter (so the user doesn't need to set an extra manifest placeholder) or setting a placeholder that the user must always define. 

If a normal user will use more `https` over a custom scheme, we can let them override **the whole Intent-Filter** in the manifest like in (2b). If a normal user will prefer custom schemes, then we can keep the 2 manifest placeholders proposal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/callback
 Remember to replace `{YOUR_APP_PACKAGE_NAME}` with your actual application's package name.
 
 
-Next, define placeholders for the Auth0 Domain and Scheme which are going to be used internally by the library to register an **intent-filter**. Go to your application's `build.gradle` file and add the `manifestPlaceholders` line as shown below:
+Next, define a placeholder for the Auth0 Domain which is going to be used internally by the library to register an **intent-filter**. Go to your application's `build.gradle` file and add the `manifestPlaceholders` line as shown below:
 
 ```groovy
 apply plugin: 'com.android.application'
@@ -286,16 +286,18 @@ android {
         //...
         
         //---> Add the next line
-        manifestPlaceholders = [auth0Domain: "@string/auth0_domain", auth0Scheme: "https"]
+        manifestPlaceholders = [auth0Domain: "@string/auth0_domain"]
         //<---
     }
     //...
 }
 ```
 
-It's a good practice to define reusable resources like `@string/auth0_domain` but you can also hard code the value. In case you're using a [custom scheme](#a-note-about-app-deep-linking) you must update the `auth0Scheme` property.
+It's a good practice to define reusable resources like `@string/auth0_domain` but you can also hard code the value in the file.
 
-Alternatively you can define your own **intent-filter** for the `RedirectActivity` in the `AndroidManifest.xml` file replacing the one defined by the library. If you choose to do this, the `manifestPlaceholders` don't need to be set. In your manifest inside your application's tag add the `RedirectActivity` declaration:
+Alternatively, you can declare the `RedirectActivity` in the `AndroidManifest.xml` file with your own **intent-filter** so it overrides the library's default. If you do this then the `manifestPlaceholders` don't need to be set. If you choose to use a [custom scheme](#a-note-about-app-deep-linking) you must define your own intent-filter as explained below. 
+
+In your manifest inside your application's tag add the `RedirectActivity` declaration:
 
 ```xml
     <application android:theme="@style/AppTheme">
@@ -322,7 +324,7 @@ Alternatively you can define your own **intent-filter** for the `RedirectActivit
     </application>
 ```
 
-Remember to replace `{YOUR_APP_PACKAGE_NAME}` with your actual application's package name.
+If you request a different scheme you must replace the `android:scheme` property. Remember to also replace `{YOUR_APP_PACKAGE_NAME}` with your actual application's package name.
 
 Finally, don't forget to add the internet permission.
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -264,60 +264,70 @@ First go to [Auth0 Dashboard](https://manage.auth0.com/#/applications) and go to
...
+        //<---
+    }
+    //...
+}
 ```
 
-Also register the intent filters inside your activity's tag, so you can receive the call in your activity. Note that you will have to specify the callback url inside the `data` tag.
+It's a good practice to define reusable resources like `@string/auth0_domain` but you can also hard code the value. In case you're using a [custom scheme](#a-note-about-app-deep-linking) you must update the `auth0Scheme` property.
+
+Alternatively you can define your own **intent-filter** for the `RedirectActivity` in the `AndroidManifest.xml` file replacing the one defined by the library. If you choose to do this, the `manifestPlaceholders` don't need to be set. In your manifest inside your application's tag add the `RedirectActivity` declaration:
```
hzalaz: Alternatively you can declare the `RedirectActivity` in the `AndroidManifest.xml` file with your own **intent-filter** so it overrides the library's default.

hzalaz:
If we promote https and is the best UX we can make it the default in the library, remove the placeholder and let the user declare the whole activity if he needs custom schemes.

lbalmaceda:
Another thing with the usage of placeholders: If the user doesn't want to use `WebAuthProvider` and just uses the `AuthenticationAPIClient` (i.e. for createUser/login) then the compiler will force him to set a valid placeholder anyway.

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Attempst to bind the Custom Tabs Service to the Context and opens a Uri as soon as possible.
+     *
+     * @param uri the uri to open in a Custom Tab or Browser.
+     * @return true if the request to bind the service was successful, false if the service was already bound or it couldn't be bound.
+     */
+    public boolean bindServiceAndLaunchUri(@NonNull Uri uri) {
+        nextUri = uri;
+        boolean boundRequestSuccess = bindService();
+        if (isBound || !boundRequestSuccess) {
```
nikolaseu: I think the negation is wrong?

nikolaseu: I mean
"launch uri if service was already bound or was just bounded successfully" 
but then.. why call bindService if it was already bound?

nikolaseu: anybody cares if it was already bound or was bound in this call?

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Attempts to unbind the Custom Tabs Service from the Context.
+     */
+    public void unbindService() {
+        Log.v(TAG, "Trying to unbind the service");
+        Context context = this.context.get();
+        if (context != null) {
+            context.unbindService(this);
+        }
+        nextUri = null;
```
nikolaseu: should this be also cleared when the uri is launched?

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
nikolaseu: ok I get it.
but from the way it's used, this can never be "already bound" because a new controller is created each time. maybe something to optimize in the activity that uses this in https://github.com/auth0/Auth0.Android/pull/95/files#diff-101f8235cd64f6d21cfc3fc82d610bd5R105

lbalmaceda:
I decided to go with the default `https` scheme option. On my tests with the different emulators, the longer it took to bind and connect to the Custom Tabs Service was ~600ms. I've set a max non-blocking wait time of 1 second for the service to connect, that IMHO could be raised to 1.5 or 2 safely. If the service is not connected in that given time, a call to `launchUri` will use a simple Browser instead of the Custom Tabs Browser.

PR #100, opened 2017-07-03
Add OIDC audience notes to the readme

lbalmaceda:


lbalmaceda:
@cocojoe will do that in a separate PR. Thanks for the suggestion

PR #102, opened 2017-07-05
Add hasValidCredentials and clearCredentials to CredentialsManager

lbalmaceda:


README.md
```diff
@@ -449,6 +455,13 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
 ```
 
 
+5. **Clear credentials**
+Whenever you need to remove the credentials from the storage and make sure that the next call to `getCredentials` fails, you must clear them from the manager:
```
nikolaseu: this is too technical, maybe it's simpler to just say "when you want to log the user out".

> and make sure that the next call to `getCredentials` fails

saying it like this sounds weird. my goal is not to have the call fail, that it fails is just a side effect for being logged out

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -319,4 +321,68 @@ public void shouldGetAndFailToRenewExpiredCredentials() throws Exception {
...
+    @Test
+    public void shouldClearCredentials() throws Exception {
+        manager.clearCredentials();
+
+        verify(storage).remove("com.auth0.id_token");
+        verify(storage).remove("com.auth0.access_token");
+        verify(storage).remove("com.auth0.refresh_token");
+        verify(storage).remove("com.auth0.token_type");
+        verify(storage).remove("com.auth0.expires_at");
+        verify(storage).remove("com.auth0.scope");
```
nikolaseu: just to avoid forgetting checks in the future (supposed you add a new key) you can also verify for "no more interactions" in storage

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -319,4 +321,68 @@ public void shouldGetAndFailToRenewExpiredCredentials() throws Exception {
...
+        long expirationTime = CredentialsMock.CURRENT_TIME_MS + 123456L * 1000;
+        when(storage.retrieveLong("com.auth0.expires_at")).thenReturn(expirationTime);
+
+        when(storage.retrieveString("com.auth0.id_token")).thenReturn("idToken");
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn(null);
+        assertTrue(manager.hasCredentials());
+
+        when(storage.retrieveString("com.auth0.id_token")).thenReturn(null);
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn("accessToken");
+        assertTrue(manager.hasCredentials());
```
nikolaseu: any reason for the change on the way you assert things?

README.md
```diff
@@ -433,8 +433,14 @@ authentication
     });
 ```
 
-3. **Retrieve credentials**
- Existing credentials will be returned if they are still valid, otherwise the `refresh_token` will be used to attempt to renew them. If the `expires_in` or both the `access_token` and `id_token` values are missing, the method will throw a `CredentialsManagerException`. The same will happen if the credentials have expired and there's no `refresh_token` available.
+3. **Check credentials existence**
+There are cases were you just want to check if a user session is still valid without the need to use the user credentials right away (i.e. to know if you should present the login screen or the main screen). You could achieve a similar result by calling `getCredentials` and checking the success/failure call, but for convenience we include a `hasCredentials` method that can know in advance if a valid token is available or if it has expired but it could be refreshed, without making an additional network call. The same rules of the `getCredentials` method apply:
```
nikolaseu: > There are cases were you just want to check if a user session is still valid without the need to use the user credentials right away

I think "without the need to use the user credentials right away" can be omitted. It's simple and the example is more than enough.

> You could achieve a similar result

It's not a similar result, this is not a second option. You haven't mentioned the "better" option yet. But I think you just don't even have to mention the "previous workaround". At least not before the simpler/official way.

> for convenience we include a `hasCredentials` method that can know in advance

that can *let you* know

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -97,6 +97,30 @@ public void onFailure(AuthenticationException error) {
...
+    /**
+     * Checks if a valid pair of credentials can be obtained from this manager.
+     */
+    public boolean hasCredentials() {
+        String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
+        String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
+        String idToken = storage.retrieveString(KEY_ID_TOKEN);
+        Long expiresAt = storage.retrieveLong(KEY_EXPIRES_AT);
+
+        return !(isEmpty(accessToken) && isEmpty(idToken) || expiresAt == null || expiresAt <= getCurrentTimeInMillis() && refreshToken == null);
```
lbalmaceda: parenthesis is not needed because the `||` separates the conditions. I will add them anyway to make it easier to read. 👍

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
lbalmaceda: you mean `assertTrue(something)` instead of `asserThat(something, is(true))`?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -97,6 +97,32 @@ public void onFailure(AuthenticationException error) {
         });
     }
 
+    /**
+     * Checks if a valid pair of credentials can be obtained from this manager.
+     */
+    public boolean hasCredentials() {
```
hzalaz: I'd mention that they are valid credentials not only he existence of one. Also I'd reflect that in the name

PR #108, opened 2017-07-14
Ability to add custom scopes

tplkn:
Now it's impossible to add custom scope to `AuthenticationAPIClient` requests.

This PR adds the ability to specify custom scopes for those requests.


lbalmaceda:
@tplkn most of the methods of the **AuthenticationAPIClient** return an `AuthenticationRequest`. That class already has setters for common values, that even return the same request instance for you to chain the calls. The `scope` setter is [here](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/request/AuthenticationRequest.java#L43). 
I'll add a section for request customization in the README. But it won't always have all the available combinations, the next time you'll have to be a little more curious. 😄

tplkn:
@lbalmaceda Thank you.

PR #110, opened 2017-07-18
Add a manifest placeholder for configuring the scheme

lbalmaceda:
Related: https://github.com/openid/AppAuth-Android/issues/248

### DIFF
--- a/auth0/src/main/AndroidManifest.xml
+++ b/auth0/src/main/AndroidManifest.xml
@@ -37,7 +37,7 @@
         <activity
             android:name="com.auth0.android.provider.RedirectActivity"
             android:exported="true">
-            <intent-filter>
+            <intent-filter android:autoVerify="true">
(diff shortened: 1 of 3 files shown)


auth0/src/main/AndroidManifest.xml
```diff
@@ -37,7 +37,7 @@
         <activity
             android:name="com.auth0.android.provider.RedirectActivity"
             android:exported="true">
-            <intent-filter>
+            <intent-filter android:autoVerify="true">
```
hzalaz: Could this break if we specify custom scheme?

lbalmaceda: It doesn't break, at least compiler doesn't scream and the apk installs and runs ok. The reason I added is because people using app links would otherwise need to manually override this intent-filter declaration with their own in the manifest.

PR #113, opened 2017-08-09
Support password expiration. Add self_change_password endpoint

lbalmaceda:


cocojoe:
@lbalmaceda I don't think we still need this, can you remove reviewers and close?

PR #115, opened 2017-08-29
Add an encrypted version of the CredentialsManager

lbalmaceda:
This PR adds a new class, _similar but separate_ to the existing `CredentialsManager` one, that encrypts the data before persisting it and can also request the user authentication before accessing the stored credentials. This authentication behavior is disabled by default. By using the `@RequiresAPI` annotation I can make the compiler show an error if the developer is using this class on lower APIs. It requires minimum API 21 because that's when the "Confirm credentials" intent is first available.

#### Security discussion
- Encryption is done with an AES key of 256 bits of size. (We can gain some performance if I lower it to 128 I guess, but it's fast enough).
- The AES key is saved in the Storage encrypted using an RSA key pair of 2048 bits of size.
- The RSA key pair is saved in the Android KeyStore. I can generate keys and ask the OS to use them in the encryption/decryption process, but I can never access the actual key values.
- Detailed implementation can be seen in the `CryptoUtil` class.

#### Pending stuff
- [x] `CryptoUtil` tests. White-box tests using PowerMock and Mockito. Not many alternatives here... The coverage report will show "zero coverage" because of Jacoco [lack of support](https://github.com/powermock/powermock/wiki/Code-coverage-with-JaCoCo#on-the-fly-instrumentation) for the PowerMock test runner.
- [x] `CryptoManager` tests. Including the "requires authentication" flow.
- [x] Readme. Make clear which manager to use depending on the Android API level and how to support the authentication feature (i.e. add an example).
- [x] Add some exception catching tests
- [x] Try this manager in a sample like "session-handling" on API 21 and 23
- [x] Rename `CryptoManager` to something cooler, like "EnhancedCredentialsManager". What do you think?


README.md
```diff
@@ -434,20 +434,26 @@ users
 > In all the cases, the `User ID` parameter is the unique identifier of the auth0 account instance. i.e. in `google-oauth2|123456789081523216417` it would be the part after the '|' pipe: `123456789081523216417`.
 
 
-### Credentials Manager
-This library ships with a `CredentialsManager` class to easily store and retrieve fresh Credentials from a given `Storage`.
+
+## Credentials Manager
+
+This library ships with two additional classes that help you store and retrieve the `Credentials` received in authentication calls. Depending on the minimum API level that your application is targeting you'd like to use a different implementation.
```
cocojoe: you store and retrieve the `Credentials` received in authentication calls. ->
you manage the `Credentials` received during authentication.

README.md
```diff
@@ -434,20 +434,26 @@ users
...
+
+The basic version supports asking for `Credentials` existence, storing them and getting them back. If the credentials have expired and a refresh_token was saved, they are automatically refreshed. The class is called `CredentialsManager`.
 
 #### Usage
-1. **Instantiate the manager**
-You'll need an `AuthenticationAPIClient` instance used to renew the credentials when they expire and a `Storage`. The Storage implementation is up to you. We provide a `SharedPreferencesStorage` that uses `SharedPreferences` to create a file in the application's directory with Context.MODE_PRIVATE mode. This implementation is thread safe and can either be obtained through a Singleton like method or be created every time it's needed.
+1. **Instantiate the manager:**
+You'll need an `AuthenticationAPIClient` instance used to renew the credentials when they expire and a `Storage`. The Storage implementation is up to you. We provide a `SharedPreferencesStorage` that uses `SharedPreferences` to create a file in the application's directory with **Context.MODE_PRIVATE** mode. This implementation is thread safe and can either be obtained through a Singleton method or be created every time it's needed.
```
cocojoe: You'll need an `AuthenticationAPIClient` instance ~~used~~ to renew the credentials when they expire and a `Storage` instance. ~~The Storage implementation is up to you.~~

This implementation is thread safe and can either be obtained through a Singleton method or be created every time it's needed. ->
This implementation is thread safe and can either be obtained through a shared method or on demand.

README.md
```diff
@@ -491,13 +497,56 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
...
 When you want to log the user out:
 
 ```java
 manager.clearCredentials();
 ```
 
+
+### Encryption enforced (Min API 21)
+
+The enhanced version contains the same methods as the _Basic_ manager but encrypts the data before storing it, and in those devices where a Secure LockScreen has been configured it can require the user authentication before letting them obtain the stored credentials. The class is called `CryptoManager`.
```
cocojoe: The enhanced version contains the same methods as the _Basic_ manager but encrypts the data before storing it, and in ->
This version expands the minimum version and adds encryption to the data storage. In

README.md
```diff
@@ -491,13 +497,56 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
...
 ```
 
+
+### Encryption enforced (Min API 21)
+
+The enhanced version contains the same methods as the _Basic_ manager but encrypts the data before storing it, and in those devices where a Secure LockScreen has been configured it can require the user authentication before letting them obtain the stored credentials. The class is called `CryptoManager`. 
+  
+  
+#### Usage
+The usage is the same as in the _Basic_ version. What changes is the way you instantiate the manager as it now requires a valid `Context`.
```
cocojoe: Change this, don't like calling them basic etc don't need to say it. 

In this version the manager requires a valid `context as shown below:

cocojoe:
Modify your codecov.yml to allow for a little variance in overall and patch.  For me -0.26% is acceptable so is 80% coverage in patch as sometimes it's just not possible. See `threshold`

https://github.com/auth0/Auth0.swift/blob/master/codecov.yml

Issue #109, opened 2017-07-17
Proposal: CredentialsManager.getCredentials() should save renewed Credentials


psam44:
CredentialsManager.getCredentials() is able to renew expired credentials and that is valuable. But this renewal is not persisted, so every time getCredentials() is called there is a network hit if the currently stored item has expired. Furthermore:

- getCredentials() doesn't say if the returned item is from the storage or a renewed one.
- hasValidCredentials() doesn't tell the difference neither.

So the only way I found for now would be to use an other storage unit to keep the last ExpiresAt for comparison with the returned credentials and in case of progression to do a saveCredentials().
I would prefer this process be done by the CredentialsManager.

Something like:
```
            @Override
            public void onSuccess(Credentials freshCredentials) {
/* + */         CredentialsManager.this.saveCredentials(freshCredentials);
                callback.onSuccess(freshCredentials);
            }
```

[Edit 28 Aug]
I discovered another issue with the call to `renewAuth(refreshToken)`, which may introduce an infeasibility:
My freshCredentials has these unexpected properties:
`getIdToken()` is null.
`getRefreshToken()` is null.
`getExpiresIn()` is 86400, i.e. 24 hours ; I don't know where it comes from.

These ones seem correct:
`getAccessToken()` is a 16 chars string.
`getType()` is "Bearer".
`getScope()` is null.

Note: I also made a trial with my own duplicate of this part of source, to insert `.addParameter("scope", "openid offline_access")` before `.start()` and the result is the same.

My setting is `setOIDCConformant(true)`.
The initial credentials comes from react-native-lock 0.6.1, with Lock 1.17.+, so the endpoints are /oauth/ro and /tokeninfo. The requested scope is the default one, i.e. "openid offline_access".
Is it possible that this refreshToken is not suitable for the oauth/token endpoint?

lbalmaceda:
The save credentials feature is currently being added here https://github.com/auth0/Auth0.Android/pull/118.

When renewing credentials using the refresh_token, you won't receive a new refresh_token because the old one is still valid 😄 , so that's OK. But that thing you mention about the id_token being null is a little weird. Could you please share the snippets on how you're 1) initializing the Auth0, 2) the AuthenticationAPIClient and 3) making the requests?

psam44:
Initial acquisition:
```javascript
import Auth0Lock from 'react-native-lock';
import credentials from './auth0-credentials';
const lock: Auth0Lock = new Auth0Lock(credentials);
//...
[...]

PR #118, opened 2017-09-27
Make Credential Managers save the refreshed value

lbalmaceda:
~My only concern in this PR is that when serializing the `Credentials` object, Gson (our Json library) will serialize Java Dates to a string value (not long timestamps), thus losing the precision (milis). Should I make the serializer serialize Longs instead? (I've to code that logic) or is it an edge case and we're ok having a 1 second expired `Credentials`?~
EDIT: Fixed. Now Gson serializes dates in a correct format including Timezone and Millis.

Depends on this PR getting merged: https://github.com/auth0/Auth0.Android/pull/115. Once merged, a rebase will make diff clearer here.

Closes https://github.com/auth0/Auth0.Android/issues/109

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,11 +441,11 @@ This library ships with two additional classes that help you manage the Credenti
 
 ### Basic (Min API 15)
 
-The basic version supports asking for `Credentials` existence, storing them and getting them back. If the credentials have expired and a refresh_token was saved, they are automatically refreshed. The class is called `CredentialsManager`.
+The basic version supports asking for `Credentials` existence, storing them and getting them back. If the credentials have expired and a refresh_token was saved, they are automatically refreshed. The class is called `CredentialsManager` and requires at minimum Android API 15.
 
 #### Usage
 1. **Instantiate the manager:**
-You'll need an `AuthenticationAPIClient` instance to renew the credentials when they expire and a `Storage`. We provide a `SharedPreferencesStorage` that uses `SharedPreferences` to create a file in the application's directory with **Context.MODE_PRIVATE** mode. This implementation is thread safe and can either be obtained through a shared method or on demand.
+You'll need an `AuthenticationAPIClient` instance to renew the credentials when they expire and a `Storage` object. We provide a `SharedPreferencesStorage` class that makes use of `SharedPreferences` to create a file in the application's directory with **Context.MODE_PRIVATE** mode. This implementation is thread safe and can either be obtained through a shared method or on demand.
 
 ```java
 AuthenticationAPIClient authentication = new AuthenticationAPIClient(account);
@@ -507,7 +507,7 @@ manager.clearCredentials();
 
 ### Encryption enforced (Min API 21)
 
-This version expands the minimum version and adds encryption to the data storage. In those devices where a Secure LockScreen has been configured it can require the user authentication before letting them obtain the stored credentials. The class is called `SecureCredentialsManager`.
+This version expands the minimum version and adds encryption to the data storage. In those devices where a Secure LockScreen has been configured it can require the user authentication before letting them obtain the stored credentials. The class is called `SecureCredentialsManager` and requires at minimum Android API 21.
 
 
 #### Usage
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -65,7 +65,7 @@ public class CredentialsManager {
      */
     public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
         String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
-        String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
+        final String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
         String idToken = storage.retrieveString(KEY_ID_TOKEN);
         String tokenType = storage.retrieveString(KEY_TOKEN_TYPE);
         Long expiresAt = storage.retrieveLong(KEY_EXPIRES_AT);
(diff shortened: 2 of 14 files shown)


cocojoe:
@lbalmaceda is it possible to improve your code-coverage, is this an issue due to the test env only covering a specific API version so other tests can't be executed?

README.md
```diff
@@ -434,20 +434,26 @@ users
 > In all the cases, the `User ID` parameter is the unique identifier of the auth0 account instance. i.e. in `google-oauth2|123456789081523216417` it would be the part after the '|' pipe: `123456789081523216417`.
 
 
-### Credentials Manager
-This library ships with a `CredentialsManager` class to easily store and retrieve fresh Credentials from a given `Storage`.
```
cocojoe: Personally I don't like implying levels of difficulty in usage.

~class to easily store and retrieve fresh Credentials from a given~ -> utility class to manage the store, retrieve and renew credentials from a given `Storage`.

Do we need to specify from a given storage?

README.md
```diff
@@ -434,20 +434,26 @@ users
...
+
+### Basic (Min API 15)
+
+The basic version supports asking for `Credentials` existence, storing them and getting them back. If the credentials have expired and a refresh_token was saved, they are automatically refreshed. The class is called `CredentialsManager`.
 
 #### Usage
-1. **Instantiate the manager**
-You'll need an `AuthenticationAPIClient` instance used to renew the credentials when they expire and a `Storage`. The Storage implementation is up to you. We provide a `SharedPreferencesStorage` that uses `SharedPreferences` to create a file in the application's directory with Context.MODE_PRIVATE mode. This implementation is thread safe and can either be obtained through a Singleton like method or be created every time it's needed.
+1. **Instantiate the manager:**
+You'll need an `AuthenticationAPIClient` instance to renew the credentials when they expire and a `Storage`. We provide a `SharedPreferencesStorage` that uses `SharedPreferences` to create a file in the application's directory with **Context.MODE_PRIVATE** mode. This implementation is thread safe and can either be obtained through a shared method or on demand.
```
cocojoe: ~and a `Storage`~ -> and a `Storage` object.

~We provide a `SharedPreferencesStorage` that uses `SharedPreferences`~ -> We provide a `SharedPreferencesStorage` class that implements `SharedPreferences`

README.md
```diff
@@ -491,13 +497,56 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
...
-5. **Clear credentials**
+5. **Clear credentials:**
 When you want to log the user out:
 
 ```java
 manager.clearCredentials();
 ```
 
+
+### Encryption enforced (Min API 21)
```
cocojoe: Is the language used typically `min API X` or `API X+` ?

README.md
```diff
@@ -491,13 +497,56 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
...
+
+```java
+AuthenticationAPIClient authentication = new AuthenticationAPIClient(account);
+Storage storage = new SharedPreferencesStorage(this);
+SecureCredentialsManager manager = new SecureCredentialsManager(this, authentication, storage);
+```
+
+#### Requiring Authentication
+
+You can require the user authentication to obtain credentials. This will make the manager prompt the user with the device's configured LockScreen, which they must pass correctly in order to obtain the credentials. **This feature is only available on devices where the user has setup a secured LockScreen** (PIN, Pattern, Password or Fingerprint).
```
cocojoe: Seems wordy, do we need to explain what a user has to do to pass a LockScreen?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -65,7 +65,7 @@ public void saveCredentials(@NonNull Credentials credentials) {
      */
     public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
         String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
-        String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
+        final String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
```
cocojoe: How come this one is now final?

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.java
```diff
@@ -0,0 +1,251 @@
...
+     * @param description the text to use as description in the authentication screen. On some Android versions it might not be shown.
+     * @return whether this device supports requiring authentication or not.
+     */
+    public boolean requireAuthentication(@NonNull Activity activity, int requestCode, @Nullable String title, @Nullable String description) {
+        if (requestCode < 1 || requestCode > 255) {
+            throw new IllegalArgumentException("Request code must a value between 1 and 255.");
+        }
+        KeyguardManager kManager = (KeyguardManager) activity.getSystemService(Context.KEYGUARD_SERVICE);
+        this.authIntent = Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP ? kManager.createConfirmDeviceCredentialIntent(title, description) : null;
+        this.authenticateBeforeDecrypt = (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && kManager.isDeviceSecure()
```
cocojoe: Readability would be better with parenthesis.  Think of parentheses like comments in code: they aren't strictly necessary, and in theory a competent developer should be able to figure out code without them. And yet, these cues are exceedingly helpful.

cocojoe:
Regarding your question on date storage, I think 1 second is fine, given that the expiry time is say 12 or 24 hours?

README.md
lbalmaceda: i.e. `Retrofit requires at minimum Java 7 or Android 2.3.`

README.md
lbalmaceda: I think the interesting point here is letting them know that no extra PIN/Pwd needs to bet setup/remembered by them, as we use the one they already use for their phone. And of course, if the phone didn't have any secure lockscreen setup, this won't work.

README.md
lbalmaceda: 🤔  I think you're commenting on the deleted line.

README.md
lbalmaceda: Hmm. But `SharedPreferences` is a KeyValue store **class** rather than an **interface** that you could implement.

cocojoe: I wasn't sure tbh just mentioned it

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: Because I use it on an anonymous function right below and needs any variable accessed from outside its scope to be final.  -> https://github.com/auth0/Auth0.Android/pull/118/files#diff-2651976c12b43b2192543b92b10ed2e4R91

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.java
lbalmaceda: I will add them but TBH it would almost look the same, as the new line is already doing that thing you mention 😛

Issue #116, opened 2017-09-11
Cannot minimize application with opened Chrome Tabs


MikolajKakol:
If during login process user minimize application and restore it we will get an error in `AuthCallback.onFailure(AuthenticationException)` method with

> exception = {AuthenticationException@4900} "com.auth0.android.authentication.AuthenticationException: An error occurred when trying to authenticate with the server."
>  code = "invalid_request"
>  description = "Missing required parameter: code"
>  statusCode = 0
>  values = {HashMap@4908}  size = 2
>   0 = {HashMap$HashMapEntry@4917} "error" -> "invalid_request"
>   1 = {HashMap$HashMapEntry@4918} "error_description" -> "Missing required parameter: code"
>  backtrace = {Object[10]@4909} 
>  cause = {AuthenticationException@4900} "com.auth0.android.authentication.AuthenticationException: An error occurred when trying to authenticate with the server."
>  detailMessage = "An error occurred when trying to authenticate with the server."

it's important for us to handle those situations correctly because if you are trying to log in with mobile using Norwegian BankID user is prompt with system Acitivty (launched from USSD code I think) in which user must enter some number and when it's closed I'm getting that error.

It's easily to reproduce in sample 00-login. Just click "show log in" then minimize app and restore.

lbalmaceda:
@MikolajKakol This should be fixed in the latest patch. Please try `1.10.1` and reopen if necessary.

PR #121, opened 2017-10-03
Make CCT stay alive when activity is paused

lbalmaceda:
By removing the flag, the CCT is not closed when the activity is paused and resumed. This is already working for Browser (not CCT mode) and Webview authentication. It might fix https://github.com/auth0/Auth0.Android/issues/116.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
@@ -126,7 +126,6 @@ class CustomTabsController extends CustomTabsServiceConnection {
                 final Intent intent = new CustomTabsIntent.Builder(session.get())
                         .build()
                         .intent;
-                intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
                 intent.setData(uri);
                 try {
                     context.startActivity(intent);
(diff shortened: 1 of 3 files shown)


auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -126,7 +126,6 @@ public void run() {
                 final Intent intent = new CustomTabsIntent.Builder(session.get())
                         .build()
                         .intent;
-                intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
```
cocojoe: Removing does not cause any other issues? Only asking as issues that required this to fix in RNA.

lbalmaceda: CCT is not another tab that stays open on the browser, is more like an activity. By removing the flag you make the activity stay up after the app is paused/resumed. Yes, we should probably do the same for RNA and just keep that flag for browser (not CCT) authentication. I didn't do the PR yet because it's not just removing the flag.. CustomTabsIntent may fail to find a package that could "handle CCT and so default to use Browser. We must catch and add the flag to that case too.

Issue #126, opened 2017-10-30
Support TLSv1.2 on older Android API


carlyonj:
Auth0 supports down to API 15 but as detailed [here](https://github.com/ssllabs/ssllabs-scan/issues/258) TLSv1.1 and v1.2 are not used by default on older APIs. Potential solutions are detailed [here](https://stackoverflow.com/questions/28943660/how-to-enable-tls-1-2-support-in-an-android-application-running-on-android-4-1), [here](https://github.com/square/okhttp/issues/2372) and [here](https://stackoverflow.com/questions/28329652/enabling-specific-ssl-protocols-with-android-webviewclient).

Currently by using the Auth0 API you are unable to support devices using APIs lower than 21 with a backend that enforced TLSv1.2. Because of this the Auth0 android library is not very useful to any apps that plan on supporting older devices in a secure manner.

lbalmaceda:
I would prefer a method that sets a flag rather than exposing the OkHttp client. Maybe in the Auth0 class, or in both API client classes. This is not a priority for us.. Feel free to fork the repo and add the required changes or propose a PR and I'll review it when I can.

dj-mal:
PR Submitted as #128

lbalmaceda:
This has been released on version 1.12.0. Thanks

PR #128, opened 2017-11-06
Add support for TLS1.2 on pre-lollipop devices.

dj-mal:
* Add method `enableTls12OnPreLollipop` to `UsersAPIClient` and `AuthenticationAPIClient`, which would enable support for TLS 1.2 on Pre-lollipop devices when called.
* Add `OkHttpTls12Compat` and `Tls12SocketFactory ` to support TLS 1.2 by patching the socket on devices with API level 16-21.
* References:
  * https://developer.android.com/reference/javax/net/ssl/SSLSocket.html
  * https://github.com/square/okhttp/issues/2372

Fixes https://github.com/auth0/Auth0.Android/issues/126


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -160,6 +161,11 @@ public void setUserAgent(String userAgent) {
         factory.setUserAgent(userAgent);
     }
 
+    @SuppressWarnings("unused")
+    public void enableTls12OnPreLollipop() {
```
lbalmaceda: Rename to `enableTLS12OnPreLollipop`

auth0/src/test/java/com/auth0/android/util/OkHttpTls12CompatTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    @Config(sdk=22)
+    public void shouldNotConfigTlsPostApi21() {
+        OkHttpTls12Compat.enableSupportOnPreLollipop(client);
+        verify(client, never()).setSslSocketFactory((SSLSocketFactory) any());
+    }
+
+    @Test
+    public void shouldConfigTlsOnOrPreApi21() {
```
lbalmaceda: I know the class already has it, but add an explicit `@Config(sdk=21)` line here too.

auth0/src/test/java/com/auth0/android/util/OkHttpTls12CompatTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Before
+    public void setUp(){
+        MockitoAnnotations.initMocks(this);
+        activity = Robolectric.setupActivity(Activity.class);
+    }
+
+    @Test
+    @Config(sdk=22)
+    public void shouldNotConfigTlsPostApi21() {
```
lbalmaceda: Because the min API for this library is 15 and that's excluded from the TLS logic, please add a test for that one as well.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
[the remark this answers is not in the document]
dj-mal: If this is made package private, the users have no means to enable TLS 1.2 on the client, since they need to call this method explicitly to enable support. Do you want to move this into the constructor?

auth0/src/main/java/com/auth0/android/management/UsersAPIClient.java
```diff
@@ -132,6 +133,12 @@ public void setUserAgent(String userAgent) {
         factory.setUserAgent(userAgent);
     }
 
+    @SuppressWarnings("unused")
+    public void enableTls12OnPreLollipop() {
```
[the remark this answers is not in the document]
dj-mal: If this is made package private, the users have no means to enable TLS 1.2 on the client, since they need to call this method explicitly to enable support. Do you want to move this into the constructor?

auth0/src/main/java/com/auth0/android/util/OkHttpTls12Compat.java
```diff
@@ -0,0 +1,57 @@
...
+import com.squareup.okhttp.TlsVersion;
+
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.net.ssl.SSLContext;
+
+public final class OkHttpTls12Compat {
```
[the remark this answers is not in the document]
dj-mal: We can't make this package private because both `AutehnticationAPIClient` and `UsersAPIClient` use it, and they're in different package.

auth0/src/main/java/com/auth0/android/util/OkHttpTls12Compat.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private OkHttpTls12Compat() {}
+
+    /**
+     * Enable TLS 1.2 on the OkHttpClient on API 16-21, which is supported but not enabled by default.
+     * @link https://github.com/square/okhttp/issues/2372
+     * @see Tls12SocketFactory
+     * @param client OkHttpClient instance to be modified
+     */
+    public static void enableSupportOnPreLollipop(OkHttpClient client) {
```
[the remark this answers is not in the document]
dj-mal: Method renamed.

auth0/src/main/java/com/auth0/android/util/OkHttpTls12Compat.java
```diff
@@ -0,0 +1,57 @@
...
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.LOLLIPOP) {
+            // No need to modify client as TLS 1.2 is enabled by default on API21+
+            // Lollipop is included because some Samsung devices face the same problem on API 21.
+            return;
+        } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
+            Log.w(TAG, "TLS 1.2 not supported on API < 16");
```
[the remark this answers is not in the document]
dj-mal: Removed the log line and simplified the if condition.

auth0/src/main/java/com/auth0/android/util/Tls12SocketFactory.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Enables TLS v1.2 when creating SSLSockets.
+ * <p/>
+ * For some reason, android supports TLS v1.2 from API 16, but enables it by
+ * default only from API 20.
+ * @link https://developer.android.com/reference/javax/net/ssl/SSLSocket.html
+ * @see SSLSocketFactory
+ */
+public class Tls12SocketFactory extends SSLSocketFactory {
```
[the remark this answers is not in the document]
dj-mal: Renamed and made package private.

auth0/src/test/java/com/auth0/android/util/OkHttpTls12CompatTest.java
dj-mal: I tried but Robolectric doesn't support API level 15 so we can't just add that like other supported APIs.

auth0/src/test/java/com/auth0/android/util/OkHttpTls12CompatTest.java
dj-mal: Added config annotation.

dj-mal:
I tried making `OkHttpTLS12Compat` non static, but that feels a bit clumsy in my opinion. I've made a new branch with that:
https://github.com/dj-mal/Auth0.Android/compare/prelollipop-tls...dj-mal:prelollipop-tls-non-static?expand=1

If you think that's OK I can merge the changes into this PR

lbalmaceda:
I thought about this and actually, it makes sense to have the TLS flag inside the `Auth0` class, like we do for the logging and the OIDC flag. *Based on the non-static branch*, these are the changes I'd make:

- Rename `OkHttpTLS12Compat` to `OkHttpClientFactory`, which has a single method "createClient(boolean loggingEnabled, boolean tls12Enforced)". Both values are obtained from the Auth0 instance.
- Change the non-public constructors of the API Clients that receive an `OkHttpClient` instance to just receive the factory instance. These constructors exist for test purposes only.
- Factory won't be stored in any variable/field on the API Clients... Use it to create and assign the client on the constructor, but don't assign the factory anywhere (let gc do it's thing).
- Helper methods inside the Factory will be private. (or package-private if you need to test them).
- The class will be public and stored inside `com.auth0.android/request/internal`. Can't think of a better place now.. it must be public because both clients are in separate packages.. so keeping it inside the "internal" package will make people aware that we can break them at any time, and that they should not use it directly. That can be clarified in the class javadoc as well..
- Remove flag setters/getters from the API Client. Move them to Auth0. Test them.

Please share your thoughts. I think it's cleaner this way.. Ideally we should hide any helper method/class from the users (or keep them in "internal") and try not to put external dependencies on the methods signatures.

dj-mal:
All are good suggestions and can reduce duplication across the API clients. They're implemented accordingly. 

However, since `OkHttpClientFactory.createClient` uses only parameters from Auth0, maybe an Auth0 instance could be passed (i.e., `createClient(auth0)`) instead of separate booleans (i.e., `createClient(loggingEnabled, tlsEnforced)`)?

auth0/src/main/java/com/auth0/android/request/internal/OkHttpClientFactory.java
```diff
@@ -14,32 +17,49 @@
 
 import javax.net.ssl.SSLContext;
 
-public class OkHttpTLS12Compat {
+public class OkHttpClientFactory {
```
lbalmaceda: Please add an additional javadoc at the class level:


> Factory class used to configure and obtain a new OkHttpClient instance. This class is meant for internal use only, breaking changes may appear at any time without backwards compatibility guarantee. 

or something like that 🙂

auth0/src/main/java/com/auth0/android/request/internal/OkHttpClientFactory.java
```diff
@@ -14,32 +17,49 @@
...
     /**
      * Enable TLS 1.2 on the OkHttpClient on API 16-21, which is supported but not enabled by default.
      * @link https://github.com/square/okhttp/issues/2372
      * @see TLS12SocketFactory
      */
-    public OkHttpTLS12Compat enableForClient() {
+    private void enforceTls12(OkHttpClient client) {
         // No need to modify client as TLS 1.2 is enabled by default on API21+
         // Lollipop is included because some Samsung devices face the same problem on API 21.
         if (client == null || Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN
```
lbalmaceda: Remove `client==null` check as the client is always valid, since the only entry point is the `createClient` method. Add a `@NonNull` annotation to these methods parameters if you want instead, just to help devs as users won't see them.

auth0/src/test/java/com/auth0/android/request/internal/OkHttpClientFactoryTest.java
```diff
@@ -0,0 +1,180 @@
...
+    @Test
+    @Config(sdk=22)
+    public void shouldDisableLoggingTLS12NotEnforced_postLollipop(){
+        List list = generateMockList(mockClient);
+        OkHttpClient client = factory.modifyClient(mockClient, false, false);
+        verifyLoggingDisabled(client, list);
+        verifyTLS12NotEnforced(client);
+    }
+
+    private static List generateMockList(OkHttpClient client) {
```
lbalmaceda: rename `generateInterceptorsMockList`

dj-mal:
Updated according to the comments. Cheers :smiley:

PR #132, opened 2017-11-16
Fix default Locale tests

lbalmaceda:
Somehow, when running the tests the default locale wasn't specified in the environment. When this value is missing, a default hard-coded value is returned. This value should be the same as the one used in the `RequestFactory` class. I've added a public (agh!) constant to prevent this issue from happening again.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
+++ b/auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
@@ -43,6 +43,8 @@ import java.util.Map;
 
 public class RequestFactory {
 
+    public static final String DEFAULT_LOCALE_IF_MISSING = "en_US";
+
     private static final String AUTHORIZATION_HEADER = "Authorization";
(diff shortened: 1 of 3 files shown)


auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
```diff
@@ -43,6 +43,8 @@
 
 public class RequestFactory {
 
+    public static final String DEFAULT_LOCALE_IF_MISSING = "en_US";
```
cocojoe: Is this really the best place for this?

lbalmaceda: @cocojoe Yes. It's the only class that uses it, and we have locale tests both for this class and for the class that uses it (api client). This constant is only needed because of the API client tests. I could remove those locale tests and just keep these ones, but it's nice to test the final outcome as well (i.e. request to /userinfo has locale in header). Must be public because the location in the package doesn't provide visibility otherwise.

PR #141, opened 2018-01-08
Rename hosted login page to universal login

lbalmaceda:


auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -240,7 +240,7 @@ public Builder withParameters(@NonNull Map<String, Object> parameters) {
         }
 
         /**
-         * Use the given connection. By default no connection is specified, so the hosted login page will be displayed.
+         * Use the given connection. By default no connection is specified, so the Universal Login will be displayed.
          *
```
aaguiarz: "so the login page will be displayed"

PR #146, opened 2018-02-20
Add support for MFA using OIDC conformant endpoints

lbalmaceda:
This PR adds support for the new MFA OTP flow which is mainly based on https://github.com/jaredhanson/draft-oauth-mfa/blob/master/Draft-1.0.txt. ~It's still missing the `association_required` error handling (we're blocked) for when the user has not yet enrolled.~ It also adds javadoc fixes.

The only thing I'm not convinced of is throwing an exception when the new "login with otp" method is called on a non-oidc conformant client. Maybe it's better if we let the request to the server fail and return that error response to the user, WDYT?

Also to be reviewed: Name of the login method, currently `loginWithOTP(mfaToken, otp)`.

### Public API changes/additions:

- On the **AuthenticationAPIClient:** New method `loginWithOTP(@NonNull String mfaToken, @NonNull String otp)` to log in sending the mfa-otp grant and the mfa token + otp.
- On the **AuthenticationException:** New method `isMultifactorTokenInvalid()` to evaluate if the error instance corresponds to a "mfa token expired" error.

EDIT: 04/24/18. The `association_required` error code is meant for the challenge endpoint only, which this SDK is not going to implement.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
@@ -40,12 +40,12 @@ import com.auth0.android.request.ParameterizableRequest;
 import com.auth0.android.request.Request;
 import com.auth0.android.request.internal.AuthenticationErrorBuilder;
 import com.auth0.android.request.internal.GsonProvider;
+import com.auth0.android.request.internal.OkHttpClientFactory;
 import com.auth0.android.request.internal.RequestFactory;
 import com.auth0.android.result.Credentials;
 import com.auth0.android.result.DatabaseUser;
 import com.auth0.android.result.Delegation;
 import com.auth0.android.result.UserProfile;
-import com.auth0.android.request.internal.OkHttpClientFactory;
 import com.auth0.android.util.Telemetry;
 import com.google.gson.Gson;
 import com.squareup.okhttp.HttpUrl;
@@ -54,6 +54,7 @@ import com.squareup.okhttp.OkHttpClient;
 import java.util.Map;
 
 import static com.auth0.android.authentication.ParameterBuilder.GRANT_TYPE_AUTHORIZATION_CODE;
+import static com.auth0.android.authentication.ParameterBuilder.GRANT_TYPE_MFA_OTP;
 import static com.auth0.android.authentication.ParameterBuilder.GRANT_TYPE_PASSWORD;
 import static com.auth0.android.authentication.ParameterBuilder.GRANT_TYPE_PASSWORD_REALM;
 import static com.auth0.android.authentication.ParameterBuilder.ID_TOKEN_KEY;
@@ -81,6 +82,8 @@ public class AuthenticationAPIClient {
     private static final String OAUTH_CODE_KEY = "code";
     private static final String REDIRECT_URI_KEY = "redirect_uri";
     private static final String TOKEN_KEY = "token";
+    private static final String MFA_TOKEN_KEY = "mfa_token";
+    private static final String ONE_TIME_PASSWORD_KEY = "otp";
     private static final String DELEGATION_PATH = "delegation";
     private static final String ACCESS_TOKEN_PATH = "access_token";
     private static final String SIGN_UP_PATH = "signup";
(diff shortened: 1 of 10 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -235,6 +239,43 @@ public AuthenticationRequest login(@NonNull String usernameOrEmail, @NonNull Str
...
+     * }
+     * </pre>
+     *
+     * @param mfaToken the token received in the previous {@link #login(String, String, String)} response.
+     * @param otp      the one time password code provided by the resource owner, typically obtained from an
+     *                 MFA application such as Google Authenticator or Guardian.
+     * @return a request to configure and start that will yield {@link Credentials}
+     */
+    @SuppressWarnings("WeakerAccess")
+    public AuthenticationRequest loginWithOTP(@NonNull String mfaToken, @NonNull String otp) {
```
cocojoe: I can't think of a better name right now but it would be good to be consistent on platforms, In swift the counterpart might be defined as `login(withOTP: `

lbalmaceda: Ideally we can have a single "login" method that just sets the `client_id` and the path `/oauth/token`. But then users will need to know for each grant type what parameters are accepted. That's the reason we now have a new login "with otp" method that helps to construct this request.

So in swift what are you doing, is that "withOTP" like a flag?

cocojoe: No, so in Swift there are 2 login methods one is /ro the other grant `password-realm`. In this case if I had `login(withOTP otp: String, mfaToken: String)`

The auto-complete would show 3 logins and it's based on the params, in Swift the guideline is to create methods that have meaninful names. You could have `login(otp: String, token: String)` which isn't as informative.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
```diff
@@ -157,17 +157,22 @@ public boolean isInvalidConfiguration() {
...
     }
 
     /// When MFA is required and the user is not enrolled
     public boolean isMultifactorEnrollRequired() {
-        return "a0.mfa_registration_required".equals(code);
+        return "a0.mfa_registration_required".equals(code) || "unsupported_challenge_type".equals(code);
+    }
+
+    /// When the MFA Token received on the login request has expired
+    public boolean isMultifactorTokenExpired() {
```
cocojoe: Why not the more generic `isMultifactorTokenInvalid` as an expired token is invalid, does it matter how? You also have `{"error":"invalid_grant","error_description":"Malformed mfa_token"}`

lbalmaceda: I didn't catch that error response, probably because I tested it on Lock directly and the token is handled from one endpoint to the other as it is. Yeah it makes sense to handle both cases under the same method, and I agree that in that case it should be named like that -> Invalid.

cocojoe:
@lbalmaceda Should we mention that you need to enable the `MFA` grant-type? Just as we had to mention this for `Password`?

lbalmaceda:
@cocojoe Totallyl. I've updated the PR adding those changes.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
```diff
@@ -157,17 +157,23 @@ public boolean isInvalidConfiguration() {
...
     /// When MFA code is required to authenticate
     public boolean isMultifactorRequired() {
-        return "a0.mfa_required".equals(code);
+        return "mfa_required".equals(code) || "a0.mfa_required".equals(code);
     }
 
     /// When MFA is required and the user is not enrolled
     public boolean isMultifactorEnrollRequired() {
-        return "a0.mfa_registration_required".equals(code);
+        return "a0.mfa_registration_required".equals(code) || "unsupported_challenge_type".equals(code);
```
cocojoe: The error is the specific challenge type is not expected vs enrolment in general. Should we clarify this? As it could be they don't support OTP but do OOB?

lbalmaceda: `association_required` will be the new not-enrolled error code once they release the new API, currently this is `unsupported_challenge_type`. On the other side, "unsupported_challenge_type" _does_ mean that the challenge is not supported. Because we're not going to support other challenge type than OTP, I don't think that's an error we can provide an alternate path to the user to continue the authentication. I tried to find without success, but I think somewhere in the docs they mention that OTP must be supported at least.

cocojoe:
@lbalmaceda what's codecov complaining about here? Overall project looks fine.

cocojoe:
Can probably just lower the patch threshold to say 50%, the project is the most valuable one for me.

PR #150, opened 2018-03-12
Improved error handling for CredentialsManagerExceptions.

donatienthorez:
When getting an exception from the CredentialsManager, now it is possible to easily check which error comes from it thanks to the CredentialsError enum.


lbalmaceda:
Hey thanks for the contribution 🎉 
I already thought about adding this in the past. I prefer an exception class like [AuthenticationException](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java) where we assign error codes (in that particular case, we obtain them from the response body), and then we have getters for the booleans. 
Depending on the context where you called a given method, the getters amount you need to call to assert the actual mistake are short. i.e. you'd never get a `NO_AVAILABLE_REFRESH_TOKEN` error when calling `manager.saveCredentials()`.

If you want to update it we need to agree on the error codes: We probably don't want to have one `code` for each message. But the `message` can remain as a descriptive text to display to the user/dev 👍 

- Exception creation would require both the code and the message.
- Exception class will have getters for the code and the message
- Exception class will have helper getters for each code. i.e. `boolean isMissingRefreshToken()`
- Enums means bad performance on android. You can use `@IntDef` for the same purpose _if required_.

What do you think?

donatienthorez:
Hey,

Thanks for the feedback. Please check the last commit, I changed the enum to a StringDef and put it inside the CredentialsManagerException. I created an enum and something like a factory before in order to be able to create a error only with some specific messages and leave the general constructor of exception private.

The constructor now takes in parameters an CredentialsManagerError StringDef and optionally an throwable. That way it is not possible to instantiate a CredentialsManagerException with a message which is not in the StringDef.

About the codes, the only bad things I see is that, if you have a constructor as in the AuthenticationException with code and message, you could create an error which has a message and a code that does not belong to each other. 

I would suggest to create a method getCode() which check for the message and returns the corresponding code string. We could call this method either In the constructor and assign it to a String variable either each time we need to check the code (like inside an isErrorEncryption() method)

For the codes I would suggest to have
- error.credentials : NO_CREDENTIALS_SET  and  INVALID_CREDENTIALS
- error.renewing_credentials : NO_AVAILABLE_REFRESH_TOKEN, RENEW_CREDENTIALS_ERROR
- error.authentication_challenge : AUTHENTICATION_CHALLENGE_FAILED
- error.encryption : ENCRYPTION_ERROR, DECRYPTION_ERROR

For all the methods I think it is a good idea.
We could have all the other methods inside the ManagerException class like isInvalidCredentials, isErrorEncryption() etc...

Let me know what you think about the different codes above,

PR #151, opened 2018-03-13
Update to okhttp3 3.10.0

donatienthorez:
Notes : 
- There is still a test that failed that has been commented out line 212 in the BaseRequestTest.java. I cannot find out how to throw an IOException since calling twice response.body().string() does not throw an IOException anymore but just returns an empty string. 

Do you have an idea how to do it ?

- I had to put the test to run on sdk 23 or 25 because : 

 it got stucked when running the tests. The MockServer takeRequest method does not work well. 
https://github.com/square/okhttp/issues/3664

or

I have an AssertionError with clean text. More info here : https://github.com/square/okhttp/issues/2591

What do you think about the rest of the pull request ?

### DIFF
--- a/auth0/build.gradle
+++ b/auth0/build.gradle
@@ -71,8 +71,8 @@ dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile 'com.android.support:appcompat-v7:25.3.1'
     compile 'com.android.support:customtabs:25.3.1'
-    compile 'com.squareup.okhttp:okhttp:2.7.5'
-    compile 'com.squareup.okhttp:logging-interceptor:2.7.5'
+    compile 'com.squareup.okhttp3:okhttp:3.9.1'
+    compile 'com.squareup.okhttp3:logging-interceptor:3.9.1'
     compile 'com.google.code.gson:gson:2.8.2'
     compile 'com.auth0.android:jwtdecode:1.1.1'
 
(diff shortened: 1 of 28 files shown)


friederbluemle:
It would be great to get this merged soon. Right now, the Auth0 SDK adds almost 5000 methods to the dex method count - it could be a lot less if the underlying networking lib matches what most apps are using (of course it would be best to have a little third-party dependencies in a library as possible, but that's beyond the scope of this PR).
Thanks.

cocojoe:
Thanks, great contribution @donatienthorez, I'll have this reviewed. @lbalmaceda

lbalmaceda:
This was already discussed before and there's an existing PR already #83. The problem is that things like [this](https://github.com/auth0/Auth0.Android/pull/151/files#diff-8645ce5636d53d051d1bc6d651d5a637R150) are breaking changes, since users will need to update their callbacks signature to match the new class that OkHttp uses. I plan to support it if a new major version of this library is to be made. But unfortunately, not today.

TWiStErRob:
@lbalmaceda the [`auth0Domain` in Manifest](https://github.com/auth0/Auth0.Android/commit/284d9c80d8e6266efc00a06e9497b6876dd5d865#diff-ddec178edb3411a34da2d7f164d81ae7) was also a breaking change, this change would have a big benefit for app sizes. This is the last library we have that uses OkHttp 2 (2 years old version!).

TWiStErRob:
Bump, 1.5 years passed, OkHttp 4 is alive and well.

lbalmaceda:
@TWiStErRob I know, apologies for that. We will use the latest OKHttp version on the next Auth0.Android major. But while this major is around, that will not happen.

PR #152, opened 2018-03-19
Disable HTTP 2 protocol on OkHttp client

lbalmaceda:
OkHttp 2.7.5 (latest) doesn't seem to handle very well the `HTTP 2` protocol. While this is fixed on the new OkHttp 3 that implies a breaking change we can't deal with right now.

The present fix is removing the `HTTP 2` protocol from the client allowed protocols list. Since the [default protocols](https://github.com/square/okhttp/blob/okhttp_27/okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java#L54-L55) include both `HTTP 1.1` and `SPDY 3`, those remain enabled. 

This will fix https://github.com/auth0/Lock.Android/issues/457

### DIFF
--- a/auth0/src/test/java/com/auth0/android/request/internal/OkHttpClientFactoryTest.java
+++ b/auth0/src/test/java/com/auth0/android/request/internal/OkHttpClientFactoryTest.java
@@ -1,10 +1,9 @@
 package com.auth0.android.request.internal;
 
-import android.app.Activity;
-
 import com.squareup.okhttp.ConnectionSpec;
 import com.squareup.okhttp.Interceptor;
 import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Protocol;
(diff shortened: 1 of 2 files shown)


auth0/src/test/java/com/auth0/android/request/internal/OkHttpClientFactoryTest.java
```diff
@@ -37,25 +35,34 @@
 @Config(constants = com.auth0.android.auth0.BuildConfig.class, sdk = 21, manifest = Config.NONE)
 public class OkHttpClientFactoryTest {
 
-    private Activity activity;
```
lbalmaceda: unused

cocojoe:
We should update to new Major OKHTTP one day.

lbalmaceda:
Yes. The next SDK will definitely run over OkHttp 3

PR #173, opened 2018-07-19
Fix service handling when custom tabs are not available

lbalmaceda:
This will check the service was bound before attempting to unbind it


billgockeler:
@lbalmaceda good morning, any chance you can get this merged this morning? We're shooting to get this fix released today - thx.

PR #177, opened 2018-07-31
Add Id Tokens verification (RS256 only)

lbalmaceda:
This PR adds Id Token signature and claims verification for an "Auth0 ID_Token" by using the tenant's Public Key hosted in the `.well-known` url. This only affects customers that have set the `Auth0` instance to be "oidcConformant" [using this method](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/Auth0.java#L176).

### Details
Tokens are verified automatically by the Authentication API Client as soon as an Id Token is included in the server response. The JWKS file, once obtained, is kept in memory until the API client instance gets recycled. There are no methods exposed to the user for them to verify a token on their own.

#### Token verification steps

1. Token has signature. Throw exception otherwise.
2. Token has been signed using the `RS256` algorithm. ~Deem token as "valid" otherwise, to maintain legacy non-oidc behavior (support for HS256 signed tokens).~ Throw exception otherwise.
3. Using the key id from the token, fetch the JWK and verify the signature.
4. Verify claims following [this instructions](https://auth0.com/docs/tokens/id-token#validate-the-claims). This is:
   - Token is not expired
   - Issuer matches the expected one
   - Expected Audience is included

Auth0 server always includes the claims in (4) on the Id Tokens. ~Because I don't really know if they can be removed from an _Auth0 Rule_ context, I'm only checking them if they are both A) included in the token payload and B) explicitly set in the verifier "values to expect".~ Failure to comply with (4) will deem the token invalid.

#### How did I test this?
I added tests for the classes included in this PR. I also improved the tests for the networking of the library overall. For the particular case when the user is using a Custom Domain, no extra steps need to be done. The very same JWKS file is accessible from both URLs: https://auth.brucke.club/.well-known/jwks.json and https://brucke.auth0.com/.well-known/jwks.json


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.java
```diff
@@ -136,6 +136,9 @@ private AuthenticationAPIClient(Auth0 auth0, RequestFactory factory, OkHttpClien
         this.client = clientFactory.createClient(auth0.isLoggingEnabled(), auth0.isTLS12Enforced());
         this.gson = gson;
         this.factory = factory;
+        JwtVerifier verifier = new JwtVerifier(new JwkProvider(auth0.getDomainUrl()));
+        verifier.setExpectedValues(auth0.getDomainUrl(), auth0.getClientId());
+        this.factory.setJwtVerifier(verifier);
```
lbalmaceda: Make the "request factory" use a JWT Verifier

auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
```diff
@@ -126,7 +131,11 @@ public AuthenticationRequest authenticationPOST(HttpUrl url, OkHttpClient client
     }
 
     <T, U extends Auth0Exception> ParameterizableRequest<T, U> createSimpleRequest(HttpUrl url, OkHttpClient client, Gson gson, String method, Class<T> clazz, ErrorBuilder<U> errorBuilder) {
-        return new SimpleRequest<>(url, client, gson, method, clazz, errorBuilder);
+        SimpleRequest<T, U> request = new SimpleRequest<>(url, client, gson, method, clazz, errorBuilder);
+        if ("POST".equals(method) && Credentials.class.equals(clazz)) {
+            request.setJwtVerifier(verifier);
```
lbalmaceda: If this is an "Authentication request", set the verifier. Internally, verification will only happen when an Id Token is present on the Credentials response.

auth0/src/main/java/com/auth0/android/request/internal/RequestFactory.java
```diff
@@ -138,7 +147,9 @@ public AuthenticationRequest authenticationPOST(HttpUrl url, OkHttpClient client
     }
 
     AuthenticationRequest createAuthenticationRequest(HttpUrl url, OkHttpClient client, Gson gson, String method) {
-        return new BaseAuthenticationRequest(url, client, gson, method, Credentials.class);
+        BaseAuthenticationRequest request = new BaseAuthenticationRequest(url, client, gson, method, Credentials.class);
+        request.setJwtVerifier(verifier);
```
lbalmaceda: If this is an "Authentication request", set the verifier. Internally, verification will only happen when an Id Token is present on the Credentials response.

auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -72,10 +78,16 @@ public void onResponse(Response response) throws IOException {
         ResponseBody body = response.body();
         try {
             Reader charStream = body.charStream();
-            T payload = getAdapter().fromJson(charStream);
-            postOnSuccess(payload);
+            T result = getAdapter().fromJson(charStream);
+            if (result instanceof Credentials) {
+                Credentials credentials = (Credentials) result;
+                assertValidIdToken(credentials);
```
lbalmaceda: [ASYNC]  If this is a "credentials response" then verify the id token

auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -98,7 +110,7 @@ public T execute() throws Auth0Exception {
         try {
             response = client.newCall(request).execute();
         } catch (IOException e) {
-            throw new Auth0Exception("Failed to execute request to " + url, e);
+            throw getErrorBuilder().from("Failed to execute request to " + url, new Auth0Exception("Failed to execute request to " + url, e));
```
lbalmaceda: Behaves like other failure scenarios

auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -108,11 +120,31 @@ public T execute() throws Auth0Exception {
         ResponseBody body = response.body();
         try {
             Reader charStream = body.charStream();
-            return getAdapter().fromJson(charStream);
+            T result = getAdapter().fromJson(charStream);
+            if (result instanceof Credentials) {
+                Credentials credentials = (Credentials) result;
+                assertValidIdToken(credentials);
```
lbalmaceda: [SYNC] If this is a "credentials response" then verify the id token

auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -108,11 +120,31 @@ public T execute() throws Auth0Exception {
...
+            }
+            return result;
+        } catch (TokenVerificationException e) {
+            throw getErrorBuilder().from("The received Id Token is not valid", e);
         } catch (IOException e) {
-            throw new Auth0Exception("Failed to parse response to request to " + url, e);
+            Auth0Exception auth0Exception = new Auth0Exception("Failed to parse response to request to " + url, e);
+            throw getErrorBuilder().from("Failed to parse a successful response", auth0Exception);
```
lbalmaceda: Behaves like other failure scenarios

auth0/src/main/java/com/auth0/android/request/internal/SimpleRequest.java
```diff
@@ -108,11 +120,31 @@ public T execute() throws Auth0Exception {
...
         }
     }
+
+    void setJwtVerifier(JwtVerifier verifier) {
+        this.verifier = verifier;
+    }
+
+    private void assertValidIdToken(Credentials credentials) throws TokenVerificationException {
+        String idToken = credentials.getIdToken();
+        if (verifier == null || TextUtils.isEmpty(idToken)) {
```
lbalmaceda: Will only work when verifier is set and id token is present

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
```diff
@@ -95,19 +96,13 @@
...
     private AuthenticationAPI mockAPI;
+    private AuthenticationAPIClient client; //Legacy Behavior enabled Client
 
     @Before
     public void setUp() throws Exception {
         mockAPI = new AuthenticationAPI();
-        final String domain = mockAPI.getDomain();
-        auth0 = new Auth0(CLIENT_ID, domain, domain);
-        client = new AuthenticationAPIClient(auth0);
-        gson = new GsonBuilder().serializeNulls().create();
```
lbalmaceda: All this fields were not required "globally". I moved them into specific tests

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
```diff
@@ -117,14 +112,23 @@ public void tearDown() throws Exception {
...
         verify(factory).setUserAgent("nexus-5x");
     }
 
+    @Test
+    public void shouldUseJwtVerifier() throws Exception {
+        Auth0 auth0 = new Auth0(CLIENT_ID, DOMAIN);
+        RequestFactory factory = mock(RequestFactory.class);
+        OkHttpClientFactory clientFactory = mock(OkHttpClientFactory.class);
+        new AuthenticationAPIClient(auth0, factory, clientFactory);
+        verify(factory).setJwtVerifier(any(JwtVerifier.class));
```
lbalmaceda: A JWT Verifier should always be set

auth0/src/main/java/com/auth0/android/authentication/JwkProvider.java
```diff
@@ -0,0 +1,184 @@
...
+        this.url = url;
+    }
+
+    @VisibleForTesting
+    URL getURL() {
+        return this.url;
+    }
+
+    private static URL createUrl(String domain) {
+        String safeUrl = domain.startsWith("http") ? domain : "https://" + domain;
```
joshcanhelp: This would fail if the domain name starts with `http`:

https://http2.github.io/

auth0/src/main/java/com/auth0/android/authentication/JwtVerifier.java
```diff
@@ -0,0 +1,89 @@
...
+
+        byte[] content = String.format("%s.%s", parts[0], parts[1]).getBytes(Charset.defaultCharset());
+        byte[] signature = Base64.decode(parts[2], Base64.URL_SAFE | Base64.NO_WRAP | Base64.NO_PADDING);
+
+        //Step 2: Verify the signature using the Public Key
+        try {
+            String kid = jwt.getHeader().get("kid");
+            PublicKey publicKey = keyProvider.getPublicKey(kid);
+            if (!verifySignature(publicKey, content, signature)) {
+                throw new SignatureException("The signature does not match");
```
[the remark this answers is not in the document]
joshcanhelp: > it's pretty clear what this is supposed to do

I agree but in terms of debugging this, something like:

```
The signature does not match the public key.
```

... or thereabouts. Just something more to tell someone what it's checking that did not match.

auth0/src/main/java/com/auth0/android/authentication/JwkProvider.java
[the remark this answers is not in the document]
lbalmaceda: I get what you mean now. I'll push a fix, but will delete the "fail on invalid domain url" test as will stop making sense now 🆗

auth0/src/main/java/com/auth0/android/authentication/JwtVerifier.java
```diff
@@ -0,0 +1,89 @@
...
+    JwtVerifier(@NonNull KeyProvider keyProvider) {
+        this.keyProvider = keyProvider;
+    }
+
+    public void verify(String token) throws TokenVerificationException {
+        Log.d(TAG, "Verifying received token");
+
+        //Step 1: Assert token contains signature and is of type RS256
+        String[] parts = token.split("\\.");
+        if (parts.length != 3 || TextUtils.isEmpty(parts[2])) {
```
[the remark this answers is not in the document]
lbalmaceda: Something different than 3 parts is not a JWT.. But if it is one and the third one is empty, there's no signature. I really don't want threat that other case as special. "There's no signature present on the string you passed that I could verify" sounds too much 😂. This is called internally by our implementation. This is not open for users to use directly, so there's no chance that anything different than an ID Token gets passed here.

cocojoe:
@lbalmaceda can you add usage example to description please. Saw this in one of @joshcanhelp PRs and was nice to see.

lbalmaceda:
@cocojoe from the PR description:

> **Tokens are verified automatically** by the Authentication API Client as soon as an Id Token is included in the server response.

No need to do anything.

esarafianou:
LGTM

PR #185, opened 2018-10-01
Replace compile with implementation 

rosariopfernandes:
Most recent gradle versions use `implementation` instead of `compile`


lbalmaceda:
Already done by our plugin https://github.com/auth0/Auth0.Android/pull/188

PR #186, opened 2018-10-02
Change target sdk to 28 and use latest Gradle plugin

lbalmaceda:
This is probably the root cause of "androidX" not working properly when this SDK was a dependency


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Thu Sep 15 13:57:35 ART 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
```
friederbluemle: When using the `wrapper` task to update the Gradle wrapper version, please consider doing

    ./gradlew wrapper  --gradle-version 4.10.2 --distribution-type all

The _default_ for Android projects is the `all` variant, which provides additional documentation and sources helpful when working on Gradle related stuff. Since it's the default it's also much more likely to be already downloaded in a Gradle cache. It is a _one-time, machine-wide_ download.

lbalmaceda: Thanks! I'll update it

lbalmaceda:
I'm still fighting against jacoco to get the coverage reports exported as xml. Something on this oss plugin update broke it and I'm trying to figure out if it was because of the gradle version or the android plugin version. Once that's done I'll merge this. cc @friederbluemle

friederbluemle:
@lbalmaceda The Android Gradle plugin 3.2.0 changed the output directory for javac compiles classes files (required for JaCoCo). It used to be under `<build_dir>/intermediates/classes/[...]`, now it is `<build_dir>/intermediates/javac/[...]` (directory depends on build variant).

Currently, the path is hard coded inside of [AndroidLibraryPlugin.groovy](https://github.com/auth0/oss-library-gradle-plugin/blob/master/src/main/groovy/com/auth0/gradle/oss/AndroidLibraryPlugin.groovy#L254)

You probably need to change

    dir: "${project.buildDir}/intermediates/classes/${name}",

to

    dir: "${project.buildDir}/intermediates/javac/${name}",

PR #197, opened 2018-10-13
Fix Typo & Make All Badges Look Similar

CodingFrieda:


CodingFrieda:
Sorry, it should be fixed now. 😅

PR #203, opened 2018-10-28
WIP: Add Android P support for SecureCredentialsManager

### LINKED ISSUE #187: Android P support
I continuously see the following warning in the logcat logs when using Auth0 with the SecureCredentialsManager in Android 9 (P)

`W/KeyStore: KeyStore exception
    android.os.ServiceSpecificException:  (code 7)
        at android.os.Parcel.createException(Parcel.java:1956)
        at android.os.Parcel.readException(Parcel.java:1910)
        at android.os.Parcel.readException(Parcel.java:1860)
        at android.security.IKeystoreService$Stub$Proxy.get(IKeystoreService.java:786)
        at android.security.KeyStore.get(KeyStore.java:195)
        at android.security.keystore.AndroidKeyStoreSpi.engineGetCertificateChain(AndroidKeyStoreSpi.java:118)
        at java.security.KeyStoreSpi.engineGetEntry(KeyStoreSpi.java:484)
        at java.security.KeyStore.getEntry(KeyStore.java:1560)
        at com.auth0.android.authentication.storage.CryptoUtil.getRSAKeyEntry(CryptoUtil.java:88)
        at com.auth0.android.authentication.storage.CryptoUtil.RSADecrypt(CryptoUtil.java:166)
        at com.auth0.android.authentication.storage.CryptoUtil.getAESKey(CryptoUtil.java:204)
        at com.auth0.android.authentication.storage.CryptoUtil.decrypt(CryptoUtil.java:225)
        at com.auth0.android.authentication.storage.SecureCredentialsManager.continueGetCredentials(SecureCredentialsManager.java:208)
        at com.auth0.android.authentication.storage.SecureCredentialsManager.getCredentials(SecureCredentialsManager.java:175)`

The root cause is probably the same as in this other library´s issue: https://github.com/adorsys/secure-storage-android/issues/30

TheGamer007:
Fixes #187 
Please don't merge while WIP. Only writing proper tests is remaining.


auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -85,7 +86,15 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
             keyStore.load(null);
             if (keyStore.containsAlias(KEY_ALIAS)) {
                 //Return existing key
-                return (KeyStore.PrivateKeyEntry) keyStore.getEntry(KEY_ALIAS, null);
+
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
lbalmaceda: move all this block to a new method since you're using the same one a few lines below.
```java
private KeyStore.PrivateKeyEntry getKeyEntryForAlias(KeyStore keyStore, String alias) {
}
```

I suggested that name since the current method this is in is already named similarly.

lbalmaceda:
Looping you in @sebaslogen since you originally reported the issue. Would you mind testing this PR out and giving us feedback? Thanks

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
TheGamer007: @lbalmaceda I named the method `getKeyEntryCompat` since it is a wrapper that is only checking the API level to call the appropriate methods.

I figured out a method that gets all the tests to run, and gives no warnings on the emulator (I don't have an Android P device, but I believe it should have the same results). However, it is a bit hackish, so I wanted to run it by you to see if it's ok.

I added a condition to check if the `PrivateKey` is `null` at [this line](https://github.com/auth0/Auth0.Android/pull/203/files#diff-84689992633df9060c3e7070601bcf26R156).  Since I couldn't mimic an empty PrivateKey item despite looking into the source codes of some of the classes, I am simply falling back and using `getEntry` as before in such cases.

From what I observed, the `PrivateKey` object is `null` at that point only during the tests. On devices, the `getRSAKeyEntry` method generates the entry if the alias is not present in the keystore.

TheGamer007:
All the tests in `CryptoUtilTest` completed successfully. The failed tests on Circle CI seem unrelated to my changes.

sebaslogen:
Thanks for the fix. I compiled the library and manually replaced in my project the Auth0 dependency, the app works correctly  and I don't see anymore the previous exception `KeyStore exception android.os.ServiceSpecificException:  (code 7)`
So this has my approval.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -145,6 +146,22 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
...
 
+    private KeyStore.PrivateKeyEntry getKeyEntryCompat(KeyStore keyStore)
+            throws KeyStoreException, NoSuchAlgorithmException, UnrecoverableEntryException {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            //Following code is for API 28+
+            PrivateKey privateKey = (PrivateKey) keyStore.getKey(KEY_ALIAS, null);
+            Certificate certificate = keyStore.getCertificate(KEY_ALIAS);
+
+            if (privateKey == null) {
+                return (KeyStore.PrivateKeyEntry) keyStore.getEntry(KEY_ALIAS, null);
```
lbalmaceda: this line will have the same value as `privateKey`. Seems like this will always return null, am I correct?

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -145,6 +146,22 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
         }
     }
 
+    private KeyStore.PrivateKeyEntry getKeyEntryCompat(KeyStore keyStore)
+            throws KeyStoreException, NoSuchAlgorithmException, UnrecoverableEntryException {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
lbalmaceda: Please refactor this to look more like the project's code-style. Read the in-line comments I left you:

```java
if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
   return (KeyStore.PrivateKeyEntry) keyStore.getEntry(KEY_ALIAS, null);
}

//Following code is for API 28+
PrivateKey privateKey = (PrivateKey) keyStore.getKey(KEY_ALIAS, null);
if (privateKey == null) {
    //FIXME: See my review comment below
    return (KeyStore.PrivateKeyEntry) keyStore.getEntry(KEY_ALIAS, null); 
}
//I also moved this since it's used only for the "new" call
Certificate certificate = keyStore.getCertificate(KEY_ALIAS);
return new KeyStore.PrivateKeyEntry(privateKey, new Certificate[]{certificate});
```

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -145,6 +146,22 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
...
+            //Following code is for API 28+
+            PrivateKey privateKey = (PrivateKey) keyStore.getKey(KEY_ALIAS, null);
+            Certificate certificate = keyStore.getCertificate(KEY_ALIAS);
```
lbalmaceda: this is only used for the `new KeyStore.PrivateKeyEntry(privateKey, new Certificate[]{certificate});` call, we could avoid calling this if the if-case doesn't require it.

lbalmaceda:
Hi @sebaslogen. Do you need further guidance on resolving the last comments I left? Please, don't hesitate asking. I'm planning a release this week and would be nice if I can include this. Thanks!

TheGamer007:
@lbalmaceda sorry, I wasn't able to check this out yet due to work. I'll get started on this tomorrow and update the PR as soon as possible.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
TheGamer007: When the `privateKey` is null, the constructor can't be used to create a PrivateKeyEntry object and it throws a NullPointerException. But the method here returns a PrivateKeyEntry object that contains a null privateKey, and an empty certificate chain without throwing any exceptions or warnings.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
TheGamer007: Refactored to match.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
TheGamer007: @lbalmaceda Are you referring to the `Certificate` object? Since the `privateKey` check was moved before it, now it is only created when needed. Or did you mean I should create it inline as `new Certificate[] {keystore.getCertificate(KEY_ALIAS)}`?

TheGamer007:
@lbalmaceda I've made the changes you requested, just confused about that last one.

MHausBermel:
@lbalmaceda Any idea as to when an official release will come with these changes?

lbalmaceda:
@MHausBermel I'll cut a release today with these changes and then start looking into the issues reported (see open issues) around this class.

Issue #85, opened 2017-03-22
How to add user metadata on custom signup?


kiaanpillay:
Hi, I am currently trying to add a phone number as custom metadata in my sign up as follows

`    Map<String, Object> metadata = new HashMap<>();
    metadata.put("contact_num", phoneNumber);

  
            authentication
                    .signUp(email, pass,  "Username-Password-Authentication")
                    .setScope("openid role")
                    .addSignUpParameters(metadata)
                    .start(new BaseCallback<Credentials, AuthenticationException>() {`

However this never seems to work. What am I doing wrong? Is this possible?

lbalmaceda:
Hi @kiaanpillay 
The Auth API differentiates creating a user from signing up. The first one just creates the user while the second also logs him in. As the last one has 2 requests, you can set additional parameters for each of them by calling `addAuthenticationParameters` and `addSignUpParameters`.
To set the `user_metadata` you'll need to use that same key when adding the sign up parameters: 

```java
//Define your custom fields
Map<String, Object> metadata = new HashMap<>();
metadata.put("contact_num", phoneNumber);

//Define the sign up parameters, adding the user_metadata
Map<String, Object> params = new HashMap<>();
params.put("user_metadata", metadata);

//Add the sign up parameters to the request
authentication
                .signUp(email, pass,  "Username-Password-Authentication")
                .setScope("openid role")
                .addSignUpParameters(params);
```

Cheers

PR #204, opened 2018-11-05
Improve JavaDoc how to set user metadata

Syex:
Hi,

I encountered the same issue as asked in #85. Currently I think this behavior is not obvious at all and thought it would be wise to document how to actually set user metadata with a sign up request.


Syex:
Sorry for the delay, missed that you replied after that time 😉 I added link to the general documentation of user metadata, as you link to the section `#metadata-restrictions` doesn't seem to exist, at least it doesn't put me anywhere.

PR #210, opened 2018-12-04
Update robolectric to version 4

lbalmaceda:
Update tests to avoid robolectric warnings. See migration guide for those changes here http://robolectric.org/migrating

### DIFF
--- a/auth0/build.gradle
+++ b/auth0/build.gradle
@@ -60,12 +60,6 @@ android {
     lintOptions {
         warning 'InvalidPackage'
     }
-    buildTypes {
-        debug {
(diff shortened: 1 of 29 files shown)


auth0/build.gradle
```diff
@@ -60,12 +60,6 @@ android {
     lintOptions {
         warning 'InvalidPackage'
     }
-    buildTypes {
-        debug {
-            //Helps tests. buildTypes values are not included in the merged manifest
```
lbalmaceda: Added as a replacement a new AndroidManifest file scoped only for the tests

PR #218, opened 2019-01-15
Delete keys and stored Credentials on unrecoverable use cases

lbalmaceda:
Whenever the Keystore stored keys are deemed unrecoverable (e.g. when the Android Lock Screen method or password is changed), delete the Keystore keys in order to regenerate them the next time the method is called; and throw a new exception so that the Credentials Manager is aware and can either retry the operation in the case of 'write' or remove the existing encrypted Credentials in the case of 'read'.


auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.java
```diff
@@ -148,6 +148,9 @@ public void saveCredentials(@NonNull Credentials credentials) throws Credentials
             storage.store(KEY_CREDENTIALS, encryptedEncoded);
             storage.store(KEY_EXPIRES_AT, expiresAt);
             storage.store(KEY_CAN_REFRESH, canRefresh);
+        } catch (UnrecoverableContentException e) {
+            //If keys were invalidated, a retry will work fine for the "save credentials" use case.
+            saveCredentials(credentials);
```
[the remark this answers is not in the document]
lbalmaceda: I'm certain this should only happen once. But just to be super safe, I'll fail with an exception the user can capture and then, they can retry manually.

TLDR; I removed the recursive call.

auth0/src/main/java/com/auth0/android/authentication/storage/UnrecoverableContentException.java
```diff
@@ -0,0 +1,11 @@
+package com.auth0.android.authentication.storage;
+
+/**
+ * Exception thrown by the {@link CryptoUtil} class whenever the Keys are deemed invalid
+ * and so the content encrypted with them unrecoverable.
+ */
+class UnrecoverableContentException extends CryptoException {
```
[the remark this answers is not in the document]
lbalmaceda: Removed and replaced with a different class.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManagerException.java
```diff
@@ -15,4 +15,14 @@
...
         super(message);
     }
+
+    /**
+     * Returns true when this Android device doesn't support the cryptographic algorithms used
+     * to handle encryption and decryption, false otherwise.
+     *
+     * @return whether this device is compatible with {@link SecureCredentialsManager} or not.
+     */
+    public boolean isDeviceIncompatible() {
```
lbalmaceda: When users receive a `CredentialsManagerException` after calling "save" or "get" credentials methods, they can check whether the exception is recoverable or not by checking this `isDeviceIncompatible` flag.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -117,15 +125,14 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
                         .setEndDate(end.getTime());
 
                 KeyguardManager kManager = (KeyguardManager) context.getSystemService(Context.KEYGUARD_SERVICE);
-                Intent authIntent = null;
                 if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
                     //The next call can return null when the LockScreen is not configured
-                    authIntent = kManager.createConfirmDeviceCredentialIntent(null, null);
-                }
-                boolean keyguardEnabled = Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && kManager.isKeyguardSecure() && authIntent != null;
```
lbalmaceda: removed this check `Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP ` since it's already done a few lines above

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -135,17 +142,55 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
...
+             * We can delete it and let the user retry the operation.
+             *
+             * - IOException:
+             *      Thrown when there is an I/O or format problem with the keystore data.
+             * - UnrecoverableEntryException:
+             *      Thrown when the key cannot be recovered. Probably because it was invalidated by a Lock Screen change.
+             */
             deleteKeys();
-            return getRSAKeyEntry();
+            throw new CryptoException("The existing RSA key pair could not be recovered and has been deleted. " +
```
lbalmaceda: This is the only 2 exceptions scenario where I consider a retry could fix any issue, since I'm deleting any corrupted existing keys if this happens... next time it's called, it will recreate them as explained in the comments.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -162,57 +207,111 @@ public CryptoUtil(@NonNull Context context, @NonNull Storage storage, @NonNull S
...
 
-    //Used to delete recreate the key pair in case of error
+    /**
+     * Removes the AES and RSA keys generated in a previous execution.
+     * Used when we want the next call to {@link #encrypt(byte[])} or {@link #decrypt(byte[])}
+     * to recreate the keys.
+     */
     private void deleteKeys() {
+        storage.remove(KEY_ALIAS);
+        storage.remove(KEY_IV_ALIAS);
```
lbalmaceda: moved up to ensure this is called regardless the keystore calls succeed or not.

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
```diff
@@ -230,31 +329,79 @@ private void deleteKeys() {
...
+     */
+    public byte[] decrypt(byte[] encryptedInput) throws CryptoException, IncompatibleDeviceException {
         try {
             SecretKey key = new SecretKeySpec(getAESKey(), ALGORITHM_AES);
             Cipher cipher = Cipher.getInstance(AES_TRANSFORMATION);
             String encodedIV = storage.retrieveString(KEY_IV_ALIAS);
             if (TextUtils.isEmpty(encodedIV)) {
-                throw new InvalidAlgorithmParameterException("The AES Key exists but an IV was never stored. Try to encrypt something first.");
+                //AES key was JUST generated. If anything existed before, should be encrypted again first.
+                throw new CryptoException("The encryption keys changed recently. You need to re-encrypt something first.", null);
```
lbalmaceda: This is a special scenario where the keys got regenerated but nothing was saved after that.. since we're encrypting with AES in a symmetric manner, the previous stored initialization vector (iv) is no longer related to the now regenerated AES keys. Given that, anything that was already stored cannot be decrypted and should be ignored/removed. The exception message instructs the dev to call "save" credentials first before trying to call "get" again.

README.md
```diff
@@ -629,6 +629,15 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 The `checkAuthenticationResult` method will continue the retrieval of credentials on a successful authentication, and the decrypted credentials will be delivered to the callback passed on the `getCredentials` call.
 
 
+#### Handling exceptions
+
+In the event that something happened while trying to save or retrieve the credentials, a `CredentialsManagerException` will be thrown. This are some of the expected failure scenarios:
```
cocojoe: ~This are~ -> These are

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManagerException.java
cocojoe: Why have a method that checks for the negative outcome? Why not `isDeviceCompatible` ?

lbalmaceda: Thought about that but discarded it. The thing is this is already an exception (something that went wrong). The "catastrophic" reason of the error could be a `DeviceIncompatibleException`, thus the name of the method.

It wouldn't make sense to receive this exception and check "is device compatible" IMO. Maybe we could rename this method (not the exception class) to `isDeviceNotCompatible`. Thoughts?

auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
lbalmaceda: Separated this into 2 separate calls, since some scenarios required only the AES keys to be recreated

lbalmaceda:
TBH this scenarios are tricky and not easy to catch unless you try different combinations of configuration on an android device. Should be fine now but if something comes up, will open a follow up PR 👌 Thanks for the review @cocojoe

PR #233, opened 2019-05-10
Add method to specify scope when renewing a token in CredentialsManager

### LINKED ISSUE #232: Retrieving the User Profile using a refreshed token returns an empty object and errors
### Description
After retrieving user credentials for the first time, we can successfully use the returned token to get the User Profile using the `AuthenticationAPIClient.userInfo(token)` method. However, after the token is refreshed, calling the `userInfo` method with the refreshed token results in the following exception:

```
 com.google.gson.JsonParseException: user profile json is not a valid json object
        at com.auth0.android.request.internal.UserProfileDeserializer.deserialize(UserProfileDeserializer.java:21)
        at com.auth0.android.request.internal.UserProfileDeserializer.deserialize(UserProfileDeserializer.java:17)
        at com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
        at com.google.gson.TypeAdapter.fromJson(TypeAdapter.java:260)
        at com.auth0.android.request.internal.SimpleRequest.execute(SimpleRequest.java:111)
```
The reason for this is because the `/userInfo` endpoint is returning an empty JSON object (`{}`) back and this cannot be parsed. We have additionally found that an empty JSON object is being returned because the refreshed token does not contain the scopes necessary to fetch the user's profile (`openid` and `profile`). Using the CredentialsManager, there's also no way to pass in these scopes when fetching/refreshing the token:
```java
void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
```

Is there any way we can pass in these scopes or enable the refreshed token to have the same scope as the original token?

### Environment
- Android Auth0 1.15.1
- Java 8

### Reproduction

Call `CredentialsManager.getCredentials()` after a token has expired and then use the token to call `AuthenticationAPIClient.userInfo(token).execute()`. This will result in a JsonParseException

tomrom95:
### Changes
- Added a method to CredentialsManager to optionally include the scope when renewing a token
```java
public void getCredentials(@Nullable final String scope,
                           @NonNull final BaseCallback<Credentials, CredentialsManagerException> callback)
```
- Added unit test for new method

### References
- Closes #232 which causes renewed tokens to not have the right scope for accessing the user profile

### Testing
Added unit test, `./gradlew test` passes

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -1,17 +1,19 @@
 package com.auth0.android.authentication.storage;
 
-import static android.text.TextUtils.isEmpty;
-
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
+
 import com.auth0.android.authentication.AuthenticationAPIClient;
 import com.auth0.android.authentication.AuthenticationException;
 import com.auth0.android.callback.AuthenticationCallback;
 import com.auth0.android.callback.BaseCallback;
 import com.auth0.android.request.ParameterizableRequest;
 import com.auth0.android.result.Credentials;
+
 import java.util.Date;
+
+import static android.text.TextUtils.isEmpty;
 
 /**
  * Class that handles credentials and allows to save and retrieve them.
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -1,17 +1,17 @@
 package com.auth0.android.authentication.storage;
 
-import android.support.annotation.NonNull;
-import android.support.annotation.VisibleForTesting;
+import static android.text.TextUtils.isEmpty;
 
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.VisibleForTesting;
 import com.auth0.android.authentication.AuthenticationAPIClient;
 import com.auth0.android.authentication.AuthenticationException;
 import com.auth0.android.callback.AuthenticationCallback;
 import com.auth0.android.callback.BaseCallback;
+import com.auth0.android.request.ParameterizableRequest;
 import com.auth0.android.result.Credentials;
-
 import java.util.Date;
-
-import static android.text.TextUtils.isEmpty;
 
 /**
  * Class that handles credentials and allows to save and retrieve them.
@@ -64,19 +64,31 @@ public class CredentialsManager {
      * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
      */
     public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
+        getCredentials(null, callback);
+    }
+
+    /**
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with {@link CredentialsManagerException} if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
+     */
+    public void getCredentials(@Nullable final String scope,
+                               @NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
         String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
         final String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
         String idToken = storage.retrieveString(KEY_ID_TOKEN);
         String tokenType = storage.retrieveString(KEY_TOKEN_TYPE);
         Long expiresAt = storage.retrieveLong(KEY_EXPIRES_AT);
-        String scope = storage.retrieveString(KEY_SCOPE);
+        String storedScope = storage.retrieveString(KEY_SCOPE);
 
         if (isEmpty(accessToken) && isEmpty(idToken) || expiresAt == null) {
             callback.onFailure(new CredentialsManagerException("No Credentials were previously set."));
             return;
         }
         if (expiresAt > getCurrentTimeInMillis()) {
-            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), scope));
+            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), storedScope));
             return;
         }
         if (refreshToken == null) {
@@ -84,7 +96,11 @@ public class CredentialsManager {
             return;
         }
 
-        authClient.renewAuth(refreshToken).start(new AuthenticationCallback<Credentials>() {
+        ParameterizableRequest<Credentials, AuthenticationException> renewRequest = authClient.renewAuth(refreshToken);
+        if (scope != null) {
+            renewRequest = renewRequest.addParameter("scope", scope);
+        }
+        renewRequest.start(new AuthenticationCallback<Credentials>() {
             @Override
             public void onSuccess(Credentials fresh) {
                 //RefreshTokens don't expire. It should remain the same
@@ -112,8 +128,8 @@ public class CredentialsManager {
         Long expiresAt = storage.retrieveLong(KEY_EXPIRES_AT);
 
         return !(isEmpty(accessToken) && isEmpty(idToken) ||
-                expiresAt == null ||
-                expiresAt <= getCurrentTimeInMillis() && refreshToken == null);
+            expiresAt == null ||
+            expiresAt <= getCurrentTimeInMillis() && refreshToken == null);
     }
 
     /**
--- a/auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
+++ b/auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
@@ -319,6 +319,33 @@ public class CredentialsManagerTest {
         assertThat(retrievedCredentials.getScope(), is("newScope"));
     }
 
+    @Test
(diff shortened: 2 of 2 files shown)


tomrom95:
Friendly ping :) . Would love to merge this in to unblock some of the issues in our apps.

cocojoe:
Thanks for the PR 👍 
@lbalmaceda can you take a look please.

tomrom95:
Ping. We're maintaining a forked version of this, which isn't ideal because we want to keep any latest security updates

lbalmaceda:
If not specified otherwise in the token renewal request, the renewed access token should be granted the same scope it was granted on the first authentication request. Is your app using OIDC conformant mode? Does the initial request ask for the userinfo audience?

tomrom95:
@lbalmaceda App is using OIDC conformant  mode and the original request asked for the userinfo audience. We had this same issue in our iOS app, so that's why I've jumped to this solution.

lbalmaceda:
I couldn't reproduce this locally. You can copy-paste a code similar to the one I'll leave below and try it on your side.


```java
//Login method
WebAuthProvider.init(auth0)
                .withScheme("demo")
                .withAudience(String.format("https://%s/userinfo", getString(R.string.com_auth0_domain)))
                .withScope("openid offline_access")
                .start(this, webCallback);

//webCallback's success method
apiClient.userInfo(credentials.getAccessToken()).start(new BaseCallback<UserProfile, AuthenticationException>() {
                @Override
                public void onSuccess(UserProfile payload) {
                    Log.e("TEST", "FIRST USERINFO CALL SUCCEEDED");
                }

                @Override
                public void onFailure(AuthenticationException error) {
                    Log.e("TEST", "FIRST USERINFO CALL FAILED");
                }
            });

            apiClient.renewAuth(credentials.getRefreshToken()).start(new BaseCallback<Credentials, AuthenticationException>() {
                @Override
                public void onSuccess(Credentials payload) {
                    Log.e("TEST", "TOKEN RENEWED");
                    apiClient.userInfo(payload.getAccessToken()).start(new BaseCallback<UserProfile, AuthenticationException>() {
                        @Override
                        public void onSuccess(UserProfile payload) {
                            Log.e("TEST", "SECOND USERINFO CALL SUCCEEDED");
                        }

                        @Override
                        public void onFailure(AuthenticationException error) {
                            Log.e("TEST", "SECOND USERINFO CALL FAILED");
                        }
                    });
                }

                @Override
                public void onFailure(AuthenticationException error) {
                    Log.e("TEST", "FAILED TO RENEW TOKEN");
                }
            });
```

If you enable the network logs you will see that the userinfo received in both cases is complete.

tomrom95:
@lbalmaceda So we think the issue is coming from our Auth0 rules. We have a rule that returns custom scopes based on the user metadata. E.g: 
```javascript

```

This is working for the original request, by appending our custom scopes to the requested scopes (i.e. offline_access, profile, email, openid). It looks like it may not be working for renewed token requests, because it only returns our custom scopes.

Do you know of a way to get the scopes from the initial request from a rule during a token renewal?

lbalmaceda:
I think I've never tried adding scopes via a rule and then attempting to refresh the token. But sounds like you are doing your own authorization logic. I know [this feature](https://auth0.com/docs/authorization/guides/how-to) was made recently available, you might want to take a look. But given that this is not an SDK issue I suggest you contact https://support.auth0.com to be assisted on your specific use case.

tomrom95:
Submitted a ticket here and hopefully we can resolve https://community.auth0.com/t/how-to-get-originally-granted-scopes-when-renewing-token-in-auth0-rules/25694 .

In the mean time, I don't think this necessarily disqualifies the PR. Since it's possible to specify modified scopes in a token renewal request, it should also probably be possible to specify these scopes to the CredentialsManager (which, in turn, makes a renewal request). This is the same behavior in the Swift code: https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L120 , which we leverage.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -64,27 +66,43 @@ public void saveCredentials(@NonNull Credentials credentials) {
...
     public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
+        getCredentials(null, callback);
+    }
+
+    /**
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with {@link CredentialsManagerException} if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
```
lbalmaceda: javadoc for the new param

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -64,27 +66,43 @@ public void saveCredentials(@NonNull Credentials credentials) {
...
-        String scope = storage.retrieveString(KEY_SCOPE);
+        String storedScope = storage.retrieveString(KEY_SCOPE);
 
         if (isEmpty(accessToken) && isEmpty(idToken) || expiresAt == null) {
             callback.onFailure(new CredentialsManagerException("No Credentials were previously set."));
             return;
         }
         if (expiresAt > getCurrentTimeInMillis()) {
-            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), scope));
+            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), storedScope));
```
lbalmaceda: if the scope asked for in this introduced method call cannot be satisfied with the existing token, this is going to deliver an invalid token back to the user in the onSuccess call. Similarly, if the refresh token originally obtained was not granted those scopes before, trying to refresh it with the new requested set of scopes will also fail.

Maybe the safest thing to do is split the scope strin (which is going to contain scopes separated by a space.) Then check if ALL of the new requested scopes are included in the original scope string, in any order. If at least one of the new requested scopes is not present on the original scope string, then throw a runtime error such as `InvalidArgumentException` as the dev made the mistake of asking a scope he knows it can't be provided.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -64,27 +66,43 @@ public void saveCredentials(@NonNull Credentials credentials) {
...
+    }
+
+    /**
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with {@link CredentialsManagerException} if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
+     */
+    public void getCredentials(@Nullable final String scope,
```
lbalmaceda: I think I'd rather mark it as `@NonNull` and ignore the lint warning on the call above, this way users don't call this method passing null on purpose when the other method is available. Thoughts on this approach?

tomrom95:
@lbalmaceda did you have a chance to discuss this feature with the team? I'll address all the comments when I know that this feature is wanted.

After raising a ticket, there is no doable way to get the initially granted scopes in a rule when refreshing a token. So, if a mobile app ever wants to authenticate with an Auth0 client that is using rules, they'll have to use my proposed API to inject the requested scopes.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
tomrom95: I'll switch this function to have a `@NonNull` scope and then create a helper function that takes in a nullable scope.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
tomrom95: Interesting. But isn't there always the case in the initial token request that you are granted different scopes than you request? In this case, no error is thrown, so I don't think there should be an error thrown here.

lbalmaceda: > But isn't there always the case in the initial token request that you are granted different scopes than you request?
You get the scopes you ask for. Can you explain your rationale here?

lbalmaceda:
@tomrom95  I'm going to ping them again.. but this involves a few teams. I'm first going to check that what you're doing in the rule with the scopes is a supported use case. Can you please share the ticket number you mentioned above?

lbalmaceda:
@tomrom95  I've been digging into your use case. You are setting via a **Rule** the new scope for the access token. That's something doable from a Rule 👍 However, you're basing that scope value from the scope the client sent in the request. You access this from the `context.request.uri.scope` variable inside the rule code. That parameter is missing in the refresh call, as this SDK is not sending it. When the scope you want to obtain is the same as the original login request, by skipping sending the scope parameter in the refresh token, the access token would be granted the same scope.

The problem is that in the rule you don't have access to the "original scope" granted to that first log in. You need to base your solution in something different that doesn't depend on that. Is not that you opened this PR because you want to "reduce the scope" in the renew call. You want to send the same scope requested originally for your rule's code to work.

I think the use case you're trying to achieve may be better suited for a [Role-Based Access Control](https://auth0.com/docs/authorization/concepts/rbac) approach. Have you tried that, and if not, why?

tomrom95:
@lbalmaceda you are correct that by not setting the scope parameter during when a token is being renewed, the original scopes are automatically added.

We modified our rule code so that if you are renewing a token and don't request new scopes, the original scopes are returned. We've also submitted a ticket to document this behavior in the Auth0 docs.

Since there may not be a need for this PR for my usecase anymore, I'll leave it up to you to close or open.

lbalmaceda:
I'm happy that you could fix it by updating the rule's code. 
Regarding this particular PR we've decided not to include this option right now but might revisit in the future and will consider your code if that happens. 
Thanks for the effort so far 🙏 !

Issue #234, opened 2019-05-24
Auth0 renewAuth fails to propagate http status code to AuthenticationException in onFailure


dcostinett:
### Description

I’m working on an Android app and getting an unexpected response from the renewAuth API call when I try to refresh a token. The process works fine up until the token gets invalidated on the back-end. When I get a response with the http response code 403 Forbidden with body:

{“error”:“invalid_grant”,“error_description”:“Unknown or invalid refresh token.”}

what I want to do is check the AuthenticationException object in the callback and I expect that that error object will have the statusCode set to 403 to match the http response status code, but the value it holds is 0:

### Prerequisites

[x] I have checked the documentation for this library [Add a link].

[x] I have checked the [Auth0 Community](https://community.auth0.com/) for related posts.
I posted this issue there too.

[x] I have checked for related or duplicate [Issues](https://github.com/auth0/Auth0.Android/issues) and [PRs](https://github.com/auth0/Auth0.Android/pulls).

[x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md).

[x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Environment

Please provide the following:

- Version of Auth0.Android used: 1.15.1
- Version of Android being used: API 21 to API 28 (Android 5.0 Lollipop to Q). App built targetting API 28, using AndroidX.
- Additional packages that might be affecting your instance: Retrofit 2.4.0, OkHttp 3.14.1

### Reproduction

With a valid Auth0 refresh token we call the renewAuth SDK method and get a 403 http status code 
![charles_clip](https://user-images.githubusercontent.com/2453774/58343404-bcdd4780-7e07-11e9-80ff-38f9a8d23a05.png)
response from the service which fires the onFailure method of the callback. Examine the AuthenticationException object's statusCode property and I expect to see the status code there match the http status code. But it appears the AuthenticationException(String, String) constructor is called rather than the AuthenticationException(

Please include:

- Specific devices affected
Galaxy S9, Pixel 2 XL
- Log files or stacktraces (redact/remove sensitive information)
- Snippet of the code you're running (redact/remove sensitive information)
    override fun refreshToken(): Single<Boolean> {
        return Single.create<Boolean> {
            authClient.renewAuth(profileService.getRefreshToken())
[...]

PR #235, opened 2019-05-27
Expose NetworkErrorException when request fails due to networking

lbalmaceda:
### Changes

Currently, all network requests performed by OkHttp will either fail or succeed. 
- When they succeed, `onSuccess` is called and the requests have an HTTP status code set (e.g. 200, 401, 500, etc).
- When they fail, `onFailure` is called with an `IOException` instance. This scenario means _network error_ (e.g. invalid host, timeout, etc). 

Normally one doesn't need to capture this scenario as OkHttp is known for retrying the request a few times before reporting it as a failure. And additionally the Android SDK provides methods to check for connectivity before you perform a request. But in case you want a more granular control, this PR ensures that the exception is not treated as "another Auth0 error" along with the API errors (success branch above ☝️) and now has its own dedicated exception thrown. **(I don't really like the idea of adding a new exception class, but here we go..)**

Now users after calling `request.execute()` or `request.start({callback})` can capture that `Auth0Exception` and ask if it's an instance of `NetworkErrorException` class by using the helper method `isNetworkError()`.

```java
//request.execute() throws "error"
if (error.isNetworkError()){
  //handle network error differently 
}
```

### References

Closes https://github.com/auth0/Auth0.Android/issues/234.

### Testing

### Checklist


lbalmaceda:
@joshcanhelp Normally a few exceptions per project should be enough but I couldn't come up with another approach that didn't require me to create a new class. Because this is a different use case, I don't want to re-use the already generic enough class "Auth0Exception" for this scenario. 

The callback receives the same exception type, an Auth0Exception. Internally, the "cause" of this exception changes from `IOException` to `NetworkErrorException`. This must be done this way because the `IOException` is also thrown in the scenario where the JSON parse step of a successful request fails. The only way to distinguish both scenarios is by changing the cause of the error in one of them.

This is fine. The code on those checking for the exception's cause will still _compile_ and now they have the chance to check for this separate scenario.

lbalmaceda:
@joshcanhelp ended up adding that helper method that at first I decided not to. This should make things even simpler to users.

Issue #155, opened 2018-04-03
Can't logout from Webclient


Viked:
Hello,
I am using this code to start auth process :

    Auth0 auth0 = new Auth0(AppConfig.auth0clientId, AppConfig.auth0domain);
auth0 = Utils.getAuth0();
        WebAuthProvider.init(auth0)
               .withParameters(parameters)
                .withScheme(getString(R.string.scheme))
                .withScope(Config.Auth0.scopeValue)
                .start(this, callback);

Parameters contains "prompt": "login"

And login/registration flow works perfectly. 
When try to logout - it automatically login me to the app.

Used just email/password auth.
SDK version 'com.auth0.android:auth0:1.12.0'

Viked:
Clean token from local storage. Should I use some methods from sdk for the log out?
I tried to use some method to clear credentials from sdk, but it was not help.

lbalmaceda:
@Viked If you are using the `WebAuthProvider`, you're using a browser app to authenticate (i.e. Chrome). This SDK **cannot** clear the browser cookies for you, and tbh I don't think any other android app can, only the browser itself. Unless the browser apps expose an API to do so from an external app. That being said, how you handle the "logged in" state is up to you. You can implement your own logic to store and remove credentials from memory or `SharedPreferences` file or use our [Credential Managers](https://github.com/auth0/Auth0.Android#credentials-manager) to simplify that. But there's no "logging out" method available on this SDK.

Calling the `WebAuthProvider` passing `prompt=login` should force show the Lock widget (Universal Login) and thus, making the user input their credentials or click a social button manually before returning to the application. This is _if you are not_ passing a valid `connection` parameter too, since that will skip showing Lock. That's all it does the `prompt` parameter and that's how it should work. If that feature works or not, that's not part of this SDK either so you'll have to reach our support team at https://support.auth0.com to fill the issue.

I you find an SDK issue feel free to reopen this issue or create a new one.
Cheers

dustin-graham:
I read the docs here: https://auth0.com/docs/logout

and I believe it would be an easy fix to add logout functionality to this library. It simply needs to open a new ChromeCustomTab to the logout url for the user's tenant. I've tested this myself and it works perfectly. It takes a very small amount of code to accomplish. There is even support for a redirect and I believe that could be leveraged for several other special handling features as well. 

[...]

PR #245, opened 2019-06-25
Add WebAuth Logout feature

lbalmaceda:
### Changes

This PR adds a static builder to launch a call to the `/v2/logout` endpoint. 
Users can customize the scheme of the `returnTo` URL,  and the _Chrome Custom Tabs_ customization options, although TBH the process is quite fast and you barely notice the browser was open.

#### Usage
```java
//log out method
Auth0 auth0 = getAccount();
WebAuthProvider.logout(auth0)
                .withScheme("demo")
                .start(this, logoutCallback);

//somewhere in the code
private VoidCallback logoutCallback = new VoidCallback() {
    @Override
    public void onFailure(Auth0Exception error) {
        //Browser app not found or logout canceled, check error message.
    }

    @Override
    public void onSuccess(Void payload) {
        //Logged out!
    }
};
```

#### Callback
The call can 
- **succeed:**  when the browser invokes the `returnTo` URL and in turn, re-opens the user application.
- **fail:** 
   - when there is no browser app available, 
   - or the user closes the browser manually. This last scenario is also triggered if the log out URL is not whitelisted in the application or tenant settings.

The log out URL will always include the `client_id` parameter. So the `returnTo` URL **must be whitelisted in the application** allowed logout URLs section on the dashboard. More info in auth0 docs.

### References

- Docs: https://auth0.com/docs/logout/guides/logout-auth0
- Issues: https://github.com/auth0/Auth0.Android/issues/155 

### Testing

Added a bunch of tests for the introduced changes. 

**Will probably rename** some of the previous tests as now we share 2 flows (auth / logout) on the same `WebAuthProvider` class. (On a different PR)

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,10 +31,10 @@ Those using this library from version `1.14.0` and up should start targeting lat
 
 ```groovy
 apply plugin: 'com.android.application'
- android {
+android {
     //...
 }
- dependencies {
+dependencies {
     implementation ('com.auth0.android:lock:1.14.1'){
         exclude group: 'com.android.support', module: 'appcompat-v7'
         exclude group: 'com.android.support', module: 'customtabs'
@@ -93,7 +93,7 @@ Passwordless authentication *cannot be used* with this flag set to `true`. For m
 
 ### Authentication with Universal Login
 
-First go to the [Auth0 Dashboard](https://manage.auth0.com/#/applications) and go to your application's settings. Make sure you have in *Allowed Callback URLs* a URL with the following format:
+First go to the [Auth0 Dashboard](https://manage.auth0.com/#/applications) and go to your application's settings. Make sure you have in **Allowed Callback URLs** a URL with the following format:
 
 ```
 https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/callback
@@ -176,7 +176,7 @@ Finally, authenticate by showing the **Auth0 Universal Login**:
 ```java
 //Configure and launch the authentication
 WebAuthProvider.init(account)
-                .start(MainActivity.this, authCallback);
+    .start(MainActivity.this, authCallback);
 
 // Define somewhere in the code the callback
 AuthCallback authCallback = new AuthCallback() {
@@ -222,14 +222,14 @@ If you don't plan to use the _Web Authentication_ feature you will still be prom
 If you've followed this documents' configuration steps you've noticed that the default scheme used in the Callback URI is `https`. This works best for Android API 23 or newer if you're using [Android App Links](https://auth0.com/docs/applications/enable-android-app-links), but in previous Android versions this may show the intent chooser dialog prompting the user to choose either your application or the browser. You can change this behaviour by using a custom unique scheme so that the OS opens directly the link with your app.
 
 1. Update the `auth0Scheme` Manifest Placeholder on the `app/build.gradle` file or update the intent-filter declaration in the `AndroidManifest.xml` to use the new scheme.
-2. Update the allowed callback urls in your [Auth0 Dashboard](https://manage.auth0.com/#/applications) application's settings.
+2. Update the **Allowed Callback URLs** in your [Auth0 Dashboard](https://manage.auth0.com/#/applications) application's settings.
 3. Call `withScheme()` passing the custom scheme you want to use.
 
 
 ```java
 WebAuthProvider.init(account)
-                .withScheme("myapp")
-                .start(MainActivity.this, authCallback);
+    .withScheme("myapp")
+    .start(MainActivity.this, authCallback);
 ```
 
 
@@ -237,8 +237,8 @@ WebAuthProvider.init(account)
 
 ```java
 WebAuthProvider.init(account)
-                .withConnection("twitter")
-                .start(MainActivity.this, authCallback);
+    .withConnection("twitter")
+    .start(MainActivity.this, authCallback);
 ```
 
 #### Use Code grant with PKCE
@@ -248,8 +248,8 @@ WebAuthProvider.init(account)
 
 ```java
 WebAuthProvider.init(account)
-                .useCodeGrant(true)
-                .start(MainActivity.this, authCallback);
+    .useCodeGrant(true)
+    .start(MainActivity.this, authCallback);
 ```
 
 #### Specify audience
@@ -258,8 +258,8 @@ The snippet below requests the "userinfo" audience in order to guarantee OIDC co
 
 ```java
 WebAuthProvider.init(account)
-                .withAudience("https://{YOUR_AUTH0_DOMAIN}/userinfo")
-                .start(MainActivity.this, authCallback);
+    .withAudience("https://{YOUR_AUTH0_DOMAIN}/userinfo")
+    .start(MainActivity.this, authCallback);
 ```
 
 > Replace `{YOUR_AUTH0_DOMAIN}` with your actual Auth0 domain (i.e. `mytenant.auth0.com`).
@@ -268,8 +268,8 @@ WebAuthProvider.init(account)
 
 ```java
 WebAuthProvider.init(account)
-                .withScope("openid profile email")
-                .start(MainActivity.this, authCallback);
+    .withScope("openid profile email")
+    .start(MainActivity.this, authCallback);
 ```
 
 > The default scope used is `openid`
@@ -278,8 +278,8 @@ WebAuthProvider.init(account)
 
 ```java
 WebAuthProvider.init(account)
-                .withConnectionScope("email", "profile", "calendar:read")
-                .start(MainActivity.this, authCallback);
+    .withConnectionScope("email", "profile", "calendar:read")
+    .start(MainActivity.this, authCallback);
 ```
 
 
@@ -288,14 +288,14 @@ WebAuthProvider.init(account)
 If the device where the app is running has a Custom Tabs compatible Browser, a Custom Tab will be preferred for the authentication flow. You can customize the Page Title visibility and the Toolbar color by using the `CustomTabsOptions` class.
  
 ```java
- CustomTabsOptions options = CustomTabsOptions.newBuilder()
+CustomTabsOptions options = CustomTabsOptions.newBuilder()
     .withToolbarColor(R.color.ct_toolbar_color)
     .showTitle(true)
     .build();
  
-  WebAuthProvider.init(account)
-                  .withCustomTabsOptions(options)
-                  .start(MainActivity.this, authCallback);
+WebAuthProvider.init(account)
+    .withCustomTabsOptions(options)
+    .start(MainActivity.this, authCallback);
 ```
 
 
@@ -305,7 +305,7 @@ To log the user out and clear the SSO cookies that the Auth0 Server keeps attac
 
 Make sure to [revisit that section](#authentication-with-universal-login) to configure the Manifest Placeholders if you still cannot authenticate successfully. The values set there are used to generate the URL that the server will redirect the user back to after a successful log out.
 
-In order for this redirection to happen, you must copy the *Allowed Callback URLs* value you added for authentication into the *Allowed Logout URLs* field in your [application settings](https://manage.auth0.com/#/applications). Both fields should have an URL with the following format:
+In order for this redirection to happen, you must copy the **Allowed Callback URLs** value you added for authentication into the **Allowed Logout URLs** field in your [application settings](https://manage.auth0.com/#/applications). Both fields should have an URL with the following format:
 
 
 ```
@@ -316,12 +316,12 @@ Remember to replace `{YOUR_APP_PACKAGE_NAME}` with your actual application's pac
 
 
 
-Initialize the provider, this time calling the static method `clearSession`.
+Initialize the provider, this time calling the static method `logout`.
 
 ```java
 //Configure and launch the log out
-WebAuthProvider.clearSession(account)
-                .start(MainActivity.this, logoutCallback);
+WebAuthProvider.logout(account)
+    .start(MainActivity.this, logoutCallback);
 
 //Declare the callback that will receive the result
 BaseCallback logoutCallback = new BaseCallback<Void, Auth0Exception>() {
@@ -341,32 +341,32 @@ BaseCallback logoutCallback = new BaseCallback<Void, Auth0Exception>() {
 The callback will get invoked when the user returns to your application. If this is the result of being redirected back by the server, that would be considered a success. There are some scenarios in which this can fail:
 * When there is no browser application that can open a URL. The cause of the exception will be an instance of `ActivityNotFoundException`.
 * When the user closes the browser manually, e.g. by pressing the back key on their device.
-* When the `returnTo` URL is not whitelisted in your application settings.
+* When the `returnTo` URL is not found in the **Allowed Logout URLs** in your Auth0 application settings.
 
 
 #### Customize the Custom Tabs UI
 
-Similarly to when you authenticated your users, for log out you can also customize the styling of the Custom Tabs browser. However, do note the browser is briefly shown to the user.
+If the device where the app is running has a Custom Tabs compatible Browser, a Custom Tab will be preferred for the logout flow. You can customize the Page Title visibility and the Toolbar color by using the `CustomTabsOptions` class.
 
 ```java
- CustomTabsOptions options = CustomTabsOptions.newBuilder()
+CustomTabsOptions options = CustomTabsOptions.newBuilder()
     .withToolbarColor(R.color.ct_toolbar_color)
     .showTitle(true)
     .build();
 
-  WebAuthProvider.clearSession(account)
-                  .withCustomTabsOptions(options)
-                  .start(MainActivity.this, logoutCallback);
+WebAuthProvider.logout(account)
+    .withCustomTabsOptions(options)
+    .start(MainActivity.this, logoutCallback);
 ```
 
 
-#### Changing the scheme
-The scheme used can be changed as well. This configuration will probably match what you've done for the [authentication setup](#a-note-about-app-deep-linking).
+#### Changing the Return To URL scheme
+This configuration will probably match what you've done for the [authentication setup](#a-note-about-app-deep-linking).
 
 ```java
-WebAuthProvider.clearSession(account)
-                .withScheme("myapp")
-                .start(MainActivity.this, logoutCallback);
+WebAuthProvider.logout(account)
+    .withScheme("myapp")
+    .start(MainActivity.this, logoutCallback);
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -339,7 +339,6 @@ BaseCallback logoutCallback = new BaseCallback<Void, Auth0Exception>() {
 
 
 The callback will get invoked when the user returns to your application. If this is the result of being redirected back by the server, that would be considered a success. There are some scenarios in which this can fail:
-* When the domain is not [correctly set up](#usage) in the Auth0 instance. The cause of the exception will be an instance of `IllegalArgumentException`.
 * When there is no browser application that can open a URL. The cause of the exception will be an instance of `ActivityNotFoundException`.
 * When the user closes the browser manually, e.g. by pressing the back key on their device.
 * When the `returnTo` URL is not whitelisted in your application settings.
--- a/README.md
+++ b/README.md
@@ -303,9 +303,16 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
 
 To log the user out and clear the SSO cookies that the Auth0 Server keeps attached to your browser app, you need to call the [logout endpoint](https://auth0.com/docs/api/authentication?#logout). This can be done is a similar fashion to how you authenticated before: using the `WebAuthProvider` class.
 
-Note that the Manifest Placeholders must be set in order for the redirection to work, just like you were required to do so for the authentication flow. Make sure to [revisit that section](#authentication-with-universal-login) first if you still cannot authenticate successfully.
+Make sure to [revisit that section](#authentication-with-universal-login) to configure the Manifest Placeholders if you still cannot authenticate successfully. The values set there are used to generate the URL that the server will redirect the user back to after a successful log out.
 
-In addition, you need to whitelist in the [Auth0 Dashboard](https://manage.auth0.com/#/applications) the "return to" URL for your application settings, right into the *Allowed Logout URLs* The value of this URL is the very same of the Allowed Callback URL you've whitelisted when setting up the authentication. Remember, you must **save** the settings in the dashboard for them to apply.
+In order for this redirection to happen, you must copy the *Allowed Callback URLs* value you added for authentication into the *Allowed Logout URLs* field in your [application settings](https://manage.auth0.com/#/applications). Both fields should have an URL with the following format:
+
+
+```
+https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/callback
+```
+
+Remember to replace `{YOUR_APP_PACKAGE_NAME}` with your actual application's package name, available in your `app/build.gradle` file as the `applicationId` value.
 
 
 
@@ -316,7 +323,7 @@ Initialize the provider, this time calling the static method `clearSession`.
 WebAuthProvider.clearSession(account)
                 .start(MainActivity.this, logoutCallback);
 
-// Define somewhere in the code the callback
+//Declare the callback that will receive the result
 BaseCallback logoutCallback = new BaseCallback<Void, Auth0Exception>() {
     @Override
     public void onFailure(Auth0Exception exception) {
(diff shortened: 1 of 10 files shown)


auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,107 @@
...
+
+        /**
+         * Whether to also clear the session from the identity provider when possible or not.
+         *
+         * @param federatedLogout true to clear the session on the identity provider's site.
+         * @return the current builder instance
+         */
+        public LogoutBuilder useFederatedLogout(@NonNull boolean federatedLogout) {
+            if (federatedLogout) {
+                this.values.put(KEY_FEDERATED, "1");
```
lbalmaceda: empty or null values won't get added. That's why the `1` here.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,107 @@
...
+            return this;
+        }
+
+        /**
+         * Whether to also clear the session from the identity provider when possible or not.
+         *
+         * @param federatedLogout true to clear the session on the identity provider's site.
+         * @return the current builder instance
+         */
+        public LogoutBuilder useFederatedLogout(@NonNull boolean federatedLogout) {
```
lbalmaceda: Feedback on the method name?

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,107 @@
...
+            return this;
+        }
+
+        /**
+         * Request the user session to be cleared. When successful, the callback will get invoked
+         *
+         * @param context  to run the log out
+         * @param callback to invoke when log out is successful
+         */
+        public void start(Context context, BaseCallback<Void, Auth0Exception> callback) {
```
lbalmaceda: Feedback on the method name?
e.g. start or launch? The auth one uses "start"

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,107 @@
...
+                Throwable cause = new IllegalArgumentException("Auth0 logout URL not properly set. This can be related to an invalid domain.");
+                final Auth0Exception ex = new Auth0Exception("Cannot perform web log out", cause);
+                callback.onFailure(ex);
+                return;
+            }
+
+            if (!hasBrowserAppInstalled(context.getPackageManager())) {
+                Throwable cause = new ActivityNotFoundException("No Browser application installed.");
+                final Auth0Exception ex = new Auth0Exception("Cannot perform web log out", cause);
+                callback.onFailure(ex);
```
lbalmaceda: I wanted to differentiate this immediate failures by providing 2 different classes on the Cause.

-When the domain value was wrong and the logout url couldn't be generated, throw an `Auth0Exception` with `IllegalArgumentException` cause.
-When there is no browser app to handle this, throw an `Auth0Exception` with `ActivityNotFoundException` cause.


This is here to avoid creating a new type (class) of exception. Users could differentiate the cause by doing `instanceof` on it.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -269,12 +362,6 @@ Builder withPKCE(PKCE pkce) {
             return this;
         }
 
-        @VisibleForTesting
-        static boolean hasBrowserAppInstalled(@NonNull PackageManager packageManager) {
```
lbalmaceda: Moved down on as it's now shared between both builders

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -324,7 +411,22 @@ public void start(@NonNull Activity activity, @NonNull AuthCallback callback) {
...
 
     /**
-     * Initialize the WebAuthProvider instance with an account. Additional settings can be configured
+     * Initialize the WebAuthProvider instance for logging out the user using an account. Additional settings can be configured
+     * in the LogoutBuilder, like setting the federated parameter.
+     *
+     * @param account to use for authentication
+     * @return a new Builder instance to customize.
+     */
+    public static LogoutBuilder clearSession(@NonNull Auth0 account) {
```
lbalmaceda: Feedback on the method name? The comments below will get removed later.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -110,13 +116,19 @@ public void setUp() throws Exception {
...
 
+    //** ** ** ** ** **  **//
+    //** ** ** ** ** **  **//
+    //** LOG IN  FEATURE **//
+    //** ** ** ** ** **  **//
+    //** ** ** ** ** **  **//
+
     @SuppressWarnings("deprecation")
     @Test
     public void shouldInitWithAccount() throws Exception {
```
lbalmaceda: will probably rename all these old test methods with a prefix for "auth" or "login" e.g. `shouldInitLoginWithAccount`

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -133,18 +145,24 @@ public void shouldInitWithContext() throws Exception {
...
                 .start(activity, callback);
 
-        assertNotNull(WebAuthProvider.getInstance());
+        assertNotNull(WebAuthProvider.getManagerInstance());
     }
 
     @SuppressWarnings("deprecation")
     @Test
-    public void shouldNotResumeWhenNotInit() throws Exception {
+    public void shouldNotResumeWithRequestCodeWhenNotInit() throws Exception {
```
lbalmaceda: separated into 2 test cases

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -1324,7 +1342,8 @@ public void shouldCalculateExpiresAtDateOnResumeAuthentication() throws Exceptio
         WebAuthProvider.init(account)
                 .useCodeGrant(false)
                 .start(activity, callback, REQUEST_CODE);
-        WebAuthProvider.getInstance().setCurrentTimeInMillis(CURRENT_TIME_MS);
+        OAuthManager managerInstance = (OAuthManager) WebAuthProvider.getManagerInstance();
```
lbalmaceda: This is the only part where I need to cast the instance to a particular manager. Because the `LogoutManager` doesn't use (and doesn't need) to handle the system Clock. Again: These managers are package internal classes.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
jimmyjames: `logout` seems more natural. To remain consistent with authentication, what about `initLogout`?

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,86 @@
...
+        /**
+         * Request the user session to be cleared. When successful, the callback will get invoked
+         *
+         * @param context  to run the log out
+         * @param callback to invoke when log out is successful
+         */
+        public void start(Context context, BaseCallback<Void, Auth0Exception> callback) {
+            managerInstance = null;
+            if (account.getLogoutUrl() == null) {
+                Throwable cause = new IllegalArgumentException("Auth0 logout URL not properly set. This can be related to an invalid domain.");
```
joshcanhelp: Unless this is common practice throughout the library, `IllegalArgumentException` says to me that I passed something into this method incorrectly so I would be looking at the `context` and `callback` instead of the `account` in the constructor. May something like:

> An Auth0 logout URL could not be created. Please check the account settings passed when initializing this class.

... or something to that effect. If it would only fail because of a missing/invalid domain, just say that explicitly. Assuming that this message is not shown to the user.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,86 @@
...
+                final Auth0Exception ex = new Auth0Exception("Cannot perform web log out", cause);
+                callback.onFailure(ex);
+                return;
+            }
+
+            if (!hasBrowserAppInstalled(context.getPackageManager())) {
+                Throwable cause = new ActivityNotFoundException("No Browser application installed.");
```
joshcanhelp: > No Browser application installed.

Consider:

> No browser is installed on this device.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
joshcanhelp: `clearSession` says to me that this method clears the session. But it's actually `start` that does the logout. Maybe `buildLogout` then `executeLogout`? Feels like a language convention more than anything. 

`WebAuthProvider.buildLogout(account).executeLogout(MainActivity.this, logoutCallback);`

auth0/src/test/java/com/auth0/android/provider/LogoutManagerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    BaseCallback<Void, Auth0Exception> callback;
+
+    @Before
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void shouldNotHaveCustomTabsOptionsByDefault() throws Exception {
+        LogoutManager manager = new LogoutManager(account, callback, "https://auth0.com/callback");
```
joshcanhelp: I wouldn't use `auth0.com` URLs for anything that would be app-configured.

auth0/src/main/java/com/auth0/android/provider/LogoutManager.java
```diff
@@ -0,0 +1,87 @@
...
+        addClientParameters(parameters);
+        Uri uri = buildLogoutUri();
+
+        AuthenticationActivity.authenticateUsingBrowser(context, uri, ctOptions);
+    }
+
+    @Override
+    boolean resume(AuthorizeResult result) {
+        if (result.isCanceled()) {
+            Auth0Exception exception = new Auth0Exception("The user closed the browser app and the log out was cancelled.");
```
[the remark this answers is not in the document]
jimmyjames: Is the user closing the browser app the only thing that could trigger this? If so looks good, though maybe change "log out" to "logout".

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -110,13 +116,19 @@ public void setUp() throws Exception {
         prepareBrowserApp(true, null);
     }
 
+    //** ** ** ** ** **  **//
```
joshcanhelp:
```
                                    __________
                           ________|          |________
                          |       /   ||||||   \       |
                          |     ,'              `.     |
                          |   ,'                  `.   |
                          | ,'   ||||||||||||||||   `. |
                          ,'  /____________________\  `.
                         /______________________________\
                        |                                |
                        |                                |
                        |                                |
                        |________________________________|
     _ Seal _             |____________________------__|

              ,----------------------------------------------------,
              | [][][][][]  [][][][][]  [][][][]  [][__]  [][][][] |
              |                                                    |
              |  [][][][][][][][][][][][][][_]    [][][]  [][][][] |
              |  [_][][][][][][][][][][][][][ |   [][][]  [][][][] |
              | [][_][][][][][][][][][][][][]||     []    [][][][] |
              | [__][][][][][][][][][][][][__]    [][][]  [][][]|| |
              |   [__][________________][__]              [__][]|| |
              `----------------------------------------------------'
```

~ [Credit](http://www.ascii-art.de/ascii/c/computer.txt)

auth0/src/main/java/com/auth0/android/provider/LogoutManager.java
```diff
@@ -0,0 +1,87 @@
...
+            callback.onFailure(exception);
+        } else {
+            callback.onSuccess(null);
```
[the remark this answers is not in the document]
jimmyjames: Feels a little weird to require the callback handler to implement success handler using `Void`. I believe typically the Void type is used for reflection purposes. What's the downside to adding a new Callback class - wouldn't it be pretty small and simple?

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
jimmyjames: I think `start` is fine, and it's consistent.

auth0/src/main/java/com/auth0/android/provider/LogoutManager.java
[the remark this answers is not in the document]
lbalmaceda: @joshcanhelp This is an exception passed to the callback. Dev decides what to do when they receive the `onFailure` call with this error instance: e.g. if they want to display the very same message or display something different, like a retry dialog.

@jimmyjames  Now that `federated` support was removed, the only scenario here is when the user closes the browser manually. This can also be triggered if the log out URL is not whitelisted in the application or tenant settings. Because the user would need to manually exit that page, that would count as a "cancel" but TBH, this is dev error and should be catch before shipping the app.

auth0/src/main/java/com/auth0/android/provider/LogoutManager.java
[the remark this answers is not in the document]
lbalmaceda: Yes, it's used because of generics. That way a single callback class can handle different payload types and different exception types.

Yes, I can add a new class but I'm trying to use what we already have shipped. The only downside is another public class. The upside is not having to specify the types with the diamond symbol. However we come up with a good name for a callback that returns nothing on success and throws an Auth0Exception otherwise (our generic exception type) I don't mind adding it.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,107 @@
...
+        private final Auth0 account;
+        private final Map<String, String> values;
+        private String scheme;
+        private CustomTabsOptions ctOptions;
+
+        LogoutBuilder(Auth0 account) {
+            this.account = account;
+
+            //Default values
+            this.scheme = "https";
```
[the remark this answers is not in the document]
lbalmaceda: I forgot to update the description and this comment. I've removed the federated param support for now as it's not a good experience for the public client users.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
lbalmaceda: I like what @jimmyjames suggested to be consistent with the already existing `init` method that is used on authentication. Since both said "logout" I'll go with `initLogout` instead of `clearSession`.

lbalmaceda: I've discussed this with some devs here. They suggested we could deprecate the existing `init` method used for authentication, and add a new one that delegates on that called `login` (just like the new SPA SDK does), and call this `logout` instead of `initLogout`. That sounds like a sweeter alternative!

README.md
```diff
@@ -299,6 +299,71 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
 ```
 
 
+### Clearing the session
+
+To log the user out and clear the SSO cookies that the Auth0 Server keeps attached to your browser app, you need to call the [logout endpoint](https://auth0.com/docs/api/authentication?#logout). This can be done is a similar fashion to how you authenticated before: using the `WebAuthProvider` class.
+
+Note that the Manifest Placeholders must be set in order for the redirection to work, just like you were required to do so for the authentication flow. Make sure to [revisit that section](#authentication-with-universal-login) first if you still cannot authenticate successfully.
```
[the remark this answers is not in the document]
lbalmaceda: What this is trying to say is that the Manifest Placeholders that the user was told to set up for authentication, are required to be present for this to work. The intent of this paragraph is to be a reminder for the user to set up that value. They will definitely do it because:
- gradle won't compile their app if they are using this SDK and not setting them up in the build.gradle file.
- probably they are already using this for Authentication. Logout is just an addition!

This is where I link them, the first time they are mentioned in the readme https://github.com/auth0/Auth0.Android#authentication-with-universal-login. Do you think this paragraph should be written in a different way @joshcanhelp ?

README.md
```diff
@@ -299,6 +299,71 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
...
+
+In addition, you need to whitelist in the [Auth0 Dashboard](https://manage.auth0.com/#/applications) the "return to" URL for your application settings, right into the *Allowed Logout URLs* The value of this URL is the very same of the Allowed Callback URL you've whitelisted when setting up the authentication. Remember, you must **save** the settings in the dashboard for them to apply.
```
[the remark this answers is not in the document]
lbalmaceda: I've rewritten this part.

README.md
```diff
@@ -299,6 +299,71 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
...
+    @Override
+    public void onSuccess(@NonNull Void payload) {
+        //succeeded!
+    }
+};
+```
+
+
+The callback will get invoked when the user returns to your application. If this is the result of being redirected back by the server, that would be considered a success. There are some scenarios in which this can fail:
+* When the domain is not [correctly set up](#usage) in the Auth0 instance. The cause of the exception will be an instance of `IllegalArgumentException`.
```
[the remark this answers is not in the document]
lbalmaceda: I think it's fine as it is. Here "cause" refers to an actual property of the exception https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#getCause()

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
lbalmaceda: Trying to find a justification on why this is checked here, I found that `Auth0` class already does check the domain has a proper URL syntax here https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/Auth0.java#L103. As you can see, it's throwing right there when you create a new instance of `Auth0` if the domain is not valid. I will remove the checks here and in the other authentication builder -> https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java#L290

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
lbalmaceda: This is for an android device context. Will that keep the meaning that it's referring to an android browser application not being installed for this to be run on that?

auth0/src/test/java/com/auth0/android/provider/LogoutManagerTest.java
lbalmaceda: This is just a test case and that's some valid url string. Should I use something different?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
lbalmaceda: Going to do this on a different PR

auth0/src/main/java/com/auth0/android/provider/LogoutManager.java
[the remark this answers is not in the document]
joshcanhelp: For the exception text:

> The user closed the browser app so the logout was cancelled.

README.md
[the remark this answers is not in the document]
joshcanhelp: Nope, this is fine with me, thank you for the explanation.

README.md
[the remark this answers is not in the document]
joshcanhelp: Gotcha, my misunderstanding.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
joshcanhelp: I don't know that it had that meaning to begin with ...

auth0/src/test/java/com/auth0/android/provider/LogoutManagerTest.java
joshcanhelp: In the spirit of "tests as documentation," I would have this URL be something similar to what a developer would use. In this case, they would not be using an `auth0.com` domain (or, typically, an `http/s` scheme).

README.md
```diff
@@ -299,6 +299,77 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
 ```
 
 
+### Clearing the session
+
+To log the user out and clear the SSO cookies that the Auth0 Server keeps attached to your browser app, you need to call the [logout endpoint](https://auth0.com/docs/api/authentication?#logout). This can be done is a similar fashion to how you authenticated before: using the `WebAuthProvider` class.
+
+Make sure to [revisit that section](#authentication-with-universal-login) to configure the Manifest Placeholders if you still cannot authenticate successfully. The values set there are used to generate the URL that the server will redirect the user back to after a successful log out.
+
+In order for this redirection to happen, you must copy the *Allowed Callback URLs* value you added for authentication into the *Allowed Logout URLs* field in your [application settings](https://manage.auth0.com/#/applications). Both fields should have an URL with the following format:
```
joshcanhelp: `**` for **bold** (standard docs style for "calling a thing out")

README.md
```diff
@@ -299,6 +299,71 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
...
+* When the user closes the browser manually.
+* When the `returnTo` URL is not whitelisted in your application settings.
+
+
+#### Customize the Custom Tabs UI
+
+Similarly to when you authenticated your users, for log out you can also customize the styling of the Custom Tabs browser. However, do note the browser is briefly shown to the user.
+
+```java
+ CustomTabsOptions options = CustomTabsOptions.newBuilder()
```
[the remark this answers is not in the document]
lbalmaceda: Fixed on all the snippets! Good catch

README.md
```diff
@@ -299,6 +299,71 @@ If the device where the app is running has a Custom Tabs compatible Browser, a C
...
+The callback will get invoked when the user returns to your application. If this is the result of being redirected back by the server, that would be considered a success. There are some scenarios in which this can fail:
+* When the domain is not [correctly set up](#usage) in the Auth0 instance. The cause of the exception will be an instance of `IllegalArgumentException`.
+* When there is no browser application that can open a URL. The cause of the exception will be an instance of `ActivityNotFoundException`.
+* When the user closes the browser manually.
+* When the `returnTo` URL is not whitelisted in your application settings.
+
+
+#### Customize the Custom Tabs UI
+
+Similarly to when you authenticated your users, for log out you can also customize the styling of the Custom Tabs browser. However, do note the browser is briefly shown to the user.
```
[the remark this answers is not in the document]
lbalmaceda: Changed to what we currently have for authentication https://github.com/auth0/Auth0.Android#customize-the-custom-tabs-ui

auth0/src/main/java/com/auth0/android/provider/ResumableManager.java
```diff
@@ -0,0 +1,16 @@
+package com.auth0.android.provider;
+
+/**
+ * Internal class, used to generify the handling of different Web Auth flows.
+ * See {@link WebAuthProvider}
+ */
+abstract class ResumableManager {
+
+    /**
+     * Invoked when the result is available
```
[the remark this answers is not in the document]
lbalmaceda: internal class, but OK!

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,86 @@
...
+        /**
+         * Specify a custom Scheme to use on the Return To Uri. Default scheme is 'https'.
+         *
+         * @param scheme to use in the Callback Uri.
+         * @return the current builder instance
+         */
+        public LogoutBuilder withScheme(@NonNull String scheme) {
+            String lowerCase = scheme.toLowerCase();
+            if (!scheme.equals(lowerCase)) {
+                Log.w(TAG, "Please provide the scheme in lowercase and make sure it's the same configured in the intent filter. Android expects the scheme in lowercase");
```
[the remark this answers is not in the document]
lbalmaceda: The same message is currently being used for the authentication builder. I've changed both to:
> Please provide the scheme in lowercase and make sure it's the same configured in the intent filter. Android expects the scheme to be lowercase.

This check is here to warn the developer if they run into callback issues that they have not used a scheme in lowercase. I'm not changing that for them because that would need to be changed not only on this method but also on the android manifest (which I cannot access from the code). So, we just print a friendly warning message for them to be aware they are doing it wrong.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -57,17 +60,86 @@
...
+
+        LogoutBuilder(Auth0 account) {
+            this.account = account;
+
+            //Default values
+            this.scheme = "https";
+        }
+
+        /**
+         * When using a Custom Tabs compatible Browser, apply this customization options.
```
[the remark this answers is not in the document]
lbalmaceda: It's my class wrapping to only expose the options I want the user to change. Internally, it calls the `CustomTabsIntent.Builder` methods -> https://developer.android.com/reference/android/support/customtabs/CustomTabsIntent.Builder

Issue #253, opened 2019-08-01
id token not renewed during getCredentials call unless access token is expired.


camhart:
### Description

Description of the bug or feature request and why it's a problem. Consider including:

I use the id token with my AWS Cognito integration.  I'd like my id tokens to ideally be set with a 1-2 hour expiration.  However I can't currently using the `SecureCredentialsManager.getCredentials` call because that call only checks if the access token is expired when considering if it should renew.  And the access token is required to have an expiration of 24 hours.

### Prerequisites

- [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md).

- [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Environment

Please provide the following:
- Version of Auth0.Android used: 1.18.0
- Version of Android being used: 10

### Reproduction

Drop JWT expiration down to 5 minutes.  Authenticate.  Wait 5 minutes, then make a call to `getCredentials`.  The idToken won't be renewed--no call to apiClient.renewAuth is ever made.

```
private void continueGetCredentials(final BaseCallback<Credentials, CredentialsManagerException> callback) {
        String encryptedEncoded = storage.retrieveString(KEY_CREDENTIALS);
        byte[] encrypted = Base64.decode(encryptedEncoded, Base64.DEFAULT);

        String json;
        try {
            json = new String(crypto.decrypt(encrypted));
        } catch (IncompatibleDeviceException e) {
            callback.onFailure(new CredentialsManagerException(String.format("This device is not compatible with the %s class.", SecureCredentialsManager.class.getSimpleName()), e));
            decryptCallback = null;
            return;
        } catch (CryptoException e) {
            //If keys were invalidated, existing credentials will not be recoverable.
            clearCredentials();
            callback.onFailure(new CredentialsManagerException("A change on the Lock Screen security settings have deemed the encryption keys invalid and have been recreated. " +
                    "Any previously stored content is now lost. Please, try saving the credentials again.", e));
            decryptCallback = null;
            return;
        }
        final Credentials credentials = gson.fromJson(json, Credentials.class);
        if (isEmpty(credentials.getAccessToken()) && isEmpty(credentials.getIdToken()) || credentials.getExpiresAt() == null) {
            callback.onFailure(new CredentialsManagerException("No Credentials were previously set."));
            decryptCallback = null;
            return;
[...]

PR #254, opened 2019-08-26
Update CredentialManager classes to include IDToken expiration

lbalmaceda:
### Changes

Made both classes aware of the ID Token `exp` value (when available) to decide when the whole Credentials object can be renewed.

### References
Closes https://github.com/auth0/Auth0.Android/issues/253

### Testing
Added more unit tests

- [] This change adds integration test coverage

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -52,13 +52,12 @@ public class CredentialsManager {
 
         long expiresAt = credentials.getExpiresAt().getTime();
 
-        if(credentials.getIdToken() != null) {
-
+        if (credentials.getIdToken() != null) {
             JWT idToken = new JWT(credentials.getIdToken());
             Date idTokenExpiresAtDate = idToken.getExpiresAt();
 
             if (idTokenExpiresAtDate != null) {
-                expiresAt = Math.min(idTokenExpiresAtDate.getTime(), credentials.getExpiresAt().getTime());
+                expiresAt = Math.min(idTokenExpiresAtDate.getTime(), expiresAt);
             }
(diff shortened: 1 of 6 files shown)


lbalmaceda:
@stevehobbsdev 
> Having a test suite for JWTDecoder seems redundant to me

The JWT library we use has its own tests, true..  However, because there's a token decoding step inside the process of "saving a token in the manager", I need to be able to test when that step fails or succeeds, given different scenarios. So I'm [opening the constructor](https://github.com/auth0/Auth0.Android/pull/254/files#diff-2651976c12b43b2192543b92b10ed2e4R35) (package-privately) so that the test classes can see it and use that other signature that receives a "JWT decoder" instance [for my tests to mock it](https://github.com/auth0/Auth0.Android/pull/254/files#diff-11c2f568f9696a00c1aa933fc583ed23R77). 

But since my tests are mocking it, I cannot ensure that the "JWT decoding" step does actually what it is supposed to do: decode a real JWT string into a JWT instance. So I'm adding a [simple test](https://github.com/auth0/Auth0.Android/pull/254/files#diff-ba2536945b0f9722380dcee112fbe285R17-R37) for that `JWTDecoder` package-private class in which I don't mock anything but I assert a token string gets decoded into a real JWT class [here](https://github.com/auth0/Auth0.Android/pull/254/files#diff-50ed94412c09cf5cb1bce677360f7ad4R14-R16).

Say I change [this line](https://github.com/auth0/Auth0.Android/pull/254/files#diff-50ed94412c09cf5cb1bce677360f7ad4R15) and I make it return a decoded JWT for a different hardcoded token, and don't make use of the one received as param. If I wouldn't be testing the `JWTDecoder` class then I wouldn't caught that error.

lbalmaceda:
Will release this next week. 🏁

PR #265, opened 2019-11-28
Improve OIDC compliance

lbalmaceda:
### Changes

This update improves the SDK support for OpenID Connect. In particular, it modifies the sign-in verification phase by substituting backchannel based checks with id_token validation where possible.

### Testing

This PR contains some unit tests and integration tests to prove the logic is in place for the different authentication scenarios. **Additional unit tests on a separate PR** will complement the token verification logic.

### Checklist


auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
```diff
@@ -0,0 +1,149 @@
...
+
+import com.auth0.android.jwt.JWT;
+
+import java.util.Calendar;
+import java.util.Date;
+import java.util.List;
+
+import static android.text.TextUtils.isEmpty;
+
+class IdTokenVerifier {
```
lbalmaceda: Class file was taken from mvc-commons implementation. Tests to be added on a separate PR

auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
```diff
@@ -0,0 +1,149 @@
...
+        Integer clockSkew;
+        Date clock;
+
+        Options(@NonNull String issuer, @NonNull String audience, @NonNull SignatureVerifier verifier) {
+            this.issuer = issuer;
+            this.audience = audience;
+            this.verifier = verifier;
+        }
+
+        void setNonce(@Nullable String nonce) {
```
lbalmaceda: we use the setters for setting the values, but then here we access them directly. Should we remove the setters as well?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -128,40 +136,115 @@ boolean resume(AuthorizeResult result) {
         try {
             assertNoError(values.get(KEY_ERROR), values.get(KEY_ERROR_DESCRIPTION));
             assertValidState(parameters.get(KEY_STATE), values.get(KEY_STATE));
-            if (parameters.containsKey(KEY_RESPONSE_TYPE) && parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_ID_TOKEN)) {
```
lbalmaceda: nonce check moved into ID token verification logic

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -128,40 +136,115 @@ boolean resume(AuthorizeResult result) {
...
-                //Finish Code Exchange
-                pkce.getToken(values.get(KEY_CODE), new AuthCallback() {
-                    @Override
-                    public void onFailure(@NonNull Dialog dialog) {
-                        callback.onFailure(dialog);
+        final Date expiresAt = !values.containsKey(KEY_EXPIRES_IN) ? null : new Date(getCurrentTimeInMillis() + Long.valueOf(values.get(KEY_EXPIRES_IN)) * 1000);
+        final Credentials frontChannelCredentials = new Credentials(values.get(KEY_ID_TOKEN), values.get(KEY_ACCESS_TOKEN), values.get(KEY_TOKEN_TYPE), values.get(KEY_REFRESH_TOKEN), expiresAt, values.get(KEY_SCOPE));
+        boolean frontChannelIdTokenExpected = parameters.containsKey(KEY_RESPONSE_TYPE) && parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_ID_TOKEN);
+
+        if (frontChannelIdTokenExpected) {
```
lbalmaceda: always verify the front-channel ID token first, if present. Also, prefer this one over the one received on the back channel (if there is one)

auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
```diff
@@ -0,0 +1,62 @@
...
+
+                @Override
+                public void onFailure(AuthenticationException error) {
+                    callback.onFailure(new TokenValidationException(String.format("Could not find a public key for kid \"%s\"", keyId)));
+                }
+            });
+        } else if ("HS256".equals(algorithmName)) {
+            callback.onSuccess(new SymmetricSignatureVerifier());
+        } else {
+            callback.onFailure(new TokenValidationException(String.format("Signature algorithm of \"%s\" is not supported. Expected either \"RS256\" or \"HS256\".", algorithmName)));
```
lbalmaceda: Message differs from other implementations, but this one is not told by the user what algorithm is expected. It knows which ones it supports, and instantiates a verifier for the one the token was signed with.

auth0/src/main/java/com/auth0/android/provider/SymmetricSignatureVerifier.java
```diff
@@ -0,0 +1,17 @@
+package com.auth0.android.provider;
+
+import android.support.annotation.NonNull;
+
+import com.auth0.android.jwt.JWT;
+
+/**
+ * Token signature verifier for HS256 algorithms.
+ */
+class SymmetricSignatureVerifier extends SignatureVerifier {
```
lbalmaceda: Kept for similarity across implementations. But could remove this class if the main class is declared non-abstract and already implements "doing nothing". Asymmetric could subclass it and implement the new behavior, like today.

auth0/src/test/java/com/auth0/android/provider/OAuthManagerTest.java
```diff
@@ -130,22 +130,4 @@ public void shouldHaveInvalidStateWhenOneIsNull() {
         exception.expect(AuthenticationException.class);
         OAuthManager.assertValidState("0987654321", null);
     }
-
-    @Test
-    public void shouldHaveValidNonce() {
-        OAuthManager.assertValidNonce("1234567890", "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJub25jZSI6IjEyMzQ1Njc4OTAifQ.oUb6xFIEPJQrFbel_Js4SaOwpFfM_kxHxI7xDOHgghk");
```
lbalmaceda: This static method was removed. But the scenario is covered in `WebAuthProviderTest`

auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
```diff
@@ -0,0 +1,149 @@
...
+
+        cal.setTime(token.getExpiresAt());
+        cal.add(Calendar.SECOND, clockSkew);
+        Date expDate = cal.getTime();
+
+        if (now.after(expDate)) {
+            throw new TokenValidationException(String.format("Expiration Time (exp) claim error in the ID token; current time (%d) is after expiration time (%d)", now.getTime() / 1000, expDate.getTime() / 1000));
+        }
+
+        if (token.getIssuedAt() == null) {
```
joshcanhelp: Is this missing the number type check?

auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
joshcanhelp: Not sure exactly what you're asking but seems fine either way.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -128,40 +136,115 @@ boolean resume(AuthorizeResult result) {
...
-            if (!shouldUsePKCE()) {
-                callback.onSuccess(urlCredentials);
-            } else {
-                //Finish Code Exchange
-                pkce.getToken(values.get(KEY_CODE), new AuthCallback() {
-                    @Override
-                    public void onFailure(@NonNull Dialog dialog) {
-                        callback.onFailure(dialog);
+        final Date expiresAt = !values.containsKey(KEY_EXPIRES_IN) ? null : new Date(getCurrentTimeInMillis() + Long.valueOf(values.get(KEY_EXPIRES_IN)) * 1000);
+        final Credentials frontChannelCredentials = new Credentials(values.get(KEY_ID_TOKEN), values.get(KEY_ACCESS_TOKEN), values.get(KEY_TOKEN_TYPE), values.get(KEY_REFRESH_TOKEN), expiresAt, values.get(KEY_SCOPE));
```
joshcanhelp: Why would `frontChannelCredentials` ever have `values.get(KEY_REFRESH_TOKEN)`? Or token type? Or expires? Or scope? Those are not valid `response_type`s.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -128,40 +136,115 @@ boolean resume(AuthorizeResult result) {
...
+        boolean frontChannelIdTokenExpected = parameters.containsKey(KEY_RESPONSE_TYPE) && parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_ID_TOKEN);
+
+        if (frontChannelIdTokenExpected) {
+            //Must be response_type=id_token (or additional values)
+            assertValidIdToken(values.get(KEY_ID_TOKEN), new AuthenticationCallback<Void>() {
```
joshcanhelp: Why aren't you using `frontChannelCredentials` here?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -128,40 +136,115 @@ boolean resume(AuthorizeResult result) {
...
-                callback.onSuccess(urlCredentials);
-            } else {
-                //Finish Code Exchange
-                pkce.getToken(values.get(KEY_CODE), new AuthCallback() {
-                    @Override
-                    public void onFailure(@NonNull Dialog dialog) {
-                        callback.onFailure(dialog);
+        final Date expiresAt = !values.containsKey(KEY_EXPIRES_IN) ? null : new Date(getCurrentTimeInMillis() + Long.valueOf(values.get(KEY_EXPIRES_IN)) * 1000);
+        final Credentials frontChannelCredentials = new Credentials(values.get(KEY_ID_TOKEN), values.get(KEY_ACCESS_TOKEN), values.get(KEY_TOKEN_TYPE), values.get(KEY_REFRESH_TOKEN), expiresAt, values.get(KEY_SCOPE));
+        boolean frontChannelIdTokenExpected = parameters.containsKey(KEY_RESPONSE_TYPE) && parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_ID_TOKEN);
```
joshcanhelp: Can the developer generate a one-off login URL to use that would have a different response type?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -285,10 +388,10 @@ CustomTabsOptions customTabsOptions() {
 
     @VisibleForTesting
     static Credentials mergeCredentials(Credentials urlCredentials, Credentials codeCredentials) {
-        final String idToken = TextUtils.isEmpty(codeCredentials.getIdToken()) ? urlCredentials.getIdToken() : codeCredentials.getIdToken();
+        final String idToken = TextUtils.isEmpty(urlCredentials.getIdToken()) ? codeCredentials.getIdToken() : urlCredentials.getIdToken();
```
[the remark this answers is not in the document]
joshcanhelp: > prefer ID token from the front channel

What does "prefer" here mean? And why do we "prefer" it?

auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
```diff
@@ -0,0 +1,62 @@
...
+     * Validates the algorithm of the given token is supported and creates a new instance of a SignatureVerifier
+     *
+     * @param token     the ID token to create a signature verifier for
+     * @param apiClient api client instance to fetch the JWKS keys, if necessary
+     * @param callback  the callback to receive the result in
+     */
+    static void forToken(@NonNull JWT token, @NonNull AuthenticationAPIClient apiClient, @NonNull final BaseCallback<SignatureVerifier, TokenValidationException> callback) {
+        String algorithmName = token.getHeader().get("alg");
+        if ("RS256".equals(algorithmName)) {
+            final String keyId = token.getHeader().get("kid");
```
joshcanhelp: What if `kid` is empty?

auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
joshcanhelp: > user

Developer?

auth0/src/main/java/com/auth0/android/provider/SymmetricSignatureVerifier.java
joshcanhelp: I would suggest omitting it. Creates a bit of confusion having it (IMHO)

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -204,6 +206,30 @@ public Builder withNonce(@NonNull String nonce) {
...
         }
 
+        /**
+         * Set the max age value for the authentication.
+         *
+         * @param maxAge to use in the requests
+         * @return the current builder instance
+         */
+        public Builder withMaxAge(@NonNull Integer maxAge) {
+            //TODO: This is set by the user. Should that be directly a string?
```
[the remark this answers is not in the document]
joshcanhelp: > user

Developer?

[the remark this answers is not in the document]
joshcanhelp: If you're going to verify an int, I'd use an int.

auth0/src/main/java/com/auth0/android/provider/SymmetricSignatureVerifier.java
lbalmaceda: @damieng what's your take here? I know for .net you've added a `NoSignatureVerifier` that extends/implements `SignatureVerifier`. Should we keep this but rename it as yours, or remove it?

auth0/src/test/resources/rsa_jwks.json
```diff
@@ -0,0 +1,12 @@
+{
+  "keys": [
+    {
+      "kty": "RSA",
```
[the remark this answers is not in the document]
lbalmaceda: not used nor needed to recreate the public key

auth0/src/main/java/com/auth0/android/provider/AsymmetricSignatureVerifier.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param publicKey the public key to use for verification
+     * @throws InvalidKeyException if the public key provided is null or not of type RSA
+     */
+    AsymmetricSignatureVerifier(@Nullable PublicKey publicKey) throws InvalidKeyException {
+        try {
+            publicSignature = Signature.getInstance("SHA256withRSA");
+            publicSignature.initVerify(publicKey);
+        } catch (NoSuchAlgorithmException ignored) {
+            //Safe to ignore: "SHA256withRSA" is available since API 1
```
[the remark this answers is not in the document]
lbalmaceda: If I don't catch it I need to declare it on the method signature, and eventually, someone above (i.e. the caller) would need to catch it. Better to ignore it here and not pop up.

auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
```diff
@@ -0,0 +1,149 @@
...
+     * See https://openid.net/specs/openid-connect-core-1_0-final.html#IDTokenValidation
+     *
+     * @param token         the ID Token to verify.
+     * @param verifyOptions the verification options, like audience, issuer, algorithm.
+     * @throws TokenValidationException If the ID Token is null, its signing algorithm not supported, its signature invalid or one of its claim invalid.
+     */
+    void verify(@NonNull JWT token, @NonNull Options verifyOptions) throws TokenValidationException {
+        verifyOptions.verifier.verifySignature(token);
+
+        if (isEmpty(token.getIssuer())) {
```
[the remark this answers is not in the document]
lbalmaceda: Java is a strongly typed language. The `getIssuer` method will always return a String. The `null` value scenario is captured with the [TextUtils.isEmpty()](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence))

auth0/src/test/java/com/auth0/android/util/MockAuthCallback.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * MockBaseCallback.java
+ *
+ * Copyright (c) 2015 Auth0 (http://auth0.com)
```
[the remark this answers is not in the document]
lbalmaceda: This was added automatically by the IDE, using the project's default LICENSE header template. Don't know if it's needed but every other file has it. The idea is that people copy-pasting our code into their projects also take this with them, just as we should if we do the same with others' code.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
lbalmaceda: A refresh token will never be present on the front channel. I've updated that on the `mergeCredentials` method but forgot this line. Will fix 👍 
The rest of the values you mentioned are always present when an access token is returned.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
lbalmaceda: I guess they can. If they choose `token` that could give them back a pair of credentials with the unverified ID token. I think we should remove the ID token on that scenario. Let me change that.

lbalmaceda: I added a test for when an attacker could put an `id_token` value in the response's fragment. When the response_type was token, it could have reached the dev without been verified.

auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
lbalmaceda: then `keyId` will end up with a `null` value, and when used below there will be no results. Thus, creating an instance of the `AsymmetricSignatureVerifier` will throw. Covered [here](https://github.com/auth0/Auth0.Android/pull/265/files#diff-c422ae1e5c4358e1866a66f51b9440b5R1696)

lbalmaceda: Added an explicit test case for when the claim `keyId` is not set on the header.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
joshcanhelp: I always forget that all those are included in the URL with `response_type=token`

PR #274, opened 2020-01-09
Remove issued_at value check

lbalmaceda:
### Changes

The ID Token `iat` claim value check should be optional.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
+++ b/auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
@@ -68,14 +68,6 @@ class IdTokenVerifier {
             throw new TokenValidationException("Issued At (iat) claim must be a number present in the ID token");
         }
 
-        cal.setTime(token.getIssuedAt());
-        cal.add(Calendar.SECOND, -1 * clockSkew);
-        Date iatDate = cal.getTime();
-
-        if (now.before(iatDate)) {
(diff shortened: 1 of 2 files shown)


auth0/src/main/java/com/auth0/android/provider/IdTokenVerifier.java
```diff
@@ -68,14 +68,6 @@ void verify(@NonNull JWT token, @NonNull IdTokenVerificationOptions verifyOption
             throw new TokenValidationException("Issued At (iat) claim must be a number present in the ID token");
         }
 
-        cal.setTime(token.getIssuedAt());
```
damieng: Having a single mutable calendar object and then setting it to different values betweet tests seems quite an unusual pattern. Is this normal in Java?

lbalmaceda: Think of Calendar as a "dates calculator". It's easier to do _math_ operations on a Calendar than a Date instance. Note though that once the math is done, an instance of Date is obtained from that Calendar for comparison against another one.

damieng: I get why the Calendar object is being used - it's the re-use for multiple calculations that seems odd.

lbalmaceda: I've always seen it being used like this, at least for when the min JDK is 7. `Calendar` is already a heavy object. There's no harm in re-using it and is better than creating multiple instances.

PR #287, opened 2020-02-26
Migrate project to AndroidX

SUPERCILEX:
@lbalmaceda here you go! You'll have to ship JWTDecode.Android before this lib. This PR is currently invalid: it compiles, but a consumer will still get `com.android.support` dependencies transitively from the JWTDecode lib.


auth0/build.gradle
```diff
@@ -68,21 +71,34 @@ android {
...
-    implementation 'com.android.support:appcompat-v7:28.0.0'
-    implementation 'com.android.support:customtabs:28.0.0'
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.browser:browser:1.0.0'
     implementation 'com.squareup.okhttp:okhttp:2.7.5'
     implementation 'com.squareup.okhttp:logging-interceptor:2.7.5'
-    implementation 'com.google.code.gson:gson:2.8.5'
-    implementation 'com.auth0.android:jwtdecode:1.3.0'
+    implementation 'com.google.code.gson:gson:2.8.6'
+    implementation 'com.auth0.android:jwtdecode:1.4.0'
```
SUPERCILEX: This needs to be 2.0.0

gradle.properties
```diff
@@ -1,19 +1,5 @@
...
-# This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
-android.enableUnitTestBinaryResources=true
\ No newline at end of file
+org.gradle.jvmargs=-Xmx2g -Dfile.encoding=UTF-8
+org.gradle.parallel=true
+org.gradle.caching=true
+android.useAndroidX=true
+android.enableJetifier=true
```
SUPERCILEX: This needs to be removed.

lbalmaceda:
@SUPERCILEX thanks for raising the PR. Let me put a hold on this one for a few days until we figure out what is pending and release the JwtDecode one.

SUPERCILEX:
SGTM, let me know when you're ready. 👍

chrisaxiom:
Any app using Unity 2019.3 onward requires this as well, would be great if this could get in as soon as possible 🙏

lbalmaceda:
@chrisaxiom this change is not a blocker for anyone. If your app is still using appcompat/support library dependencies (vs android x dependencies) you can enable jetifier for the android plugin to resolve this for you. See this [doc](https://developer.android.com/jetpack/androidx) for more info.

auth0/build.gradle
lbalmaceda: I just made this release

SUPERCILEX:
@lbalmaceda Whoop whoop! I think this bad boy is ready to go out. :tada:

auth0/src/main/java/com/auth0/android/util/Telemetry.java
```diff
@@ -50,7 +49,7 @@ public Telemetry(String name, String version, String libraryVersion) {
         }
         values.put(ENV_KEY, env);
         String json = new Gson().toJson(values);
-        Charset utf8 = StandardCharsets.UTF_8;
+        Charset utf8 = Charset.forName("UTF-8");
```
lbalmaceda: Any reason not to use `Charset.defaultCharset()`? Which will default to UTF-8 on Android https://developer.android.com/reference/java/nio/charset/Charset

auth0/build.gradle
```diff
@@ -68,21 +71,34 @@ android {
...
-    testImplementation 'org.robolectric:robolectric:4.0.2'
-    testImplementation 'com.android.support.test.espresso:espresso-intents:3.0.2'
-}
\ No newline at end of file
+    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'androidx.test.espresso:espresso-intents:3.2.0'
+}
+
+tasks.withType(Test) {
+    jacoco.includeNoLocationClasses = true
```
lbalmaceda: Maybe this is something to add to all libraries through the plugin instead? https://github.com/auth0/oss-library-gradle-plugin/

build.gradle
```diff
@@ -1,29 +1,28 @@
...
+
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath 'com.android.tools.build:gradle:3.6.1'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
     }
 }
 
+plugins {
+    id 'lifecycle-base'
```
lbalmaceda: Something to add to https://github.com/auth0/oss-library-gradle-plugin/?

auth0/src/main/java/com/auth0/android/util/Telemetry.java
SUPERCILEX: I could, but the goal was to only change imports. This change was a mistake from lint auto-correcting stuff.

auth0/build.gradle
SUPERCILEX: I'd be happy to, do you want me to open a PR on the plugin?

build.gradle
SUPERCILEX: No, since this a root project thing. Unless you want to create a new plugin to configure the root project?

auth0/build.gradle
lbalmaceda: Yes, that will do. Note the plugin has 2 plugins, one for Java and for Android.

build.gradle
lbalmaceda: What do you mean by root project? Maybe I'm missing something, but if I'm correct, you replaced the "clean" task we had with this lifecycle plugin, which seems to remove the build folder for us. You are just extending that to remove the temporary changelog file, which this same oss library plugin we had generates. 
Wouldn't it make sense for this oss library plugin to apply internally the lifecycle-base to both java and android libs, and also extend it to remove the temporary changelog?

SUPERCILEX:
@lbalmaceda Created https://github.com/auth0/oss-library-gradle-plugin/pull/15

lbalmaceda:
@SUPERCILEX thank you, I just released 0.12.0. You can now update this project 👍

lbalmaceda:
@SUPERCILEX your fix is on 0.12.1 now 👍

Issue #293, opened 2020-03-20
SecureCredentialsManager's tokens expiration checks contains bug.


Tans5:
```java


    private void continueGetCredentials(final BaseCallback<Credentials, CredentialsManagerException> callback) {
        ...
        // Here should use expires time in storage(key is KEY_EXPIRES_AT) as judgment.
        if (credentials.getExpiresAt().getTime() > getCurrentTimeInMillis()) {
            callback.onSuccess(credentials);
            decryptCallback = null;
            return;
        }
        ...
    }

```

lbalmaceda:
👋 Thanks for your report.

Actually, before you get to `continueGetCredentials` you must have called `getCredentials`. And the first thing that is being done is checking for valid stored credentials. See https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.java#L192-L195.

This check above considers the right Expires At value, the stored one, and not the one on the credentials. So while the conditional you pointed should use the stored one, there's no bug since the credentials are valid at the time you called this function.

In the worst-case scenario, what could happen is that the manager doesn't give you back the stored Credentials with a good ID Token because it considers the Access Token has already expired, which is really edge case. The only way this could happen is if the expiration occurs in between your call to `getCredentials` and the invocation of the callback you passed with the result/error. Furthermore, when this rare scenario happens, if a Refresh Token is available it will attempt to renew the Credentials with it. 

Let me anyway add a test for the authentication part when the Credentials have expired. And let's change that into the right value as you pointed out for the sake of consistency across the class.

luckpoint:
@lbalmaceda 
Thank you so much for your clarification!
```
public void getCredentials(@NonNull BaseCallback<Credentials, CredentialsManagerException> callback) {
    if (!hasValidCredentials()) {
        callback.onFailure(new CredentialsManagerException("No Credentials were previously set."));
        return;
    }
```
when hasValidCredentials() which checks `expiresAt <= getCurrentTimeInMillis()` returns false, subsequent processing (apiClient.renewAuth) is not called.

There is an inconsistency between validation logics of CredentialsManager#getCredentials and SecureCredentialsManager#getCredentials. Is it ok?

lbalmaceda:
@luckpoint That expires at value comes from the storage. In order to be there, first you'd have to save credentials. 

The save method is the one that contains the logic for checking which time is lower or closer to today, if the id token or the access token expires at. 

[...]

PR #295, opened 2020-03-27
Improve consistency around Expires At in CredentialsManager

lbalmaceda:
### Changes

Add a missing test case. Also, improve consistency on the usage of Expires At for the `CredentialsManager` class.

### References

Will close https://github.com/auth0/Auth0.Android/issues/293

### Testing

### Checklist

### DIFF
--- a/auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.java
+++ b/auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.java
@@ -195,7 +195,6 @@ public class SecureCredentialsManagerTest {
         assertThat(storedCredentials.getExpiresAt(), is(notNullValue()));
         assertThat(storedCredentials.getExpiresAt().getTime(), is(accessTokenExpirationTime));
         assertThat(storedCredentials.getScope(), is("scope"));
-
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.java
```diff
@@ -750,6 +751,36 @@ public void shouldGetCredentialsAfterAuthentication() {
...
+        boolean willRequireAuthentication = manager.requireAuthentication(activity, 123, "theTitle", "theDescription");
+        assertThat(willRequireAuthentication, is(true));
+
+        manager.getCredentials(callback);
+
+        //Should fail because of expired credentials
+        verify(callback).onFailure(exceptionCaptor.capture());
+        CredentialsManagerException exception = exceptionCaptor.getValue();
+        assertThat(exception, is(notNullValue()));
+        assertThat(exception.getMessage(), is("No Credentials were previously set."));
```
Widcket: Shouldn't the message convey the fact that the credentials have expired, instead of saying that there were no credentials?

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.java
```diff
@@ -750,6 +751,36 @@ public void shouldGetCredentialsAfterAuthentication() {
...
+
+        //A second call to checkAuthenticationResult should fail as callback is set to null
+        final boolean retryCheck = manager.checkAuthenticationResult(123, Activity.RESULT_OK);
```
Widcket: Should this be in this test?

lbalmaceda: Yes. One of the assumptions is that this method cannot be called without previously having started the authentication. This checks that any failure scenario, including this particular one, does invalidate any further call until the whole flow it's started again.

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.java
lbalmaceda: No. Having expired credentials is the same as not having them, as the developer won't be able to obtain them. This is the same response given when the user checks for "valid credentials" with the corresponding method.

Issue #309, opened 2020-06-30
NPE in SecureCredentialsManager.getCredentials


iancboswell:
### Description

My app experienced a strange series of crashes last night, all due to a NPE in getCredentials. Here's the stack trace:

```
java.lang.NullPointerException · chain == null
KeyStore.java:1206java.security.KeyStore$PrivateKeyEntry.<init>
KeyStoreSpi.java:374java.security.KeyStoreSpi.engineGetEntry
KeyStore.java:645java.security.KeyStore.getEntry
CryptoUtil.java:151com.auth0.android.authentication.storage.CryptoUtil.getKeyEntryCompat
CryptoUtil.java:88com.auth0.android.authentication.storage.CryptoUtil.getRSAKeyEntry
CryptoUtil.java:200com.auth0.android.authentication.storage.CryptoUtil.RSAEncrypt
CryptoUtil.java:228com.auth0.android.authentication.storage.CryptoUtil.getAESKey
CryptoUtil.java:241com.auth0.android.authentication.storage.CryptoUtil.decrypt
SecureCredentialsManager.java:208com.auth0.android.authentication.storage.SecureCredentialsManager.continueGetCredentials
SecureCredentialsManager.java:175com.auth0.android.authentication.storage.SecureCredentialsManager.getCredentials
```

### Reproduction

Unfortunately I have no idea why this happened or how it could be reproduced. The app code that calls `getCredentials` has been working in production for some time and hasn't been changed recently.

### Environment

- Auth0 library version: 1.15.0
- Android version: 6.0.1
- Language: Java, some Kotlin
- Other libraries: Gson, OKHttp, Google Play

iancboswell:
Actually, a couple of earlier crashes might be related. These happened a couple days before the one above.

```
 java.lang.IllegalArgumentException · key.length == 0
SecretKeySpec.java:62javax.crypto.spec.SecretKeySpec.<init>
CryptoUtil.java:241com.auth0.android.authentication.storage.CryptoUtil.decrypt
SecureCredentialsManager.java:208com.auth0.android.authentication.storage.SecureCredentialsManager.continueGetCredentials
SecureCredentialsManager.java:175com.auth0.android.authentication.storage.SecureCredentialsManager.getCredentials
```

```
 com.auth0.android.authentication.storage.CredentialsManagerException · An error occurred while encrypting the credentials.
[...]

PR #315, opened 2020-07-07
Use of a unique keystore key alias across apps

lbalmaceda:
### Changes

The intention is to use the App's Package Name from the `Context` vs the library's Application ID from the `BuildConfig`, to generate a key alias that is **unique across app** installs. This should prevent errors where apps using this very same library (or sharing the key alias) would have their keys collide.

#### Note
Since the key alias is changing with this PR, when you update to this last version of the SDK, any existing key will not be found and it will act as if the user was logged out. Double-check that you are not ignoring the exceptions thrown when the credential manager methods are being used. See our [docs](https://github.com/auth0/Auth0.Android#credentials-manager) for more information.

### References
Closes https://github.com/auth0/Auth0.Android/issues/309

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
@@ -74,8 +74,8 @@ class CryptoUtil {
         if (TextUtils.isEmpty(keyAlias)) {
             throw new IllegalArgumentException("RSA and AES Key alias must be valid.");
         }
-        this.KEY_ALIAS = keyAlias;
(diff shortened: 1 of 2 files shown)


FilipKastrupGP:
How would I migrate from 1.23.0 to 1.24.0 without logging the user out?

lbalmaceda:
@FilipKastrupGP When you migrate, logged in users will be logged out. You should be first checking if credentials are available calling the `hasCredentials` method. Depending on its result, you'd be prompting the user to log in and saving the new credentials, or getting the ones stored from the manager. Always [checking for any exception](https://github.com/auth0/Auth0.Android#handling-exceptions) that might be raised in the process.

PR #323, opened 2020-07-30
Use deprecated key alias if already present.

morganlutz:
This change affects pre-existing data. New logins should encrypt/decrypt data with the new key alias


morganlutz:
Opened against the wrong master instead of my fork!

Issue #349, opened 2020-09-21
Preemptive Credentials refresh [SDK-1819]


skjolber:
### Describe the problem you'd like to have solved
We're building a ticket inspection app. We want to avoid interaction with the authorization server as much as possible during the inspection session, which can last from a few minutes to a few hours (all limited by the authorization server settings). 

### Describe the ideal solution

We would like preemptively refresh the access-token, and/or trigger a new (manual) login if the refresh-token is about to expire. Ideally some kind of `forceRefresh` parameter could be added and we would get the relevant timestamps from `Credentials`.

## Alternatives and current work-arounds

Get the credentials, look at the id-token expire time, clear them if not acceptable.

### Additional context

[SecureCredentialsManager](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.java)

lbalmaceda:
Hey @skjolber ! 👋  If I understand correctly what you're asking, you need the "Credentials Managers" to expose some way to manually refresh the tokens, even when they have not expired yet and shouldn't be refreshed. Is that correct?

If that's the case, one of the solutions we implemented in iOS and want to bring to Android as well is passing some TTL value on the `getCredentials` call, so that you can specify that you require an access token that will last for at least X (TTL) seconds. 

Would a solution like that work for you? If so, I can link this issue to that card.

skjolber:
That is right. Yes, I think the TTL approach you're describing will work for us :+1:

lbalmaceda:
Excellent. Tracking under `SDK-1819`.

lbalmaceda:
@skjolber the 2 PRs above add support for this in both credential managers. I'll have it released in the next minor.

lbalmaceda:
Feature added in PRs 363 and 369

PR #363, opened 2020-10-21
CredentialsManager: Allow to pass scope and minTTL

lbalmaceda:
### Changes

This PR improves the feature parity with iOS' CredentialManager class. Now developers can specify a reduced scope for when the tokens expire and need to be refreshed. And also a minimum time to live that the received tokens must have. 

In order to add support for this, a new method was introduced. The old functionality remains the same.

A following PR will add the same support for the `SecureCredentialsManager`.

### References

Relates to https://github.com/auth0/Auth0.Android/issues/349

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -75,7 +75,7 @@ public class CredentialsManager {
             throw new CredentialsManagerException("Credentials must have a valid date of expiration and a valid access_token or id_token value.");
         }
 
-        long expiresAt = calculateExpiresAt(credentials);
+        long cacheExpiresAt = calculateCacheExpiresAt(credentials);
 
         storage.store(KEY_ACCESS_TOKEN, credentials.getAccessToken());
         storage.store(KEY_REFRESH_TOKEN, credentials.getRefreshToken());
@@ -83,7 +83,7 @@ public class CredentialsManager {
         storage.store(KEY_TOKEN_TYPE, credentials.getType());
         storage.store(KEY_EXPIRES_AT, credentials.getExpiresAt().getTime());
         storage.store(KEY_SCOPE, credentials.getScope());
-        storage.store(KEY_CACHE_EXPIRES_AT, expiresAt);
+        storage.store(KEY_CACHE_EXPIRES_AT, cacheExpiresAt);
     }
 
     /**
@@ -103,7 +103,7 @@ public class CredentialsManager {
      * or if the tokens have already expired and the refresh_token is null.
      *
      * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
-     * @param minTtl   the minimum time in seconds that both the access token and id token should last before expiration.
+     * @param minTtl   the minimum time in seconds that the access token should last before expiration.
      * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
      */
     public void getCredentials(@Nullable String scope, final int minTtl, @NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
@@ -124,10 +124,11 @@ public class CredentialsManager {
             return;
         }
 
-        boolean willExpire = willExpire(cacheExpiresAt, minTtl);
+        boolean hasEitherExpired = hasExpired(cacheExpiresAt);
+        boolean willAccessTokenExpire = willExpire(expiresAt, minTtl);
         boolean scopeChanged = hasScopeChanged(storedScope, scope);
 
-        if (!willExpire && !scopeChanged) {
+        if (!hasEitherExpired && !willAccessTokenExpire && !scopeChanged) {
             callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), storedScope));
             return;
         }
@@ -143,10 +144,11 @@ public class CredentialsManager {
         request.start(new AuthenticationCallback<Credentials>() {
             @Override
             public void onSuccess(@Nullable Credentials fresh) {
-                long nextCacheExpiresAt = calculateExpiresAt(fresh);
-                boolean willExpire = willExpire(nextCacheExpiresAt, minTtl);
-                if (willExpire) {
-                    long tokenLifetime = (nextCacheExpiresAt - getCurrentTimeInMillis() - minTtl * 1000) / -1000;
+                //noinspection ConstantConditions
+                long expiresAt = fresh.getExpiresAt().getTime();
+                boolean willAccessTokenExpire = willExpire(expiresAt, minTtl);
+                if (willAccessTokenExpire) {
+                    long tokenLifetime = (expiresAt - getCurrentTimeInMillis() - minTtl * 1000) / -1000;
                     CredentialsManagerException wrongTtlException = new CredentialsManagerException(String.format("The lifetime of the renewed Access Token or Id Token (%d) is less than the minTTL requested (%d). Increase the 'Token Expiration' setting of your Auth0 API or the 'ID Token Expiration' of your Auth0 Application in the dashboard, or request a lower minTTL.", tokenLifetime, minTtl));
                     callback.onFailure(wrongTtlException);
                     return;
@@ -183,7 +185,11 @@ public class CredentialsManager {
         return expiresAt <= nextClock;
     }
 
-    private long calculateExpiresAt(@NonNull Credentials credentials) {
+    private boolean hasExpired(long expiresAt) {
+        return expiresAt <= getCurrentTimeInMillis();
+    }
+
+    private long calculateCacheExpiresAt(@NonNull Credentials credentials) {
         long expiresAt = credentials.getExpiresAt().getTime();
 
         if (credentials.getIdToken() != null) {
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -144,12 +144,11 @@ public class CredentialsManager {
         request.start(new AuthenticationCallback<Credentials>() {
             @Override
             public void onSuccess(@Nullable Credentials fresh) {
-                //noinspection ConstantConditions
                 long expiresAt = fresh.getExpiresAt().getTime();
                 boolean willAccessTokenExpire = willExpire(expiresAt, minTtl);
                 if (willAccessTokenExpire) {
                     long tokenLifetime = (expiresAt - getCurrentTimeInMillis() - minTtl * 1000) / -1000;
-                    CredentialsManagerException wrongTtlException = new CredentialsManagerException(String.format("The lifetime of the renewed Access Token or Id Token (%d) is less than the minTTL requested (%d). Increase the 'Token Expiration' setting of your Auth0 API or the 'ID Token Expiration' of your Auth0 Application in the dashboard, or request a lower minTTL.", tokenLifetime, minTtl));
+                    CredentialsManagerException wrongTtlException = new CredentialsManagerException(String.format("The lifetime of the renewed Access Token (%d) is less than the minTTL requested (%d). Increase the 'Token Expiration' setting of your Auth0 API in the dashboard, or request a lower minTTL.", tokenLifetime, minTtl));
                     callback.onFailure(wrongTtlException);
                     return;
                 }
@@ -209,6 +208,16 @@ public class CredentialsManager {
      * @return whether there are valid credentials stored on this manager.
      */
     public boolean hasValidCredentials() {
+        return hasValidCredentials(0);
+    }
+
+    /**
+     * Checks if a non-expired pair of credentials can be obtained from this manager.
+     *
+     * @param minTtl the minimum time in seconds that the access token should last before expiration.
+     * @return whether there are valid credentials stored on this manager.
+     */
+    public boolean hasValidCredentials(long minTtl) {
         String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
         String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
         String idToken = storage.retrieveString(KEY_ID_TOKEN);
@@ -218,9 +227,8 @@ public class CredentialsManager {
             cacheExpiresAt = expiresAt;
         }
 
-        return !(isEmpty(accessToken) && isEmpty(idToken) ||
-                cacheExpiresAt == null ||
-                hasExpired(cacheExpiresAt) && refreshToken == null);
+        boolean emptyCredentials = isEmpty(accessToken) && isEmpty(idToken) || cacheExpiresAt == null || expiresAt == null;
+        return !(emptyCredentials || (hasExpired(cacheExpiresAt) || willExpire(expiresAt, minTtl)) && refreshToken == null);
     }
 
     /**
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -132,7 +132,7 @@ public class CredentialsManager {
             return;
         }
         if (refreshToken == null) {
-            callback.onFailure(new CredentialsManagerException("Credentials have expired and no Refresh Token was available to renew them."));
+            callback.onFailure(new CredentialsManagerException("Credentials need to be renewed but no Refresh Token is available to renew them."));
             return;
         }
 
@@ -175,7 +175,7 @@ public class CredentialsManager {
         Arrays.sort(stored);
         String[] required = requiredScope.split(" ");
         Arrays.sort(required);
-        return stored != required;
+        return !Arrays.equals(stored, required);
     }
 
     private boolean willExpire(long expiresAt, long minTtl) {
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
@@ -9,9 +9,11 @@ import com.auth0.android.authentication.AuthenticationException;
 import com.auth0.android.callback.AuthenticationCallback;
 import com.auth0.android.callback.BaseCallback;
 import com.auth0.android.jwt.JWT;
+import com.auth0.android.request.ParameterizableRequest;
 import com.auth0.android.result.Credentials;
 import com.auth0.android.util.Clock;
 
+import java.util.Arrays;
 import java.util.Date;
 
 import static android.text.TextUtils.isEmpty;
@@ -73,16 +75,7 @@ public class CredentialsManager {
             throw new CredentialsManagerException("Credentials must have a valid date of expiration and a valid access_token or id_token value.");
         }
 
-        long expiresAt = credentials.getExpiresAt().getTime();
-
-        if (credentials.getIdToken() != null) {
-            JWT idToken = jwtDecoder.decode(credentials.getIdToken());
-            Date idTokenExpiresAtDate = idToken.getExpiresAt();
-
-            if (idTokenExpiresAtDate != null) {
-                expiresAt = Math.min(idTokenExpiresAtDate.getTime(), expiresAt);
-            }
-        }
+        long expiresAt = calculateExpiresAt(credentials);
 
         storage.store(KEY_ACCESS_TOKEN, credentials.getAccessToken());
         storage.store(KEY_REFRESH_TOKEN, credentials.getRefreshToken());
@@ -100,24 +93,42 @@ public class CredentialsManager {
      *
      * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
      */
-    public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
+    public void getCredentials(@NonNull BaseCallback<Credentials, CredentialsManagerException> callback) {
+        getCredentials(null, 0, callback);
+    }
+
+    /**
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with {@link CredentialsManagerException} if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
+     * @param minTtl   the minimum time in seconds that both the access token and id token should last before expiration.
+     * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
+     */
+    public void getCredentials(@Nullable String scope, final int minTtl, @NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
         String accessToken = storage.retrieveString(KEY_ACCESS_TOKEN);
         final String refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN);
         String idToken = storage.retrieveString(KEY_ID_TOKEN);
         String tokenType = storage.retrieveString(KEY_TOKEN_TYPE);
         Long expiresAt = storage.retrieveLong(KEY_EXPIRES_AT);
-        String scope = storage.retrieveString(KEY_SCOPE);
+        String storedScope = storage.retrieveString(KEY_SCOPE);
         Long cacheExpiresAt = storage.retrieveLong(KEY_CACHE_EXPIRES_AT);
         if (cacheExpiresAt == null) {
             cacheExpiresAt = expiresAt;
         }
 
-        if (isEmpty(accessToken) && isEmpty(idToken) || expiresAt == null) {
+        boolean hasEmptyCredentials = isEmpty(accessToken) && isEmpty(idToken) || expiresAt == null;
+        if (hasEmptyCredentials) {
             callback.onFailure(new CredentialsManagerException("No Credentials were previously set."));
             return;
         }
-        if (cacheExpiresAt > getCurrentTimeInMillis()) {
-            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), scope));
+
+        boolean willExpire = willExpire(cacheExpiresAt, minTtl);
+        boolean scopeChanged = hasScopeChanged(storedScope, scope);
+
+        if (!willExpire && !scopeChanged) {
+            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), storedScope));
             return;
         }
         if (refreshToken == null) {
@@ -125,9 +136,22 @@ public class CredentialsManager {
             return;
         }
 
-        authClient.renewAuth(refreshToken).start(new AuthenticationCallback<Credentials>() {
+        final ParameterizableRequest<Credentials, AuthenticationException> request = authClient.renewAuth(refreshToken);
+        if (scope != null) {
+            request.addParameter("scope", scope);
+        }
+        request.start(new AuthenticationCallback<Credentials>() {
             @Override
             public void onSuccess(@Nullable Credentials fresh) {
+                long nextCacheExpiresAt = calculateExpiresAt(fresh);
+                boolean willExpire = willExpire(nextCacheExpiresAt, minTtl);
+                if (willExpire) {
+                    long tokenLifetime = (nextCacheExpiresAt - getCurrentTimeInMillis() - minTtl * 1000) / -1000;
+                    CredentialsManagerException wrongTtlException = new CredentialsManagerException(String.format("The lifetime of the renewed Access Token or Id Token (%d) is less than the minTTL requested (%d). Increase the 'Token Expiration' setting of your Auth0 API or the 'ID Token Expiration' of your Auth0 Application in the dashboard, or request a lower minTTL.", tokenLifetime, minTtl));
+                    callback.onFailure(wrongTtlException);
+                    return;
+                }
+
                 //non-empty refresh token for refresh token rotation scenarios
                 String updatedRefreshToken = isEmpty(fresh.getRefreshToken()) ? refreshToken : fresh.getRefreshToken();
                 Credentials credentials = new Credentials(fresh.getIdToken(), fresh.getAccessToken(), fresh.getType(), updatedRefreshToken, fresh.getExpiresAt(), fresh.getScope());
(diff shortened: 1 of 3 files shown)


auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -140,6 +164,37 @@ public void onFailure(@NonNull AuthenticationException error) {
...
+            Date idTokenExpiresAtDate = idToken.getExpiresAt();
+
+            if (idTokenExpiresAtDate != null) {
+                expiresAt = Math.min(idTokenExpiresAtDate.getTime(), expiresAt);
+            }
+        }
+        return expiresAt;
     }
 
     /**
```
Widcket, line 200: `hasValidCredentials()` should have an overload that takes `minTtl`.

Widcket: See https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L124

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -140,6 +164,37 @@ public void onFailure(@NonNull AuthenticationException error) {
...
+
+    private boolean willExpire(long expiresAt, long minTtl) {
+        long nextClock = getCurrentTimeInMillis() + minTtl * 1000;
+        return expiresAt <= nextClock;
+    }
+
+    private long calculateExpiresAt(@NonNull Credentials credentials) {
+        long expiresAt = credentials.getExpiresAt().getTime();
+
+        if (credentials.getIdToken() != null) {
```
Widcket, line 189: The original use case that motivated this was about sending an Access Token (with enough lifetime left) to the backend. Shouldn't just the AT expire time be used here, instead of involving the ID Token as well?  See https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L205.

`getCredentials()` uses this method to calculate the value that is then passed to `willExpire()`.

lbalmaceda: I based my solution on your implementation but I'm keeping compatibility with the feature requested a while ago [here](https://github.com/auth0/Auth0.Android/issues/253). The TTL will consider both expirations and pick the most short one when comparing.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: @Widcket Thanks for pointing this! I've thought about this as well and got stuck thinking about the use case. When access/ID tokens are stored and have expired, the method would only return `true` if a refresh token is also stored, meaning that it can refresh it for new non-expired access/ID tokens. How will knowing the minTTL value affect the returned value if the refresh token is present to refresh it anyway?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -100,34 +93,65 @@ public void saveCredentials(@NonNull Credentials credentials) {
      *
      * @param callback the callback that will receive a valid {@link Credentials} or the {@link CredentialsManagerException}.
      */
-    public void getCredentials(@NonNull final BaseCallback<Credentials, CredentialsManagerException> callback) {
+    public void getCredentials(@NonNull BaseCallback<Credentials, CredentialsManagerException> callback) {
```
Widcket, line 96: Mark the `callback` parameter as `final`, like it is in the overload.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
         assertThat(retrievedCredentials.getScope(), is("scope"));
     }
 
+    @Test
+    public void shouldRenewCredentialsWhenScopeHasChanged() {
+        when(storage.retrieveString("com.auth0.id_token")).thenReturn("idToken");
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn("accessToken");
+        when(storage.retrieveString("com.auth0.refresh_token")).thenReturn("refreshToken");
+        when(storage.retrieveString("com.auth0.token_type")).thenReturn("type");
+        long expirationTime = CredentialsMock.CURRENT_TIME_MS + 123456L * 1000; //non expired credentials
```
Widcket, line 354:
```suggestion
        long expirationTime = CredentialsMock.CURRENT_TIME_MS + 123456L * 1000; // non expired credentials
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 1234 * 1000);
+        JWT jwtMock = mock(JWT.class);
+        when(jwtMock.getExpiresAt()).thenReturn(newDate);
+        when(jwtDecoder.decode("newId")).thenReturn(jwtMock);
+
+        manager.getCredentials("some scope", 0, callback);
+        verify(request).start(requestCallbackCaptor.capture());
+        verify(request).addParameter(eq("scope"), eq("some scope"));
+
+        //Trigger success
```
Widcket, line 369:
```suggestion
        // Trigger success
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        String newRefresh = null;
+        Credentials renewedCredentials = new Credentials("newId", "newAccess", "newType", newRefresh, newDate, "newScope");
+        requestCallbackCaptor.getValue().onSuccess(renewedCredentials);
+        verify(callback).onSuccess(credentialsCaptor.capture());
+
+        // Verify the credentials are property stored
+        verify(storage).store("com.auth0.id_token", renewedCredentials.getIdToken());
+        verify(storage).store("com.auth0.access_token", renewedCredentials.getAccessToken());
+        //RefreshToken should not be replaced
```
Widcket, line 378:
```suggestion
        // RefreshToken should not be replaced
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        verify(storage, never()).store("com.auth0.refresh_token", newRefresh);
+        verify(storage).store("com.auth0.refresh_token", "refreshToken");
+        verify(storage).store("com.auth0.token_type", renewedCredentials.getType());
+        verify(storage).store("com.auth0.expires_at", renewedCredentials.getExpiresAt().getTime());
+        verify(storage).store("com.auth0.scope", renewedCredentials.getScope());
+        verify(storage).store("com.auth0.cache_expires_at", renewedCredentials.getExpiresAt().getTime());
+        verify(storage, never()).remove(anyString());
+
+        //// Verify the returned credentials are the latest
```
Widcket, line 387:
```suggestion
        // Verify the returned credentials are the latest
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        assertThat(retrievedCredentials.getScope(), is("newScope"));
+    }
+
+    @Test
+    public void shouldRenewCredentialsWithMinTtl() {
+        when(storage.retrieveString("com.auth0.id_token")).thenReturn("idToken");
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn("accessToken");
+        when(storage.retrieveString("com.auth0.refresh_token")).thenReturn("refreshToken");
+        when(storage.retrieveString("com.auth0.token_type")).thenReturn("type");
+        long expirationTime = CredentialsMock.CURRENT_TIME_MS; //Same as current time --> expired
```
Widcket, line 404:
```suggestion
        long expirationTime = CredentialsMock.CURRENT_TIME_MS; // Same as current time --> expired
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        when(storage.retrieveLong("com.auth0.expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveLong("com.auth0.cache_expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveString("com.auth0.scope")).thenReturn("scope");
+        when(client.renewAuth("refreshToken")).thenReturn(request);
+
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 61 * 1000); //new token expires in minTTL + 1 seconds
```
Widcket, line 410:
```suggestion
        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 61 * 1000); // New token expires in minTTL + 1 seconds
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        JWT jwtMock = mock(JWT.class);
+        when(jwtMock.getExpiresAt()).thenReturn(newDate);
+        when(jwtDecoder.decode("newId")).thenReturn(jwtMock);
+
+        manager.getCredentials(null, 60, callback); //60 seconds of minTTL
+        verify(request, never()).addParameter(eq("scope"), anyString());
+        verify(request).start(requestCallbackCaptor.capture());
+
+        //Trigger success
```
Widcket, line 419:
```suggestion
        // Trigger success
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        String newRefresh = null;
+        Credentials renewedCredentials = new Credentials("newId", "newAccess", "newType", newRefresh, newDate, "newScope");
+        requestCallbackCaptor.getValue().onSuccess(renewedCredentials);
+        verify(callback).onSuccess(credentialsCaptor.capture());
+
+        // Verify the credentials are property stored
+        verify(storage).store("com.auth0.id_token", renewedCredentials.getIdToken());
+        verify(storage).store("com.auth0.access_token", renewedCredentials.getAccessToken());
+        //RefreshToken should not be replaced
```
Widcket, line 428:
```suggestion
        // RefreshToken should not be replaced
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        verify(storage, never()).store("com.auth0.refresh_token", newRefresh);
+        verify(storage).store("com.auth0.refresh_token", "refreshToken");
+        verify(storage).store("com.auth0.token_type", renewedCredentials.getType());
+        verify(storage).store("com.auth0.expires_at", renewedCredentials.getExpiresAt().getTime());
+        verify(storage).store("com.auth0.scope", renewedCredentials.getScope());
+        verify(storage).store("com.auth0.cache_expires_at", renewedCredentials.getExpiresAt().getTime());
+        verify(storage, never()).remove(anyString());
+
+        //// Verify the returned credentials are the latest
```
Widcket, line 437:
```suggestion
        // Verify the returned credentials are the latest
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -393,6 +495,46 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
         assertThat(retrievedCredentials.getScope(), is("newScope"));
     }
 
+    @Test
+    public void shouldGetAndFailToRenewExpiredCredentialsWhenReceivedTokenHasLowerTtl() {
+        when(storage.retrieveString("com.auth0.id_token")).thenReturn("idToken");
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn("accessToken");
+        when(storage.retrieveString("com.auth0.refresh_token")).thenReturn("refreshToken");
+        when(storage.retrieveString("com.auth0.token_type")).thenReturn("type");
+        long expirationTime = CredentialsMock.CURRENT_TIME_MS; //Same as current time --> expired
```
Widcket, line 504:
```suggestion
        long expirationTime = CredentialsMock.CURRENT_TIME_MS; // Same as current time --> expired
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -393,6 +495,46 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
...
+        when(storage.retrieveLong("com.auth0.expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveLong("com.auth0.cache_expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveString("com.auth0.scope")).thenReturn("scope");
+        when(client.renewAuth("refreshToken")).thenReturn(request);
+
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 59 * 1000); //new token expires in minTTL - 1 seconds
```
Widcket, line 510:
```suggestion
        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 59 * 1000); // New token expires in minTTL - 1 seconds
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -393,6 +495,46 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
...
+        JWT jwtMock = mock(JWT.class);
+        when(jwtMock.getExpiresAt()).thenReturn(newDate);
+        when(jwtDecoder.decode("newId")).thenReturn(jwtMock);
+
+        manager.getCredentials(null, 60, callback); //60 seconds of minTTL
```
Widcket, line 515:
```suggestion
        manager.getCredentials(null, 60, callback); // 60 seconds of minTTL
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -393,6 +495,46 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
...
+        verify(request, never()).addParameter(eq("scope"), anyString());
+        verify(request).start(requestCallbackCaptor.capture());
+
+        //Trigger success
```
Widcket, line 519:
```suggestion
        // Trigger success
```

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -100,34 +93,65 @@ public void saveCredentials(@NonNull Credentials credentials) {
...
+
+        boolean willExpire = willExpire(cacheExpiresAt, minTtl);
+        boolean scopeChanged = hasScopeChanged(storedScope, scope);
+
+        if (!willExpire && !scopeChanged) {
+            callback.onSuccess(recreateCredentials(idToken, accessToken, tokenType, refreshToken, new Date(expiresAt), storedScope));
             return;
         }
         if (refreshToken == null) {
             callback.onFailure(new CredentialsManagerException("Credentials have expired and no Refresh Token was available to renew them."));
```
Widcket, line 135: Why not update the error message to denote that the credentials "need to be refreshed" instead of saying that they expired? That error will also trigger if the credentials have not expired, but the scope changed

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        when(storage.retrieveString("com.auth0.access_token")).thenReturn("accessToken");
+        when(storage.retrieveString("com.auth0.refresh_token")).thenReturn("refreshToken");
+        when(storage.retrieveString("com.auth0.token_type")).thenReturn("type");
+        long expirationTime = CredentialsMock.CURRENT_TIME_MS + 123456L * 1000; //non expired credentials
+        when(storage.retrieveLong("com.auth0.expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveLong("com.auth0.cache_expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveString("com.auth0.scope")).thenReturn("some new scope");
+        when(client.renewAuth("refreshToken")).thenReturn(request);
+
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 1234 * 1000);
```
Widcket, line 360: Why not use the same `expirationTime`?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -100,34 +93,65 @@ public void saveCredentials(@NonNull Credentials credentials) {
...
+            request.addParameter("scope", scope);
+        }
+        request.start(new AuthenticationCallback<Credentials>() {
             @Override
             public void onSuccess(@Nullable Credentials fresh) {
+                long nextCacheExpiresAt = calculateExpiresAt(fresh);
+                boolean willExpire = willExpire(nextCacheExpiresAt, minTtl);
+                if (willExpire) {
+                    long tokenLifetime = (nextCacheExpiresAt - getCurrentTimeInMillis() - minTtl * 1000) / -1000;
+                    CredentialsManagerException wrongTtlException = new CredentialsManagerException(String.format("The lifetime of the renewed Access Token or Id Token (%d) is less than the minTTL requested (%d). Increase the 'Token Expiration' setting of your Auth0 API or the 'ID Token Expiration' of your Auth0 Application in the dashboard, or request a lower minTTL.", tokenLifetime, minTtl));
```
Widcket, line 150: Idem, the use case that brought the TTL renewal feature to life was about making sure the AT would have enough lifetime before sending it to a backend server. Shouldn't only the AT be considered here? Specially considering the developer is not expected to send the ID Token anywhere.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
Widcket, line 410:
```suggestion
        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 61 * 1000); // New token expires in minTTL + 1 second
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -344,6 +345,106 @@ public void shouldGetNonExpiredCredentialsFromStorageWhenOnlyAccessTokenIsAvaila
...
+        when(storage.retrieveLong("com.auth0.cache_expires_at")).thenReturn(expirationTime);
+        when(storage.retrieveString("com.auth0.scope")).thenReturn("scope");
+        when(client.renewAuth("refreshToken")).thenReturn(request);
+
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 61 * 1000); //new token expires in minTTL + 1 seconds
+        JWT jwtMock = mock(JWT.class);
+        when(jwtMock.getExpiresAt()).thenReturn(newDate);
+        when(jwtDecoder.decode("newId")).thenReturn(jwtMock);
+
+        manager.getCredentials(null, 60, callback); //60 seconds of minTTL
```
Widcket, line 415:
```suggestion
        manager.getCredentials(null, 60, callback); // 60 seconds of minTTL
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -356,12 +457,13 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
         when(storage.retrieveString("com.auth0.scope")).thenReturn("scope");
         when(client.renewAuth("refreshToken")).thenReturn(request);
 
-        Date newDate = new Date(123412341234L);
+        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 1234 * 1000);
```
Widcket, line 460: The numeric value could be extracted as a `validExpirationTime` constant, as it's used in a few places.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
Widcket, line 510:
```suggestion
        Date newDate = new Date(CredentialsMock.CURRENT_TIME_MS + 59 * 1000); // New token expires in minTTL - 1 second
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
Widcket, line 519:
```suggestion
        // Trigger failure
```

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -405,7 +547,7 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentialsWithRefreshTokenRotat
...
```
Widcket, line 550: Idem as above, the numerical value could be extracted as a constant.

Widcket (changes requested):
The logic differs from the Auth0.swift implementation.

Widcket:
Also I'd add another test case where the credentials have expired and the scope is different, to see that the refreshed credentials have the new scope.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: > How will knowing the minTTL value affect the returned value if the refresh token is present to refresh it anyway?

If there's not enough TTL left, then it should return `false` to inform that it does not satisfy the developer's TTL requirements.

>  the method would only return true if a refresh token is also stored

Why would a method that is meant to check the validity of the stored credentials concern itself with the presence/absence of the refresh token?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: In Auth0.swift there's 2 methods: `hasExpired` (that considers the ID Token) and `willExpire` (that doesn't). This way, if the ID Token has expired it will still renew the credentials.
See https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L175

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
```diff
@@ -140,6 +164,37 @@ public void onFailure(@NonNull AuthenticationException error) {
...
+
+    private boolean hasScopeChanged(@NonNull String storedScope, @Nullable String requiredScope) {
+        if (requiredScope == null) {
+            return false;
+        }
+        String[] stored = storedScope.split(" ");
+        Arrays.sort(stored);
+        String[] required = requiredScope.split(" ");
+        Arrays.sort(required);
+        return stored != required;
```
Widcket, line 178: Shouldn't this be `!Arrays.equals(stored, required)`? With `!=` the references will be compared, not the contents, AFAIK.

Widcket: In Auth0.swift these methods are tested individually: https://github.com/auth0/Auth0.swift/blob/master/Auth0Tests/CredentialsManagerSpec.swift#L300, https://github.com/auth0/Auth0.swift/blob/master/Auth0Tests/CredentialsManagerSpec.swift#L324

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: Because the refresh token presence implies that the expired or not long-lived enough (minTTL) token can be renewed for a fresh pair. The end goal of the "hasCredentials" method is not to check if the _currently stored_ tokens are valid but if valid tokens would be obtainable when "retrieveCredentials" is called. The same is done in iOS [here](https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L128).

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: There's no need to have two methods. If `willExpire` is passed 0 is the same as calculating a `hasExpired`, so both conditions can be checked with a single method.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: the `final` modifier is only required when passed to a nested body / lambda. In this case, that's no longer required.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: I'll use "Credentials need to be renewed but no Refresh Token is available to renew them."

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: No, that would break compatibility with the feature requested in the linked PR (see the other comment).

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: Good catch. Yes, I should be using that util method.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
lbalmaceda: I could but, in theory, shouldn't the new tokens have a newer date? I'll refactor these into constants and use something like NOW and TOMORROW

lbalmaceda: I've ended up using CURRENT_TIME_MS and added one new that adds one hour on top of that ONE_HOUR_AHEAD_MS. Looks cleaner now.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: The refresh token presence means the credentials can be renewed, but not that they're currently valid. The `hasValid(minTTL:)` method doc comment states `Checks if a non-expired set of credentials are stored`, so the goal of that method is to check if the stored credentials are still valid or not indeed. Since `hasValid(minTTL:)` is a public method that can be used on its own, it should not do other things like check the refresh token presence/absence. Such a method should:
- Return `true` if the AT and ID Token have not expired, and if the AT will not expire within the TTL.
- Return `false` otherwise.

Widcket: If valid tokens are not obtainable because there is no refresh token, then the method that refreshes the credentials should fail, not `hasValid(minTTL:)`.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: The point to have two methods is to check ID Token in one and not in the other.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: That would be the outcome of applying the Principle of Least Astonishment, I think.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: It's possible to check the TTL of just the AT without breaking compatibility, if you also check that the ID Token has not expired yet. So: 
- Check that the AT and the ID Token have not expired yet
- Check that the AT will not expire within the TTL
That will keep the logic consistent with Auth0.swift and will not break compatibility with that feature request.

Widcket: After all, the existing behavior (that needs to be preserved) is about checking if the credentials already expired, while the new behavior is about checking if they will expire. So the new, additive behavior can be limited to the AT, while keeping the existing behavior unchanged.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: We agreed to not change this method now. We can discuss for the next version.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: Fixed. Do you mind taking a look, please?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
lbalmaceda: Actually, let's talk about this. Maybe now that we've split the methods it makes sense to add.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: Please also update the error message, it still references the ID Token.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.java
Widcket: Yes, it makes sense. Otherwise in one method (`getCredentials`) it's taking minTTL into account and in another (`hasValidCredentials`) it isn't. A customer that is using minTTL with `getCredentials` would reasonably expect `hasValidCredentials` to work the same way.

Widcket: Please add it.

Widcket (changes requested):
Looking good, just missing the `minTtl`  overload of `hasValidCredentials`.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.java
```diff
@@ -393,6 +546,46 @@ public void shouldGetAndSuccessfullyRenewExpiredCredentials() {
...
+        verify(storage, never()).store(anyString(), anyInt());
+        verify(storage, never()).store(anyString(), anyLong());
+        verify(storage, never()).store(anyString(), anyString());
+        verify(storage, never()).store(anyString(), anyBoolean());
+        verify(storage, never()).remove(anyString());
+
+        CredentialsManagerException exception = exceptionCaptor.getValue();
+        assertThat(exception, is(notNullValue()));
+        assertThat(exception.getCause(), is(nullValue()));
+        assertThat(exception.getMessage(), is("The lifetime of the renewed Access Token or Id Token (1) is less than the minTTL requested (60). Increase the 'Token Expiration' setting of your Auth0 API or the 'ID Token Expiration' of your Auth0 Application in the dashboard, or request a lower minTTL."));
```
Widcket, line 586: Update the error message here as well.

lbalmaceda:
@Widcket ready! 🤞

Widcket (dismissed):

Widcket (approved):

PR #371, opened 2020-11-04
Handle API response for mobile OTP code incorrect.

nicbell:
### Changes

When logging in with a mobile phone and SMS code: 

```kotlin
AuthenticationAPIClient(account)
            .loginWithPhoneNumber(mobileNumber, code, smsConnection)
            .setScope(scope)
            .setAudience(audience)
            .start
```

With an incorrectly entered code I expect to for `AuthenticationException.isMultifactorCodeInvalid` to be `true`.

This PR just adds another string match.

### References

image

### Testing

Test logging in with mobile and SMS code, entering the code incorrectly. `AuthenticationException.isMultifactorCodeInvalid` should be true.

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java
@@ -200,7 +200,9 @@ public class AuthenticationException extends Auth0Exception {
 
     /// When MFA code sent is invalid or expired
     public boolean isMultifactorCodeInvalid() {
-        return "a0.mfa_invalid_code".equals(code) || "invalid_grant".equals(code) && "Invalid otp_code.".equals(description);
+        return "a0.mfa_invalid_code".equals(code) 
+            || "invalid_grant".equals(code) && "Invalid otp_code.".equals(description)
+            || "invalid_grant".equals(code) && "Wrong phone number or verification code.".equals(description);
     }
 
     /// When password used for SignUp does not match connection's strength requirements.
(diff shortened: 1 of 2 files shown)


lbalmaceda:
@nicbell thanks for this. I assume this changed at some point to prevent user enumeration. I've checked on my side using email and I receive a different error message. Can you please update the PR to add support to this other description as well? Thanks

image

nicbell:
@lbalmaceda thanks, I've added email OTP and tests for both.

lbalmaceda (approved):

lbalmaceda:
Thanks for making the changes. I imagine this can be put into the next release train, in one week from now.

lbalmaceda:
@nicbell Patch released!

lbalmaceda:
Having a second look at this after released. The `isMultifactorCodeInvalid` method is meant to be used for MFA flows, and while I understand that Passwordless could be confused with that, these are indeed different flows.

We already had a method that related to this error case. It is currently missing the Passwordless error descriptions so it's not currently matching that. It's more accurate if we move this PR's changes to the [isInvalidCredentials](https://github.com/nicbell/Auth0.Android/blob/73cd65c05ef25e0894c663a90ef30014ce621832/auth0/src/main/java/com/auth0/android/authentication/AuthenticationException.java#L230-L232) method instead. I'll make the changes and ship that as another patch release. Just wanted you give you a heads up, since you're the only one that reached out for this use case, so you can update your usage.

PR #381, opened 2020-12-03
Prefer robolectric.properties instead of @Config

lbalmaceda:
Removes usage of `@Config` to set up the SDK version that Robolectric will use on a given test, in favor of [the recommended approach](http://robolectric.org/configuring/) of using the `robolectric.properties`. 

There are a few tests that kept it, as they check specifically code for one particular version.


lbalmaceda:
These changes need to be done as part of #380

PR #386, opened 2020-12-11
Only support OIDC Conformant Authentication Flow

jimmyjames:
### Changes

This change updates the SDK to only support OIDC-compliant flows and used endpoints during authentication. Behaviorally, this change makes the default behavior of the SDK as if `oidcConformant` flag were true. Logic handling that only supported the non-OIDC-compliant cases has been removed, and tests have been updated or removed.

#### Removed methods

- `Auth0#setOIDCConformant(boolean enabled)`
- `Auth0#isOIDCConformant()`

#### Removed classes

- `AlgorithmNameVerifier` (package-private). No longer needed as only RS256-signed tokens are supported.

### Checklist


auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
```diff
@@ -41,7 +41,11 @@ void verify(@NonNull JWT token) throws TokenValidationException {
 
     private void checkAlgorithm(String tokenAlgorithm) throws TokenValidationException {
         if (!supportedAlgorithms.contains(tokenAlgorithm)) {
-            throw new TokenValidationException(String.format("Signature algorithm of \"%s\" is not supported. Expected the ID token to be signed with any of %s.", tokenAlgorithm, supportedAlgorithms));
+            if (supportedAlgorithms.size() == 1) {
+                throw new TokenValidationException(String.format("Signature algorithm of \"%s\" is not supported. Expected the ID token to be signed with %s.", tokenAlgorithm, supportedAlgorithms.get(0)));
+            } else {
+                throw new TokenValidationException(String.format("Signature algorithm of \"%s\" is not supported. Expected the ID token to be signed with any of %s.", tokenAlgorithm, supportedAlgorithms));
+            }
```
jimmyjames, line 48: I kept the supported algorithms as an array, in the event additional signing algorithms are supported. I did update the exception message for the usual case, for better readability. That said, I'm open to changing it to a single string if that is preferred.

V2_MIGRATION_GUIDE.md
```diff
@@ -6,6 +6,14 @@
 
 - Android API version 21 or later.
 
+## OIDC only
+
+Verson 2 only supports OIDC-compliant applications. When using this SDK to authenticate users, only the current Auth0 authentication pipeline and endpoints will be used.
```
lbalmaceda, line 11:
```suggestion
Version 2 only supports OIDC-Conformant applications. When using this SDK to authenticate users, only the current Auth0 authentication pipeline and endpoints will be used.
```

lbalmaceda: Let's use the same wording as the dashboard -> Comformant

auth0/src/main/java/com/auth0/android/Auth0.java
```diff
@@ -41,15 +40,10 @@
  * <pre>{@code
  * Auth0 auth0 = new Auth0("YOUR_CLIENT_ID", "YOUR_DOMAIN");
  * }</pre>
+ *
+ * This SDK only supports OIDC Conformant clients, and will use Auth0's current authentication pipeline.
  * It is strongly encouraged that this SDK be used in OIDC Conformant mode.
```
lbalmaceda, line 45: this line here no longer applies

auth0/src/main/java/com/auth0/android/provider/SignatureVerifier.java
lbalmaceda: I'm not aware of any plans to support further algorithms. That said, if you rather keep a one-item constant array instead of a constant string, that's OK

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.java
```diff
@@ -202,24 +201,6 @@ public void shouldLoginWithMFAOTPCode() throws Exception {
...
-
-        final RecordedRequest request = mockAPI.takeRequest();
-        assertThat(request.getHeader("Accept-Language"), is(getDefaultLocale()));
-        Map<String, String> body = bodyFromRequest(request);
-        assertThat(body, hasEntry("connection", MY_CONNECTION));
-        assertThat(body, not(hasKey("realm")));
-    }
-
     @Test
     public void shouldLoginWithUserAndPasswordSync() throws Exception {
```
lbalmaceda, line 205: I'll add a card to track these Async/Sync test variations. We should definitely cover them but on the networking client class (generic) rather than for every available endpoint. Tracking in SDK-2203.

lbalmaceda (changes requested):
Looks fine. A few comments left, also, when it comes to documenting the flag name or the new behavior, let's use what the dashboard uses -> OIDC-Conformant
image

V2_MIGRATION_GUIDE.md
jimmyjames: 👍 pushed a new commit to use consistent terminology.

auth0/src/main/java/com/auth0/android/Auth0.java
jimmyjames: 👍 removed

lbalmaceda (approved):

PR #389, opened 2020-12-15
Always use Authorization Code + PKCE 

jimmyjames:
### Changes

This change removes the ability to use the `WebAuthProvider` to authenticate using any flow other than Authorization Code + PKCE. 

**Removed (public) methods:**

- `WebAuthProvider.Builder#withResponseType(@ResponseType int type)` has been removed, as the `code` response type will be now be supported.

Logic that handled non-code response type processing has been removed. Tests were either removed where there was already coverage for the code response type in place, and updated where there was not.

### Testing

In addition to updating unit tests to accommodate this change, I verified successful login using the sample app with these changes.

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
@@ -68,6 +68,7 @@ class OAuthManager extends ResumableManager {
         this.account = account;
         this.callback = callback;
         this.parameters = new HashMap<>(parameters);
+        this.parameters.put(KEY_RESPONSE_TYPE, RESPONSE_TYPE_CODE);
         this.apiClient = new AuthenticationAPIClient(account);
         this.ctOptions = ctOptions;
     }
@@ -104,7 +105,7 @@ class OAuthManager extends ResumableManager {
         try {
             PKCE.throwIfNotAvailable();
         } catch (Throwable t) {
-            callback.onFailure(new AuthenticationException("Algorithims required to use PKCE are not available on this device. Authentication can not be done securely.", t));
+            callback.onFailure(new AuthenticationException("Algorithms required to use PKCE are not available on this device. Authentication can not be done securely.", t));
             return true;
         }
 
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
@@ -99,16 +99,6 @@ class OAuthManager extends ResumableManager {
     @SuppressWarnings("ConstantConditions")
     @Override
     boolean resume(AuthorizeResult result) {
-        // Should never happen where the algorithms required are not present, but just in case it does,
-        // this will allow us to at least call the error handler callback with an exception including
-        // the cause
-        try {
-            PKCE.throwIfNotAvailable();
-        } catch (Throwable t) {
-            callback.onFailure(new AuthenticationException("Algorithms required to use PKCE are not available on this device. Authentication can not be done securely.", t));
-            return true;
-        }
-
         if (!result.isValid(requestCode)) {
             Log.w(TAG, "The Authorize Result is invalid.");
             return false;
@@ -253,12 +243,6 @@ class OAuthManager extends ResumableManager {
     }
 
     private void addPKCEParameters(Map<String, String> parameters, String redirectUri) {
-        // If PKCE is not available on the device (should never  happen), just return here and
-        // we will fail later so we can call the error callback and not cause a hard failure
-        if (!PKCE.isAvailable()) {
-            return;
-        }
-
         createPKCE(redirectUri);
         String codeChallenge = pkce.getCodeChallenge();
         parameters.put(KEY_CODE_CHALLENGE, codeChallenge);
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
@@ -98,6 +98,16 @@ class OAuthManager extends ResumableManager {
     @SuppressWarnings("ConstantConditions")
     @Override
     boolean resume(AuthorizeResult result) {
+        // Should never happen where the algorithms required are not present, but just in case it does,
+        // this will allow us to at least call the error handler callback with an exception including
+        // the cause
+        try {
+            PKCE.throwIfNotAvailable();
+        } catch (Throwable t) {
+            callback.onFailure(new AuthenticationException("Algorithims required to use PKCE are not available on this device. Authentication can not be done securely.", t));
+            return true;
+        }
+
         if (!result.isValid(requestCode)) {
             Log.w(TAG, "The Authorize Result is invalid.");
             return false;
@@ -122,12 +132,6 @@ class OAuthManager extends ResumableManager {
             assertValidState(parameters.get(KEY_STATE), values.get(KEY_STATE));
         } catch (AuthenticationException e) {
             callback.onFailure(e);
-            return true;
-        }
-
-        if (!shouldUsePKCE()) {
-            // PKCE required but not available. This is not an expected scenario.
-            callback.onFailure(new AuthenticationException("PKCE not available on this device."));
             return true;
         }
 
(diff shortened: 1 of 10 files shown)


auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -131,56 +125,21 @@ boolean resume(AuthorizeResult result) {
...
-                @Override
-                public void onFailure(@NonNull Auth0Exception error) {
-                    AuthenticationException wrappedError = new AuthenticationException(ERROR_VALUE_ID_TOKEN_VALIDATION_FAILED, error);
-                    callback.onFailure(wrappedError);
-                }
-            });
-            return true;
-        }
-
         if (!shouldUsePKCE()) {
```
lbalmaceda, line 128: Do we need to keep a check for something that is documented to never happen? How can a developer recover the authentication from this unexpected scenario?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -289,30 +248,22 @@ private Uri buildAuthorizeUri() {
...
-        if (!shouldUsePKCE()) {
-            return;
-        }
         try {
             createPKCE(redirectUri);
             String codeChallenge = pkce.getCodeChallenge();
             parameters.put(KEY_CODE_CHALLENGE, codeChallenge);
             parameters.put(KEY_CODE_CHALLENGE_METHOD, METHOD_SHA_256);
             Log.v(TAG, "Using PKCE authentication flow");
         } catch (IllegalStateException e) {
```
lbalmaceda, line 257: what conditions are required for this exception to be thrown?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -335,18 +286,6 @@ private boolean shouldUsePKCE() {
         return parameters.containsKey(KEY_RESPONSE_TYPE) && parameters.get(KEY_RESPONSE_TYPE).contains(RESPONSE_TYPE_CODE) && PKCE.isAvailable();
```
lbalmaceda, line 286: On the migration notes you mention code+PKCE is the only supported flow. Seems like this method is no longer relevant.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
```diff
@@ -170,7 +169,7 @@ public void start(@NonNull Context context, @NonNull VoidCallback callback) {
             //Default values
             this.scheme = "https";
             this.ctOptions = CustomTabsOptions.newBuilder().build();
-            withResponseType(ResponseType.CODE);
+            this.values.put(KEY_RESPONSE_TYPE, RESPONSE_TYPE_CODE);
```
lbalmaceda, line 172: Probably something we can hardcode inside the OauthManager class instead?

lbalmaceda (changes requested):

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
jimmyjames: Given it is documented to never happen, we could remove the check. That would mean if for some reason SHA_256 is not supported, and an exception is thrown when creating the code challenge as part of creating the `PKCE` instance, there would be a hard failure. But if it can never really happen, we can just allow it to propagate if for some reason it does happen.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
jimmyjames: When creating an instance of `PKCE`, it generates the code challenge, which wraps and rethrows if it cannot obtain the signature:

```java
try {
        MessageDigest md = MessageDigest.getInstance(SHA_256);
        md.update(input, 0, input.length);
        signature = md.digest();
} catch (NoSuchAlgorithmException e) {
        Log.e(TAG, "Failed to get SHA-256 signature", e);
        throw new IllegalStateException("Failed to get SHA-256 signature", e);
}
```

Previously this logged an error that PKCE isn't available, and instead will fall back to the token response_type. Related to your other questions about how to best handle something that should never happen, and cannot be recovered from, we could just ignore this exception and let it propagate if you prefer.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
jimmyjames: We could add the response type in `OAuthManager` when we add the PKCE code challenge params. I kept them here since the values are used for other auth parameters, but can move them if you prefer.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
jimmyjames: Related to how to handle something that should never happen, we could remove it along with any attempt to handle a non-recoverable situation that should never happen. I suppose the question is if we should send the exception back on the error callback instead of just letting the exception propagate.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
lbalmaceda: If we keep it, we should rename the method since there is no choice other than "using PKCE". 

The questions left are: what the developers can do when this error is raised? and how can they differentiate this error from any other error also raised by this sdk?

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.java
lbalmaceda: Given that the OAuthManager is the ultimate class that starts this, that one should be setting the "unchangeable" parameters. That would be, for example, the code flow required parameters. I'd prefer it to be moved internally.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -98,6 +98,16 @@ void startAuthentication(Activity activity, String redirectUri, int requestCode)
     @SuppressWarnings("ConstantConditions")
     @Override
     boolean resume(AuthorizeResult result) {
+        // Should never happen where the algorithms required are not present, but just in case it does,
```
lbalmaceda, line 101: I'd move this check to the WebAuthProvider Builder constructor, or at least the `start` method. Failing faster than having to wait to come back from the browser sounds better IMO.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -248,15 +252,17 @@ private Uri buildAuthorizeUri() {
...
     private void addPKCEParameters(Map<String, String> parameters, String redirectUri) {
-        try {
-            createPKCE(redirectUri);
-            String codeChallenge = pkce.getCodeChallenge();
-            parameters.put(KEY_CODE_CHALLENGE, codeChallenge);
-            parameters.put(KEY_CODE_CHALLENGE_METHOD, METHOD_SHA_256);
-            Log.v(TAG, "Using PKCE authentication flow");
-        } catch (IllegalStateException e) {
-            Log.e(TAG, "Some algorithms aren't available on this device and PKCE can't be used for authentication, which is required.", e);
+        // If PKCE is not available on the device (should never  happen), just return here and
```
lbalmaceda, line 255: Let's keep the check-in a single place. Given that OAuthManager is only used by the WebAuthProvider class (for login and logout), let's remove this from here and keep it in the login scenario outside.

auth0/src/main/java/com/auth0/android/provider/PKCE.java
```diff
@@ -112,14 +113,29 @@ public static boolean isAvailable() {
...
+     * the exception stack trace back to the error handler.
+     *
+     * @throws {@linkplain RuntimeException} if the device does not support PKCE
+     */
+    public static void throwIfNotAvailable() throws IllegalStateException {
+        checkIfAvailable(new AlgorithmHelper());
+    }
+
     @VisibleForTesting
     static boolean isAvailable(@NonNull AlgorithmHelper algorithmHelper) {
```
lbalmaceda, line 128: the class is already package-private. What is the reason this method got split into two private and public methods?

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
```diff
@@ -104,6 +98,16 @@ void startAuthentication(Activity activity, String redirectUri, int requestCode)
     @SuppressWarnings("ConstantConditions")
     @Override
     boolean resume(AuthorizeResult result) {
+        // Should never happen where the algorithms required are not present, but just in case it does,
+        // this will allow us to at least call the error handler callback with an exception including
+        // the cause
+        try {
+            PKCE.throwIfNotAvailable();
+        } catch (Throwable t) {
+            callback.onFailure(new AuthenticationException("Algorithims required to use PKCE are not available on this device. Authentication can not be done securely.", t));
```
lbalmaceda, line 107: typo. Also, is this change unit tested?

```suggestion
            callback.onFailure(new AuthenticationException("Algorithms required to use PKCE are not available on this device. Authentication can not be done securely.", t));
```

lbalmaceda (changes requested):

auth0/src/test/java/com/auth0/android/authentication/AuthenticationExceptionTest.java
```diff
@@ -337,4 +337,11 @@ public void shouldHaveMissingBrowserApp() {
         assertThat(ex.isBrowserAppNotAvailable(), is(true));
     }
 
+    @Test
+    public void shouldHavePKCENotAvailalbe() {
```
lbalmaceda, line 341:
```suggestion
    public void shouldHavePKCENotAvailable() {
```

lbalmaceda (approved):

auth0/src/main/java/com/auth0/android/provider/OAuthManager.java
jimmyjames: Updated to throw an exception if PKCE is not available when starting login.

PR #396, opened 2020-12-21
Remove AuthorizableRequest interface

lbalmaceda:
The class was adding a helper `setBearer` method to set the authorization header to use a JWT. This would now only apply for the extensibility of the `UsersAPIClient`, which will be phased out and replaced with documentation.

In order to achieve the same, you can add the header to any `Request` instance this way:

```kotlin
request.addHeader("Authorization", "Bearer TOKEN_HERE")
```


auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.java
```diff
@@ -158,13 +157,6 @@ public void onFailure(Request request, IOException e) {
         return this;
     }
 
-    @NonNull
-    @Override
-    public AuthorizableRequest<T, U> setBearer(@NonNull String jwt) {
```
lbalmaceda, line 163: this is covered by the `addHeader` tests above

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.java
```diff
@@ -161,12 +161,6 @@ public void shouldAddParameters() {
         assertThat(result, IsMapContaining.hasEntry("asd", (Object) "123"));
     }
 
-    @Test
-    public void shouldSetBearer() {
```
lbalmaceda, line 165: remove as this is already covered by the `addHeader` tests above

jimmyjames (approved):

PR #401, opened 2020-12-22
Allow to set a custom networking client instance [SDK-2214]

lbalmaceda:
### Changes

This change adds the ability to pass a custom networking client to:

- AuthenticationAPIClient
- UsersAPIClient
- WebAuthProvider#login

### References

See SDK-2214

### Testing
I've added unit tests

### DIFF
--- a/auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
+++ b/auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
@@ -1135,7 +1135,7 @@ public class WebAuthProviderTest {
     }
 
     @Test
-    public void shouldResumeUsingCustomNetworkingClient() throws Exception {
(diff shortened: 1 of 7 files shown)


auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.java
```diff
@@ -1120,7 +1132,7 @@ public Object answer(InvocationOnMock invocation) {
     }
 
     @Test
-    public void shouldResumeLoginWithIntentWithCodeGrant() throws Exception {
+    public void shouldResumeUsingCustomNetworkingClient() throws Exception {
```
jimmyjames, line 1135: This test name and the one below should be switched, I think?

jimmyjames (changes requested):

jimmyjames (approved):

PR #405, opened 2020-12-23
Remove doNotSendTelemetry method

jimmyjames:
### Changes

- Removed `Auth0#doNotSendTelemetry`
- If empty values passed to `Auth0UserAgent`, default values will be used for the name and version
- Adds no-arg constructor to `Auth0UserAgent`, which will construct an instance using default values
- Updates `Auth0` to use the new no-arg constructor, removing the need to specify the values there

### Testing

Unit tests added for the new constructor as well as the behavior to use defaults if empty values passed.

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/Auth0.java
+++ b/auth0/src/main/java/com/auth0/android/Auth0.java
@@ -30,7 +30,6 @@ import android.content.Context;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
-import com.auth0.android.auth0.BuildConfig;
 import com.auth0.android.util.Auth0UserAgent;
 import com.squareup.okhttp.HttpUrl;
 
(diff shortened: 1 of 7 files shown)


auth0/src/main/java/com/auth0/android/Auth0.java
```diff
@@ -183,20 +182,11 @@ public int getWriteTimeoutInSeconds() {
      * Setter for the user agent info to send in every request to Auth0.
      *
      * @param auth0UserAgent to send in every request to Auth0.
-     * @see #doNotSendAuth0UserAgent()
      */
     public void setAuth0UserAgent(@Nullable Auth0UserAgent auth0UserAgent) {
```
lbalmaceda, line 186: Let's make the param non-null

```suggestion
    public void setAuth0UserAgent(@NonNull Auth0UserAgent auth0UserAgent) {
```

lbalmaceda (changes requested):

auth0/src/main/java/com/auth0/android/Auth0.java
jimmyjames: Updated in new commit 👍

lbalmaceda (approved):

PR #409, opened 2020-12-29
Update Robolectric to 4.4

lbalmaceda:
### Changes

Robolectric changed the Looper implementation and now has a Mode that will pause until we explicitly tell it to continue. This PR bumps the version to the latest minor available and fixes the tests where different threads were used (network requests tests). Hopefully, this will get rid of the randomly failing CI. 

**Note to the reviewer:** Might be easier to follow the changes if each commit is inspected individually. 

### References

If in doubt, see http://robolectric.org/blog/2019/06/04/paused-looper/.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,37 +2,36 @@ version: 2
 jobs:
   build:
     environment:
-      GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.compiler.execution.strategy=in-process
-      _JAVA_OPTIONS: -Xms512m -Xmx1024m
+      GRADLE_OPTS: -Dkotlin.compiler.execution.strategy=in-process
     docker:
-      - image: circleci/android:api-29
+      - image: circleci/android:api-30
 
     steps:
-    - checkout
-    
-    - restore_cache:
-        keys:
-        - v2-dep-{{ checksum "build.gradle" }}-{{ checksum  "auth0/build.gradle" }}
-    
-    - run:
-        name: Run checks
-        command: ./gradlew clean test jacocoTestReport lint --continue --console=plain --no-daemon --max-workers=2
+      - checkout
 
-    - save_cache:
-        key: v2-dep-{{ checksum "build.gradle" }}-{{ checksum  "auth0/build.gradle" }}
-        paths:
-          - ~/.gradle
-          - ~/.android
-          - /usr/local/android-sdk-linux/extras
+      - restore_cache:
+          keys:
+            - v2-dep-{{ checksum "build.gradle" }}-{{ checksum  "auth0/build.gradle" }}
 
-    - store_artifacts:
-        path: auth0/build/reports
-        destination: reports
+      - run:
+          name: Run checks
+          command: ./gradlew clean test jacocoTestReport lint --continue --console=plain --max-workers=2
 
-    - store_test_results:
-        path: auth0/build/test-results
+      - save_cache:
+          key: v2-dep-{{ checksum "build.gradle" }}-{{ checksum  "auth0/build.gradle" }}
+          paths:
+            - ~/.gradle
+            - ~/.android
+            - /usr/local/android-sdk-linux/extras
 
-    - run:
+      - store_artifacts:
+          path: auth0/build/reports
+          destination: reports
+
+      - store_test_results:
+          path: auth0/build/test-results
+
+      - run:
           name: Upload Coverage
           when: on_success
           command: bash <(curl -s https://codecov.io/bash) -Z -C $CIRCLE_SHA1
--- a/auth0/build.gradle
+++ b/auth0/build.gradle
@@ -102,6 +102,6 @@ dependencies {
     testImplementation 'org.mockito:mockito-core:3.6.28'
     testImplementation 'com.squareup.okhttp:mockwebserver:2.7.5'
     testImplementation 'com.jayway.awaitility:awaitility:1.7.0'
-    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'org.robolectric:robolectric:4.4'
     testImplementation 'androidx.test.espresso:espresso-intents:3.3.0'
 }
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -2,37 +2,36 @@ version: 2
 jobs:
   build:
     environment:
-      GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.compiler.execution.strategy=in-process
```
lbalmaceda, line 5: Only two changes on this file:
1) this now defaults to what Docker reports as available in the CI. Also, some were specified through the `build.gradle` files instead

.circleci/config.yml
```diff
@@ -2,37 +2,36 @@ version: 2
...
-      _JAVA_OPTIONS: -Xms512m -Xmx1024m
+      GRADLE_OPTS: -Dkotlin.compiler.execution.strategy=in-process
     docker:
-      - image: circleci/android:api-29
+      - image: circleci/android:api-30
```
lbalmaceda, line 7: 2) bumps to latest image

auth0/build.gradle
```diff
@@ -63,25 +63,28 @@ android {
...
-        ]
-    }
     lintOptions {
         htmlReport true
         abortOnError true
     }
     testOptions {
         unitTests {
+            all {
+                maxHeapSize = "1024m"
```
lbalmaceda, line 73: alternative for `-Xmx1024m`

auth0/build.gradle
```diff
@@ -63,25 +63,28 @@ android {
...
+            }
             includeAndroidResources = true
         }
     }
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    kotlinOptions {
```
lbalmaceda, line 82: moved from above

auth0/build.gradle
```diff
@@ -102,6 +105,6 @@ dependencies {
     testImplementation 'org.mockito:mockito-core:3.6.28'
     testImplementation 'com.squareup.okhttp:mockwebserver:2.7.5'
     testImplementation 'com.jayway.awaitility:awaitility:1.7.0'
-    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'org.robolectric:robolectric:4.4'
```
lbalmaceda, line 108: actual testing library bump

auth0/src/test/java/com/auth0/android/management/UsersAPIClientTest.java
```diff
@@ -48,7 +48,6 @@
 import com.google.gson.reflect.TypeToken;
 import com.squareup.okhttp.mockwebserver.RecordedRequest;
 
-import org.apache.tools.ant.filters.StringInputStream;
```
lbalmaceda, line 51: wrong import being used. This class doesn't exist in this project, so the usage was replaced with a `ByteArrayInputStream` in 4 files.

auth0/src/test/java/com/auth0/android/management/UsersAPIClientTest.java
```diff
@@ -128,14 +129,16 @@ public void tearDown() throws Exception {
...
+        String jsonResponse = "{\"id\":\"undercover\"}";
+        InputStream inputStream = new ByteArrayInputStream(jsonResponse.getBytes());
         ServerResponse response = new ServerResponse(200, inputStream, Collections.emptyMap());
         NetworkingClient networkingClient = mock(NetworkingClient.class);
         when(networkingClient.load(anyString(), any(RequestOptions.class))).thenReturn(response);
         UsersAPIClient client = new UsersAPIClient(account, "token.token", networkingClient);
 
         Request<UserProfile, ManagementException> request = client.getProfile("undercover");
         request.execute();
+        ShadowLooper.idleMainLooper();
```
lbalmaceda, line 141: This line is required to be run after every request is executed or scheduled to be run. Because Robolectric blocks the thread until we say "idle", when it will release the lock and continue the execution.

auth0/src/test/java/com/auth0/android/provider/RedirectActivityTest.java
```diff
@@ -49,13 +52,13 @@ public void shouldLaunchAuthenticationActivityWithDataOnSuccess() {
         createActivity(resultIntent);
         activityController.create().start().resume();
 
-        Intent authenticationIntent = shadowOf(activity).getNextStartedActivity();
```
lbalmaceda, line 52: Use cases tested using a shadow that keeps only what we need to assert: that it started a specific activity and that it was then finished.

auth0/src/test/java/com/auth0/android/request/internal/ThreadSwitcherShadow.java
```diff
@@ -16,10 +17,10 @@
 @SuppressWarnings({"unused", "RedundantSuppression"})
 public class ThreadSwitcherShadow {
 
-    private final RoboExecutorService executor;
+    private final InlineExecutorService executor;
```
lbalmaceda, line 20: `RoboExecutorService` does not work with Mode.PAUSED. Updated to use `InlineExecutorService`, which behaves the same.

auth0/src/test/java/com/auth0/android/request/internal/ThreadSwitcherShadow.java
```diff
@@ -1,5 +1,6 @@
 package com.auth0.android.request.internal;
 
+import org.robolectric.android.util.concurrent.InlineExecutorService;
 import org.robolectric.android.util.concurrent.RoboExecutorService;
```
Widcket, line 4: Remove this now unused import.

Widcket (changes requested):
Looks good, just remove an unused import.

auth0/src/test/java/com/auth0/android/request/internal/ThreadSwitcherShadow.java
lbalmaceda: Done

Widcket (approved):

PR #411, opened 2021-01-05
Make callback interfaces take non-null arguments

lbalmaceda:
### Changes
Before this PR, the `onSuccess` method of the callback implementations received a "nullable" parameter as the result. This would never be the case, unless for representing a success that doesn't have a result (e.g. reset password method). This should help Kotlin callers to avoid null-checking the received value, when we know it's never going to be null. A failure when parsing the JSON would probably fire an exception via the callback's on error instead.


auth0/src/main/java/com/auth0/android/callback/BaseCallback.kt
```diff
@@ -35,5 +35,5 @@ public interface BaseCallback<T, U : Auth0Exception> : Callback<U> {
      *
      * @param payload Request payload or null
```
jimmyjames: update javadoc if it cannot be null

jimmyjames (changes requested):

jimmyjames (approved):

PR #412, opened 2021-01-05
Remove Callback interface [SDK-2185]

jimmyjames:
### Changes

Prior to this PR, the `Callback` interface defined a single method, `onFailure`. `BaseCallback` extended this interface, and defined the method `onSuccess`. This change removes the `Callback` interface, and moves the `onFailure` method to the `BaseCallback` method.

#### Alternatives considered

The initial plan as I started this refactor was to remove the `BaseCallback` interface, and instead merge `onSuccess` into `Callback`. I started the other way to make refactoring easier and more incremental, with the plan to then rename `BaseCallback` to `Callback` to achieve the same result. However, it became clear that merging `Callback` into `BaseCallback` has some advantages:

- No need for callers to change all references to `BaseCallback` to `Callback`
- This change is smaller

Additionally, I don't think `BaseCallback` is a bad name, as it is the base of all callback interfaces. That, combined with the reasons above, are why I removed `Callback` instead of `BaseCallback`.

### Checklist


auth0/src/main/java/com/auth0/android/callback/BaseCallback.kt
```diff
@@ -26,14 +26,20 @@ package com.auth0.android.callback
 import com.auth0.android.Auth0Exception
 
 /**
- * Callback that receives a single value on success.
+ * Interface for all callbacks used with Auth0 API clients
  */
-//TODO [SDK-2185]: Merge this interface into Callback
-public interface BaseCallback<T, U : Auth0Exception> : Callback<U> {
+public interface BaseCallback<T, U : Auth0Exception> {
```
lbalmaceda, line 31: Let's rename `BaseCallback` into `Callback`; and create a new no-op interface named `BaseCallback` that extends `Callback`. That way, while we can guide devs into using the preferred "Callback" class name, we won't introduce breaking changes.

Also, let's `@Deprecated` the introduced `BaseCallback` interface so those still using it can receive the _warning_ to migrate to `Callback`.

lbalmaceda (changes requested):

jimmyjames:
Replaced with https://github.com/auth0/Auth0.Android/pull/416

PR #420, opened 2021-01-08
Update readme in preparation of v2

lbalmaceda:
This PR also orders some entries into the FAQ (second commit), as these are very specific use cases that just make the setup section longer than required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Auth0.Android
+# Auth0.Android (BETA)
 
 [![CircleCI](https://img.shields.io/circleci/project/github/auth0/Auth0.Android.svg?style=flat-square)](https://circleci.com/gh/auth0/Auth0.Android/tree/master)
 [![Coverage Status](https://img.shields.io/codecov/c/github/auth0/Auth0.Android/master.svg?style=flat-square)](https://codecov.io/github/auth0/Auth0.Android)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -775,6 +652,61 @@ android {
...
+            </intent-filter>
+        </activity>
+
+        <!-- ... -->
+
+    </application>
+</manifest>
+```
+
+Remember that if you request a different scheme you must replace the above `android:scheme` property value and initialize the provider with the new scheme. Read [this section](#a-note-about-app-deep-linking) to learn more.
```
jimmyjames, line 691:
```suggestion
Recall that if you request a different scheme, you must replace the above `android:scheme` property value and initialize the provider with the new scheme. Read [this section](#a-note-about-app-deep-linking) to learn more. 
```

README.md
```diff
@@ -775,6 +652,61 @@ android {
...
+    </application>
+</manifest>
+```
+
+Remember that if you request a different scheme you must replace the above `android:scheme` property value and initialize the provider with the new scheme. Read [this section](#a-note-about-app-deep-linking) to learn more. 
+
+
+* Is the Web Authentication module setup optional?
+
+If you don't plan to use the _Web Authentication_ feature you will notice that the compiler will still prompt you to provide the `manifestPlaceholders` values, since the `RedirectActivity` included in this library will require them and the Gradle tasks won't be able to run without them.
```
jimmyjames, line 696:
```suggestion
If you don't plan to use the _Web Authentication_ feature, you will notice that the compiler will still prompt you to provide the `manifestPlaceholders` values, since the `RedirectActivity` included in this library will require them, and the Gradle tasks won't be able to run without them. 
```

jimmyjames (changes requested):

jimmyjames (approved):

PR #421, opened 2021-01-08
Update WebAuthProvider#start required Context

lbalmaceda:
### Changes
Before we required an `Activity` context because we were launching an internal activity for the web view scenario (legacy). This is no longer the case, so the required context can be changed to just Context. 

**This is not a breaking change.**

### DIFF
--- a/auth0/src/test/java/com/auth0/android/provider/OAuthManagerTest.java
+++ b/auth0/src/test/java/com/auth0/android/provider/OAuthManagerTest.java
@@ -1,32 +1,16 @@
 package com.auth0.android.provider;
 
-import com.auth0.android.Auth0;
 import com.auth0.android.authentication.AuthenticationException;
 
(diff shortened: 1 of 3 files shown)


auth0/src/test/java/com/auth0/android/provider/OAuthManagerTest.java
```diff
@@ -1,33 +1,17 @@
...
 import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
 import org.robolectric.RobolectricTestRunner;
 
 
 @RunWith(RobolectricTestRunner.class)
 public class OAuthManagerTest {
 
-    @Mock
```
lbalmaceda, line 19: all these are never used from the tests

jimmyjames (approved):

PR #422, opened 2021-01-08
Enforce the "openid" scope for the WebAuthProvider

lbalmaceda:
### Changes

This PR enforces the scope of "openid", required as part of the ID Token validation that happens during Web Auth.

### DIFF
--- a/auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
+++ b/auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
@@ -289,7 +289,7 @@ public class WebAuthProviderTest {
         val uri =
             intentCaptor.firstValue.getParcelableExtra<Uri>(AuthenticationActivity.EXTRA_AUTHORIZE_URI)
         MatcherAssert.assertThat(uri, `is`(notNullValue()))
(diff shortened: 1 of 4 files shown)


auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -322,7 +322,7 @@ public class WebAuthProviderTest {
         val uri =
             intentCaptor.firstValue.getParcelableExtra<Uri>(AuthenticationActivity.EXTRA_AUTHORIZE_URI)
         MatcherAssert.assertThat(uri, `is`(notNullValue()))
-        MatcherAssert.assertThat(uri, UriMatchers.hasParamWithValue("scope", "profile super_scope"))
+        MatcherAssert.assertThat(uri, UriMatchers.hasParamWithValue("scope", "profile super_scope openid"))
```
lbalmaceda, line 325: since the scope is now enforced, I needed to update these tests

jimmyjames (approved):

lbalmaceda:
@jimmyjames holding off this one until we decide if a different "default scope" should be set when no scope is given by the developer. The enforcement of "openid" will still be made.

jimmyjames (approved):

PR #428, opened 2021-01-12
Migrate default NetworkingClient implementation to use OkHttp

lbalmaceda:
### Changes

This migration was required due to `HttpUrlConnection` not having support for the PATCH HTTP method, used for patching user metadata.


auth0/src/main/java/com/auth0/android/Auth0.kt
```diff
@@ -188,11 +191,11 @@ public open class Auth0 @JvmOverloads constructor(
         if (url == null) {
             return null
         }
-        val normalizedUrl = url.toLowerCase()
+        val normalizedUrl = url.toLowerCase(Locale.ROOT)
```
jimmyjames, line 194: 👍  good fix, saw this in the warnings and was going to address it, good to do it here.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -72,7 +72,7 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
     @JvmOverloads
     public constructor(
         auth0: Auth0,
-        networkingClient: NetworkingClient = DefaultClient(auth0.connectTimeoutInSeconds)
+        networkingClient: NetworkingClient = DefaultClient()
```
jimmyjames, line 75: When I make updates to the DefaultClient constructor params, I'll wire the timeout/logging, etc params into the DefaultClient constructor here.

jimmyjames (approved):

PR #429, opened 2021-01-12
Fix request to patch user metadata

lbalmaceda:
### Changes

The serialization of the `user_metadata` parameter was broken, as the internal classes required a String value to be passed, and the server expected an object / dictionary.

The solution proposed here involves an internal method (not exposed) and a cast to a type we know it's the right expected one, as the `RequestFactory` implementation that provides the request instance is ours as well.

The `RequestOptions` had to be changed to support being passed parameters of the `Any` (`Object`) type. This is this way since if we provide a hidden internal implementation that handles this one scenario and let the interface as it was, we could be leaving out those developers that decide to extend the NetworkingClient and implement their own PATCH requests; as they would still be receiving a String value.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/management/UsersAPIClient.kt
+++ b/auth0/src/main/java/com/auth0/android/management/UsersAPIClient.kt
@@ -28,6 +28,7 @@ import com.auth0.android.Auth0
 import com.auth0.android.Auth0Exception
 import com.auth0.android.authentication.ParameterBuilder
 import com.auth0.android.request.*
+import com.auth0.android.request.internal.BaseRequest
 import com.auth0.android.request.internal.GsonAdapter
 import com.auth0.android.request.internal.GsonAdapter.Companion.forListOf
 import com.auth0.android.request.internal.GsonAdapter.Companion.forMap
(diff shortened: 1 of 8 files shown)


auth0/src/main/java/com/auth0/android/management/UsersAPIClient.kt
```diff
@@ -186,8 +187,12 @@ public class UsersAPIClient @VisibleForTesting(otherwise = VisibleForTesting.PRI
         val userProfileAdapter: JsonAdapter<UserProfile> = GsonAdapter(
             UserProfile::class.java, gson
         )
-        return factory.patch(url.toString(), userProfileAdapter)
-            .addParameter(USER_METADATA_KEY, gson.toJson(userMetadata))
+        val patch = factory.patch(
+            url.toString(),
+            userProfileAdapter
+        ) as BaseRequest<UserProfile, ManagementException>
```
lbalmaceda, line 193: "unsafe cast": would throw if the cast cannot be made. But we know the type our RequestFactor uses internally.

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -40,7 +40,8 @@ public class DefaultClient() : NetworkingClient {
         when (options.method) {
             is HttpMethod.GET -> {
                 // add parameters as query
-                options.parameters.map { urlBuilder.addQueryParameter(it.key, it.value) }
+                options.parameters.filterValues { it is String }
```
lbalmaceda, line 43: when building the query, OKHttp requires us to pass `String?` type for the values. I'm filtering those that are not a String. Since the Request interface requires `String, String` this should not be an issue for Requests built by our API clients.

jimmyjames (approved):

PR #430, opened 2021-01-12
Add timeout and logging configurability to DefaultClient

jimmyjames:
### Changes

This change adds some named constructor parameters to `DefaultClient`, all with default values, to enable easier customization of some more common configurations, without needing to resort to implementing a custom `NetworkingClient`. The configuration options included in this PR are:

- `connectTimeout` - The connect timeout when making requests. Default is 10 seconds
- `readTimeout` - The read timeout when making requests. Default is 10 seconds
- `defaultHeaders` - Any headers that should be sent on all requests. Default is an empty map
- `enableLogging` - Whether to log HTTP request/response info. Default is false

As the configurations listed above are currently available on the `Auth0` class, the API clients have been updated to use these values when constructing the `DefaultClient` if a custom client is not specified.

Changes **not** in this PR, but planned to come soon, are:

- deprecations of the above configuration points on `Auth0`, in favor of using the networking client.
- removal of the `writeTimeoutInSeconds` configuration on `Auth0`. This is too tightly coupled to a specific networking client, and will likely be removed without replacement.
- possible removal of `isTLS12Enforced`. Investigation needs to be done to understand if this is still applicable with API version 21+ and OkHttp4.
- updated README documentation that demonstrates the configuration options of the networking client.

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/Auth0.kt
+++ b/auth0/src/main/java/com/auth0/android/Auth0.kt
@@ -25,6 +25,7 @@ package com.auth0.android
 
 import android.content.Context
 import androidx.annotation.VisibleForTesting
+import com.auth0.android.request.DefaultClient
 import com.auth0.android.util.Auth0UserAgent
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
@@ -88,7 +89,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
-    public var connectTimeoutInSeconds: Int = 10
+    public var connectTimeoutInSeconds: Int = DefaultClient.DEFAULT_TIMEOUT_SECONDS
     /**
      * @return Auth0 request readTimeoutInSeconds
      */
@@ -98,7 +99,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
-    public var readTimeoutInSeconds: Int = 10
+    public var readTimeoutInSeconds: Int = DefaultClient.DEFAULT_TIMEOUT_SECONDS
     /**
      * @return Auth0 request writeTimeoutInSeconds
      */
@@ -108,6 +109,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
+    // TODO - remove this, only expose connect and read timeouts
     public var writeTimeoutInSeconds: Int = 0
 
     /**
--- a/auth0/src/main/java/com/auth0/android/Auth0.kt
+++ b/auth0/src/main/java/com/auth0/android/Auth0.kt
@@ -88,7 +88,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
-    public var connectTimeoutInSeconds: Int = 0
+    public var connectTimeoutInSeconds: Int = 10
     /**
      * @return Auth0 request readTimeoutInSeconds
      */
@@ -98,7 +98,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
-    public var readTimeoutInSeconds: Int = 0
+    public var readTimeoutInSeconds: Int = 10
     /**
      * @return Auth0 request writeTimeoutInSeconds
      */
--- a/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
+++ b/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
@@ -22,8 +22,8 @@ import java.util.concurrent.TimeUnit
  * @param enableLogging whether HTTP request and response info should be logged. This should only be set to `true` for debugging purposes in non-production environments, as sensitive information is included in the logs. Defaults to `false`.
  */
 public class DefaultClient(
-        private val connectTimeout: Long = DEFAULT_TIMEOUT_SECONDS,
-        private val readTimeout: Long = DEFAULT_TIMEOUT_SECONDS,
+        private val connectTimeout: Int = DEFAULT_TIMEOUT_SECONDS,
+        private val readTimeout: Int = DEFAULT_TIMEOUT_SECONDS,
         private val defaultHeaders: Map<String, String> = mapOf(),
         private val enableLogging: Boolean = false
 ) : NetworkingClient {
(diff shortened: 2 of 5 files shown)


auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -14,12 +15,24 @@ import java.util.concurrent.TimeUnit
 
 /**
  * Default implementation of a Networking Client.
+ *
+ * @param connectTimeout the connection timeout, in seconds, to use when executing requests. Default is ten seconds.
+ * @param readTimeout the read timeout, in seconds, to use when executing requests. Default is ten seconds.
+ * @param defaultHeaders any headers that should be sent on all requests. If a specific request specifies a header with the same key as any header in the default headers, the header specified on the request will take precedence. Default is an empty map.
```
lbalmaceda, line 21: this one was left out of the PR description, can you add it there as well, please?
Also, consider the removal of `setUserAgent` methods in both API clients now that we have these.

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -14,12 +15,24 @@ import java.util.concurrent.TimeUnit
...
+        private val defaultHeaders: Map<String, String> = mapOf(),
+        private val enableLogging: Boolean = false
+) : NetworkingClient {
 
     //TODO: receive this via internal constructor parameters
     private val gson: Gson = GsonProvider.buildGson()
-    private var client: OkHttpClient
+
+    @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var okHttpClient: OkHttpClient
```
lbalmaceda, line 35: Please change this into a `val` instead. I'd also add this on the line below to make the setter private. Unless there's a reason to open that?

```
private set
```

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
```diff
@@ -48,6 +51,59 @@ public class DefaultClientTest {
...
+
+        val interceptor: Interceptor = netClient.okHttpClient.interceptors[0]
+        assertThat((interceptor as HttpLoggingInterceptor).level, equalTo(HttpLoggingInterceptor.Level.BODY))
+    }
+
+    @Test
+    public fun shouldHaveDefaultTimeoutValues() {
+        val client = DefaultClient()
+        assertThat(client.okHttpClient.connectTimeoutMillis, equalTo(10 * 1000))
+        assertThat(client.okHttpClient.connectTimeoutMillis, equalTo(10 * 1000))
```
lbalmaceda, line 97: this line should check for the **read** timeout. The same mistake is on the test below. If we make _internal_ the constant, should we use it here as well in replacement of that `10`?

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
```diff
@@ -163,20 +222,21 @@ public class DefaultClientTest {
...
-            hasEntry(
-                equalTo("a-header"),
-                hasItem("b-value")
-            )
-        )
+
+        val requestHeaders = request.headers.toMultimap()
+        headers.forEach { (k, v) ->
+            assertThat(requestHeaders, hasEntry(
+                equalTo(k), hasItem(v)
```
lbalmaceda, line 237: Given that the multimap is `String, List<String>` I think that `hasEntry` is not enough to check that we're actually preferring request headers over global ones., as they could both exist under the same key. We probably should check for `size == 1` as well.

lbalmaceda (changes requested):

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
jimmyjames: Updated to val. No need (or ability) to add `private set` as it is now a val.

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
jimmyjames: Added to PR. description. We can discuss removal of setUserAgent seperately

auth0/src/main/java/com/auth0/android/Auth0.kt
```diff
@@ -88,7 +88,7 @@ public open class Auth0 @JvmOverloads constructor(
      *
      * @param timeout the new timeout value in seconds
      */
-    public var connectTimeoutInSeconds: Int = 0
+    public var connectTimeoutInSeconds: Int = 10
```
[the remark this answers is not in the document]
jimmyjames: Pushed a commit to make it an internal constant on DefaultClient

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
jimmyjames: Yes, updated to check for read. I'm not using the constant because I like the test to specify expectations explicitly, not be tied to how it is implemented. Also if we do change the default, the test should fail IMO as we are changing the behavior.

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
jimmyjames: Updated to check the size.

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -75,16 +84,16 @@ public class DefaultClient() : NetworkingClient {
...
-        client = builder.build()
+        okHttpClient = builder.build()
     }
 
 
-    private companion object {
-        private const val DEFAULT_TIMEOUT_SECONDS: Long = 10
-        private val APPLICATION_JSON_UTF8: MediaType =
+    internal companion object {
+        const val DEFAULT_TIMEOUT_SECONDS: Int = 10
```
jimmyjames, line 95: I changed this to an Int, to enable using it in `Auth0` where the corresponding values are Int's. As the setting in OkHttp should not exceed `Integer.MAX_VALUE`, I think having it as an int makes sense anyways.

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
lbalmaceda: >  I like the test to specify expectations explicitly, not be tied to how it is implemented.

Yes, that's what I prefer as well. Cool 💪

lbalmaceda (approved):

PR #427, opened 2021-01-12
Refactor classes visibility and packages

lbalmaceda:
### Public & non-used classes
- `VoidCallback` was previously used for web auth, no longer used in public APIs. It's now `internal`.

### The internal package
The internal package contains classes not meant to be used directly. With Kotlin, these now make use of the _internal_ visibility modifier.
- `BaseAuthenticationRequest`
- `BaseRequest`
- `RequestFactory`

## Package changes:
These are still used in the AuthenticationAPIClient.
- `com.auth0.android.authentication.request.ProfileRequest` to `com.auth0.android.request.ProfileRequest`
- `com.auth0.android.authentication.request.SignUpRequest` to `com.auth0.android.request.SignUpRequest`

PR #433, opened 2021-01-14
Move to a JSON client singleton

lbalmaceda:
### Changes

`GsonProvider` has been historically constructing instances of the `Gson` client that were not being shared. This PR changes that into using a kotlin object (thread-safe singleton) that will be reused across the entire library. Since this client property has an _internal_ visibility, classes using it were required to be migrated to kotlin. That includes tests.

Depends on #427. Merge that and rebase before reviewing this or review from `97f0099` on.


auth0/src/main/java/com/auth0/android/request/internal/GsonProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.google.gson.GsonBuilder
+import com.google.gson.reflect.TypeToken
+import java.security.PublicKey
+import java.text.SimpleDateFormat
+import java.util.*
+
+internal object GsonProvider {
+    internal val gson: Gson
+    private var sdf: SimpleDateFormat
+    private const val DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
```
jimmyjames, line 17: I thinks this date format could cause issues, as the milliseconds are optional and not always sent by Auth0. We had an issue in `auth0-java` related to this, fixed with https://github.com/auth0/auth0-java/pull/295

jimmyjames (changes requested):

auth0/src/main/java/com/auth0/android/request/internal/GsonProvider.kt
lbalmaceda: The library uses this for the user profile deserialization (users API / get profile) and for storing/recovering credentials from the local storage (we first save the credentials with this format and then recover back, it would not impact this case). 

We are using Gson rather than Jackson so the proposed solution of not specifying any date format and defaulting to the library's format could not apply. I'd like to check if the dates that the server sends in the User Profile object are correctly deserialized.

jimmyjames: It's something to keep in mind in case this does come up - an edge case for sure, but it is possible. Since this isn't new to this PR, we can track it separately and proceed with this PR.

jimmyjames (approved):

PR #435, opened 2021-01-15
Update the Credentials class

lbalmaceda:
### Changes

This class was originally created using an expires_in value. This is no longer possible, and an expiration date should be passed instead. 

You probably will never be instantiating a Credentials on your own. This is a wrapper class intended to contain the tokens received from the server after a successfull authentication.


V2_MIGRATION_GUIDE.md
```diff
@@ -96,6 +96,7 @@ We will not provide support and will change these as required without any previo
 - `UsersAPIClient` can no longer be constructed from a `Context`. Use UsersAPIClient(auth0: Auth0, token: String)` instead. You can create an instance of `Auth0` using a `Context`.
 - `SignupRequest` now requires the second parameter to be an `AuthenticationRequest`.
 - `ProfileRequest` now requires an `AuthenticationRequest` and a `Request<UserProfile, AuthenticationException>`.
+- `Credentials` can no longer be constructed with an "expires in" value. The date of expiration or "expires at" should be given instead. You typically won't be constructing Credentials on your own.
```
jimmyjames, line 99:
```suggestion
- `Credentials` can no longer be constructed with an "expires in" value. The date of expiration or "expires at" should be given instead. It is not expected for an instance of this class to be constructed by consumers of this SDK.
```

jimmyjames (changes requested):

jimmyjames (approved):

PR #452, opened 2021-02-04
Make JsonRequired annotation internal 

lbalmaceda:
### Changes
This annotation is supposed to be internal and was public since 2016. This PR moves it into the `request.internal` package and also changes its visibility to internal / package-private.

These changes will help to reduce the exposure of Gson as a third-party dependency, making it easier to migrate off of it at some point in the future, if desired.

I've tested this with the sample app we ship with this v2 branch. The way I tested it was making a call to /oauth/token to authenticate, then /userinfo to grab the subject, and then /api/v2/users to get the full user profile. The intent was to pull the "UserIdentities" that are annotated with this class.


V2_MIGRATION_GUIDE.md
```diff
@@ -136,6 +136,8 @@ We will not provide support and will change these as required without any previo
 - The `com.auth0.android.request.ErrorBuilder` interface has been removed.
 - The `com.auth0.android.RequestBodyBuildException` class has been removed.
 - The `com.auth0.android.util.CheckHelper` class has been removed.
+- The `com.auth0.android.util.JsonRequired` interface has been removed.
+- The `com.auth0.android.util.JsonREquiredTypeAdapterFactory` class has been removed.
```
jimmyjames, line 140:
```suggestion
- The `com.auth0.android.util.JsonRequiredTypeAdapterFactory` class has been removed.
```

jimmyjames (changes requested):

jimmyjames (approved):

PR #455, opened 2021-02-09
Enforce openid scope on the AuthenticationAPIClient

lbalmaceda:
### Changes

The requests affected by this change include all `/oauth/token` calls but the one that is used to renew the credentials with a Refresh Token. 

The logic is as follows:
- Each of these requests will be set an initial default `scope` of "openid profile email" to match the `WebAuthProvider` default scope.
- If a `scope` value is set by the developer, it will be checked that it includes "openid" and it will append "openid" if it's missing.

This change is to ensure the tokens response from the server is compliant with the OpenID Connect spec. 

This is NOT a breaking change

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
@@ -77,10 +77,9 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
         password: String,
         realmOrConnection: String
     ): AuthenticationRequest {
-        val builder = ParameterBuilder.newBuilder()
+        val parameters = ParameterBuilder.newAuthenticationBuilder()
             .set(USERNAME_KEY, usernameOrEmail)
             .set(PASSWORD_KEY, password)
-        val parameters = builder
             .setGrantType(ParameterBuilder.GRANT_TYPE_PASSWORD_REALM)
             .setRealm(realmOrConnection)
             .asDictionary()
@@ -103,7 +102,7 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
      * @return a request to configure and start that will yield [Credentials]
      */
     public fun login(usernameOrEmail: String, password: String): AuthenticationRequest {
-        val requestParameters = ParameterBuilder.newBuilder()
+        val requestParameters = ParameterBuilder.newAuthenticationBuilder()
             .set(USERNAME_KEY, usernameOrEmail)
             .set(PASSWORD_KEY, password)
             .setGrantType(ParameterBuilder.GRANT_TYPE_PASSWORD)
@@ -129,7 +128,7 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
      * @return a request to configure and start that will yield [Credentials]
      */
     public fun loginWithOTP(mfaToken: String, otp: String): AuthenticationRequest {
-        val parameters = ParameterBuilder.newBuilder()
+        val parameters = ParameterBuilder.newAuthenticationBuilder()
             .setGrantType(ParameterBuilder.GRANT_TYPE_MFA_OTP)
             .set(MFA_TOKEN_KEY, mfaToken)
             .set(ONE_TIME_PASSWORD_KEY, otp)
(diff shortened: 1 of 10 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -155,27 +168,18 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
...
             .set(SUBJECT_TOKEN_KEY, token)
             .set(SUBJECT_TOKEN_TYPE_KEY, tokenType)
             .asDictionary()
-        val credentialsAdapter = GsonAdapter(
-            Credentials::class.java, gson
-        )
-        val request = BaseAuthenticationRequest(factory.post(url.toString(), credentialsAdapter))
-        request.addParameters(parameters)
-        return request
+        return loginWithToken(parameters)
```
lbalmaceda, line 177: most logic was already available on this method, so I simplified this call

auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.kt
```diff
@@ -64,7 +66,7 @@ public class ParameterBuilder private constructor(parameters: Map<String, String
      * @return itself
      */
     public fun setScope(scope: String): ParameterBuilder {
-        return set(SCOPE_KEY, scope)
+        return set(SCOPE_KEY, OidcUtils.includeRequiredScope(scope))
```
lbalmaceda, line 69: when a scope is set through this setter, it will be checked so that it contains "openid"

auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.kt
```diff
@@ -166,15 +168,16 @@ public class ParameterBuilder private constructor(parameters: Map<String, String
...
-         * Creates a new instance of the builder using default values for login request, e.g. 'openid' for scope.
+         * Creates a new instance of the builder using default values for login request, e.g. 'openid profile email' for scope.
          *
          * @return a new builder
          */
         @JvmStatic
         public fun newAuthenticationBuilder(): ParameterBuilder {
             return newBuilder()
-                .setScope(SCOPE_OPENID)
+                .setScope(OidcUtils.DEFAULT_SCOPE)
```
lbalmaceda, line 178: this method will always add as default the scope of "openid profile email". It can be changed later by the dev.

auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
```diff
@@ -55,7 +56,7 @@ internal class OAuthManager(
     }
 
     fun startAuthentication(context: Context, redirectUri: String, requestCode: Int) {
-        addRequiredScope(parameters)
+        OidcUtils.includeDefaultScope(parameters)
```
lbalmaceda, line 59: moved the method to a different class, but it's the same logic (test remains unchanged)

auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.kt
```diff
@@ -31,13 +31,23 @@ internal open class BaseRequest<T, U : Auth0Exception>(
...
 
     override fun addParameter(name: String, value: String): Request<T, U> {
-        options.parameters[name] = value
-        return this
+        val anyValue: Any = if (name == OidcUtils.KEY_SCOPE) {
+            OidcUtils.includeRequiredScope(value)
+        } else {
+            value
+        }
+        return addParameter(name, anyValue)
```
lbalmaceda, line 50: delegates to the `Any` method signature

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -1505,6 +1509,30 @@ public class AuthenticationAPIClientTest {
         assertThat(body, Matchers.hasEntry("client_id", CLIENT_ID))
         assertThat(body, Matchers.hasEntry("refresh_token", "refreshToken"))
         assertThat(body, Matchers.hasEntry("grant_type", "refresh_token"))
+        assertThat(body, Matchers.not(Matchers.hasKey("scope")))
```
lbalmaceda, line 1512: this ensures the renew auth method is not setting a scope value by default

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -1505,6 +1509,30 @@ public class AuthenticationAPIClientTest {
...
+        assertThat(credentials, Matchers.`is`(Matchers.notNullValue()))
+    }
+
+    @Test
+    public fun shouldRenewAuthWithOAuthTokenAndCustomScope() {
```
lbalmaceda, line 1517: this checks that it can still be set a scope by the dev later, if desired

jimmyjames (approved):

PR #460, opened 2021-02-17
Add SLA to v1 README

lbalmaceda:
Adds the SLA we're committing to for supporting version v1 of this SDK. This header will be visible when the v1 branch is visited.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,9 @@ Android Java toolkit for Auth0 API
 
 ---
 
-**Version 2.0.0 of this SDK is already available!**
+**Version 2.0.0 of this SDK is available!**
 
-Check out the [Quickstart article](https://auth0.com/docs/quickstart/native/android) or the [source code](https://github.com/auth0/Auth0.Android). Version v1 will continue to receive *Bug & Security fixes* until 11th May 2021. After that date, the only version supported will become v2.
+Check out the [Quickstart article](https://auth0.com/docs/quickstart/native/android) or the [source code](https://github.com/auth0/Auth0.Android). Version 1 will continue to receive Bug & Security fixes until 11th May 2021. After that date, only version 2 will be supported.
 
 ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,14 @@
 
 Android Java toolkit for Auth0 API
 
+---
+
+**Version 2.0.0 of this SDK is already available!**
```
jimmyjames, line 13:
```suggestion
**Version 2.0.0 of this SDK is available!**
```

README.md
```diff
@@ -8,6 +8,14 @@
...
+
+Check out the [Quickstart article](https://auth0.com/docs/quickstart/native/android) or the [source code](https://github.com/auth0/Auth0.Android). Version v1 will continue to receive *Bug & Security fixes* until 11th May 2021. After that date, the only version supported will become v2.
```
jimmyjames, line 15: What's the purpose of linking back to the repository from the README? It may make more sense to link to the samples?

Also, consider:
"Version 1 will continue to receive *Bug & Security fixes* until 11th May 2021. After that date, only version 2 will be supported."

jimmyjames (changes requested):
Some suggestions/questions. Also please add a description to the PR.

README.md
lbalmaceda: Good point. I was pointing to the `main` branch (the default actually) since this PR is for updating the README on the `v1` branch. That way, those looking for v1 docs would be welcomed with this message inviting them to check the newer version instead of on the branch `main`. Is that something you rather I remove? 

The sample app is linked from the quickstart article. I'm not sure linking to the quickstart sample app from here would add much value, as the how-to is in the article, and the sample without those steps is too raw.

I'll take the suggested text and also apply the other suggestion you added separately locally 👍

jimmyjames: ah, I understand now. Makes sense.

jimmyjames (approved):

PR #475, opened 2021-04-16
Accept UserMetadata for creating users [SDK-2429]

lbalmaceda:
### Changes

This SDK no longer accepts passing object values for the request parameters. There's the exception of the `user_metadata`, which was already supported via the `UsersAPIClient#updateMetadata` method.

This PR expands that implementation into the `AuthenticationAPIClient`, so `createUser` and `signUp` methods can now take an extra parameter for when the user metadata needs to be specified.

The map requires "String to String" entries, as seen in the server error message below.
image

### References
See `SDK-2429`

### Testing

Added unit tests and also changed one test helper method to be able to parse requests that have values of type different than string.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
@@ -4,12 +4,9 @@ import androidx.annotation.VisibleForTesting
 import com.auth0.android.Auth0
 import com.auth0.android.Auth0Exception
 import com.auth0.android.request.*
-import com.auth0.android.request.internal.BaseAuthenticationRequest
-import com.auth0.android.request.internal.GsonAdapter
+import com.auth0.android.request.internal.*
 import com.auth0.android.request.internal.GsonAdapter.Companion.forMap
 import com.auth0.android.request.internal.GsonAdapter.Companion.forMapOf
-import com.auth0.android.request.internal.GsonProvider
-import com.auth0.android.request.internal.RequestFactory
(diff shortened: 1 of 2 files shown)


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -285,14 +282,16 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
...
      * @return a request to start
      */
     @JvmOverloads
     public fun createUser(
         email: String,
         password: String,
         username: String? = null,
-        connection: String
+        connection: String,
+        userMetadata: Map<String, String>? = null
```
lbalmaceda, line 294: optional parameter with a default value of `null`. The `JvmOverloads` annotation covers the java use case by generating one extra method with the added parameter

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -1624,8 +1712,8 @@ public class AuthenticationAPIClientTest {
         )
     }
 
-    private fun bodyFromRequest(request: RecordedRequest): Map<String, String> {
-        val mapType = object : TypeToken<Map<String?, String?>?>() {}.type
+    private fun <T> bodyFromRequest(request: RecordedRequest): Map<String, T> {
```
lbalmaceda, line 1715: this change required me to specify the type in all of the tests that were using it above

jimmyjames (approved):

Issue #463, opened 2021-02-24
Unit tests involving passwordlessWithEmail fail with ExceptionInInitializerError on upgrade to version 2.0.0


alvindizon:
### Describe the problem

Unit tests involving `AuthenticationAPIClient::passwordlessWithEmail` break on upgrade to version 2.0.0

The full stacktrace is:

```
java.lang.ExceptionInInitializerError
	at com.auth0.android.request.internal.RequestFactory.setupRequest(RequestFactory.kt:82)
	at com.auth0.android.request.internal.RequestFactory.post(RequestFactory.kt:32)
	at com.auth0.android.request.internal.RequestFactory.post(RequestFactory.kt:35)
	at com.auth0.android.authentication.AuthenticationAPIClient.passwordless(AuthenticationAPIClient.kt:531)
	at com.auth0.android.authentication.AuthenticationAPIClient.passwordlessWithEmail(AuthenticationAPIClient.kt:480)
	at com.auth0.android.authentication.AuthenticationAPIClient.passwordlessWithEmail$default(AuthenticationAPIClient.kt:473)
	at com.someapp.login.auth0.AuthManagerImpl$sendEmailCode$1.subscribe(AuthManager.kt:90)
	at io.reactivex.rxjava3.internal.operators.single.SingleCreate.subscribeActual(SingleCreate.java:40)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.core.Single.test(Single.java:5513)
	at com.someapp.login.auth0.AuthManagerTest.sendEmailCode returns EmailSent when response is successful(AuthManagerTest.kt:104)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
[...]

PR #482, opened 2021-06-03
Rewrite ThreadSwitcher class so that it is not tied to Looper

alvindizon:
### Changes

This PR addresses the feature request made in #463, which is a request to allow users to set their own implementation of `ThreadSwitcher` to be used in unit tests. This is useful for unit testing, so that `java.lang.RuntimeException:
Method getMainLooper in android.os.Looper not mocked.` errors can be avoided without having to use Robolectric.

### References

Ticket for this can be found here: #463 

### Testing

I've tested this PR by generating an AAR and adding it to our project, checking that our unit tests pass using the JUnit 5 `Extension` and seeing that login/logout works normally in our app.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -821,6 +821,71 @@ Re-declare the activity manually with `tools:node="remove"` in your app's Androi
     tools:node="remove"/>
 ```
 
+### Unit testing with JUnit 4 or JUnit 5
+#### Handling `Method getMainLooper in android.os.Looper not mocked` errors
+Your unit tests might break with `Caused by: java.lang.RuntimeException: Method getMainLooper in android.os.Looper not mocked` due to the Looper being used internally by this library. There are two options to handle this:
+1. Use Robolectric Shadows - see this [test](https://github.com/auth0/Auth0.Android/blob/main/auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt#L44-L45) for an example
+2. If your project does not use Robolectric and uses JUnit 4, you can create a `Rule` that you can add to your unit test:
+```kotlin
+import com.auth0.android.request.internal.CommonThreadSwitcher
+import com.auth0.android.request.internal.ThreadSwitcher
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+public class CommonThreadSwitcherRule : TestWatcher() {
+    override fun starting(description: Description) {
+        super.starting(description)
+        CommonThreadSwitcher.getInstance().setDelegate(object : ThreadSwitcher {
+            override fun mainThread(runnable: Runnable) {
+                runnable.run()
+            }
+
+            override fun backgroundThread(runnable: Runnable) {
+                runnable.run()
+            }
+        })
+    }
+
+    override fun finished(description: Description) {
+        super.finished(description)
+        CommonThreadSwitcher.getInstance().setDelegate(null)
+    }
+}
+```
+See this [test](https://github.com/auth0/Auth0.Android/blob/main/auth0/src/test/java/com/auth0/android/request/internal/CommonThreadSwitcherDelegateTest.kt) for an example of it being used.
+
+3. If you use JUnit 5 then you can create an `Extension` similar to the previous `Rule` for JUnit 4:
+```kotlin
+import com.auth0.android.request.internal.CommonThreadSwitcher
+import com.auth0.android.request.internal.ThreadSwitcher
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+
+class CommonThreadSwitcherExtension : BeforeEachCallback, AfterEachCallback {
+
+    override fun beforeEach(context: ExtensionContext?) {
+        CommonThreadSwitcher.getInstance().setDelegate(object : ThreadSwitcher {
+            override fun mainThread(runnable: Runnable) {
+                runnable.run()
+            }
+
+            override fun backgroundThread(runnable: Runnable) {
+                runnable.run()
+            }
+        })
+    }
+
+    override fun afterEach(context: ExtensionContext?) {
+        CommonThreadSwitcher.getInstance().setDelegate(null)
+    }
+
+}
+```
+
+#### Handling SSL errors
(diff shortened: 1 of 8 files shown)


auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -18,31 +18,23 @@ import javax.net.ssl.X509TrustManager
...
+ * @param trustManager manages which X509 certificates will be used for authentication.  Can be set to a custom X509TrustManager for unit testing
  */
-public class DefaultClient @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE) internal constructor(
-    connectTimeout: Int,
-    readTimeout: Int,
-    private val defaultHeaders: Map<String, String>,
-    enableLogging: Boolean,
-    sslSocketFactory: SSLSocketFactory?,
-    trustManager: X509TrustManager?
+public class DefaultClient constructor(
```
lbalmaceda, line 29: We really _don't want_ to expose this constructor that takes the socket factory and trust manager parameters. These were previously exposed internally (for the module) so we could reference and mock them from the tests. Can we roll this visibility back?

auth0/src/main/java/com/auth0/android/request/internal/ThreadSwitcher.kt
```diff
@@ -2,53 +2,125 @@ package com.auth0.android.request.internal
...
-    internal companion object {
-        /**
-         * The maximum concurrent threads to execute
-         * in the background. Value taken from the Android docs.
-         * @see <a href="https://developer.android.com/guide/background/threading#creating-multiple-threads">Android: creating-multiple-threads</a>
-         */
-        internal const val MAX_CONCURRENT_THREADS = 4
+    private fun createAsync(looper: Looper): Handler {
+        try {
+            return Handler::class.java.getDeclaredConstructor(
```
lbalmaceda, line 45: 😬 using java reflection is a big red flag to us. Is there any other way of achieving this?

lbalmaceda (changes requested):
Thanks for the contribution. I performed a quick scan to provide initial feedback. Left you some comments.

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
alvindizon: Thanks for the feedback. We need this exposed for our unit tests to succeed as well as I've noted in the updated Kdoc I made for DefaultClient. Supplying a test ssl socket factory and trustManager is a must as HTTPs is now used by Auth0, or else `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target` will occur during unit tests. I see no reason for this constructor to be not exposed to developers, do you have any alternatives?

auth0/src/main/java/com/auth0/android/request/internal/ThreadSwitcher.kt
alvindizon: This is an oversight on my part, I'll try to redo this. Thanks

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
alvindizon: I've done more thinking on this and maybe we'll just provide our own internal implementation of NetworkingClient instead of exposing SSLSocketFactory and TrustManager in DefaultClient's constructor. Will update this PR if all goes well with our unit tests.

README.md
```diff
@@ -827,6 +827,70 @@ By default you should at least use the following files:
 * `proguard-okio.pro`
 * `proguard-gson.pro`
 
+## Unit testing with JUnit 4 or JUnit 5
```
lbalmaceda, line 830: I suggest we use a level 3 header and nest it under "FAQ" above.

README.md
```diff
@@ -827,6 +827,70 @@ By default you should at least use the following files:
...
+
+    override fun afterEach(context: ExtensionContext?) {
+        CommonThreadSwitcher.getInstance().setDelegate(null)
+    }
+
+}
+```
+
+#### Handling SSL errors
+You might encounter errors similar to `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`, which means that you need to setup your unit tests in a way  that ignores or trusts all SSL certificates. In that case, you may have to implement your own `NetworkingClient` so that you can supply your own `SSLSocketFactory` and `X509TrustManager`, and use that in creating your `Auth0` object. See the [`DefaultClient`](https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt) class for an idea on how to extend `NetworkingClient`.
```
lbalmaceda, line 893:
```suggestion
You might encounter errors similar to `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`, which means that you need to set up your unit tests in a way that ignores or trusts all SSL certificates. In that case, you may have to implement your own `NetworkingClient` so that you can supply your own `SSLSocketFactory` and `X509TrustManager`, and use that in creating your `Auth0` object. See the [`DefaultClient`](https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt) class for an idea on how to extend `NetworkingClient`.
```

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
```diff
@@ -275,30 +275,32 @@ public class BaseRequestTest {
     public fun shouldExecuteRequestOnBackgroundThreadAndPostSuccessToMainThread() {
         val pausedExecutorService = PausedExecutorService()
         val threadSwitcher =
```
lbalmaceda, line 277: nit: we could avoid most of the diff in this file if we rename `threadSwitcher` to `defaultThreadSwitcher` and later call the spy `threadSwitcher`.

lbalmaceda (changes requested):
Thanks for the effort you put into this PR ✨ . I left a few minor comments, please re-request the review when ready.

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
alvindizon: That's a good call out. Will update accordingly

README.md
alvindizon: I just updated the PR with your suggestion 👍

README.md
alvindizon: I just updated the PR with your suggestion 👍

lbalmaceda (approved):
Thanks for the changes!

PR #491, opened 2021-06-22
Explicitly specify charset, don't rely on default charset

### LINKED ISSUE #484: `BaseRequest` uses `Charset.defaultCharset()`
### Describe the problem
`BaseRequest` uses [`Charset.defaultCharset()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/charset/Charset.html#defaultCharset()), which is the operating system charset:
https://github.com/auth0/Auth0.Android/blob/5f20b76c8bca3ec5a5fba135e6ce4ae183e2ec5b/auth0/src/main/java/com/auth0/android/request/internal/BaseRequest.kt#L96

Is there a reason for this? Normally JSON is UTF-8 encoded; using the operating system charset will most likely cause issues if the charset does not support all Unicode code points or uses an encoding which is not even compatible with ASCII.
It looks like #390 introduced the change [here](https://github.com/auth0/Auth0.Android/pull/390/files#diff-c61404ff99127bb379a5afa779eab8b6597a495d3165af1fea98a6bc5501efdaR118) but I did not find any explanation.

Ideally one of OkHttp `ResponseBody` methods would be used, such as [`charStream()`](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/char-stream/), since they also consider the HTTP headers, maybe `ServerResponse` should have a `Reader` instead of an `InputStream` as `body` then.

Also note that exception handling of `BaseRequest.execute()` should probably be improved by using [`Closeable.use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) so in all cases the `reader` is closed.

### LINKED ISSUE #485: `AsymmetricSignatureVerifier` should not use `Charset.defaultCharset()`
### Describe the problem
`AsymmetricSignatureVerifier` uses the default charset during token validation:
https://github.com/auth0/Auth0.Android/blob/5f20b76c8bca3ec5a5fba135e6ce4ae183e2ec5b/auth0/src/main/java/com/auth0/android/provider/AsymmetricSignatureVerifier.java#L42

It should probably instead specify an explicit charset, most likely either [`StandardCharsets.ISO_8859_1`](https://developer.android.com/reference/java/nio/charset/StandardCharsets#ISO_8859_1) or [`StandardCharsets.US_ASCII`](https://developer.android.com/reference/java/nio/charset/StandardCharsets#US_ASCII).

Marcono1234:
### Changes
Fixes #484, fixes #485

Replaces (implicit) usage of the default charset with explicit usage of [`StandardCharsets.UTF_8`](https://developer.android.com/reference/java/nio/charset/StandardCharsets#UTF_8) (added in API level 19).

Android guarantees that [`Charset.defaultCharset()` is UTF-8](https://developer.android.com/reference/java/nio/charset/Charset#defaultCharset()) (note that this is not true for the JDK, though that might not be important here), however to make the expected behavior clearer this pull request explicitly uses UTF-8 as charset.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
@@ -16,6 +16,7 @@ import androidx.annotation.VisibleForTesting;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.KeyPairGenerator;
@@ -367,7 +368,7 @@ class CryptoUtil {
             byte[] aes = keyGen.generateKey().getEncoded();
             //Save encrypted encoded version
             byte[] encryptedAES = RSAEncrypt(aes);
-            String encodedEncryptedAESText = new String(Base64.encode(encryptedAES, Base64.DEFAULT));
+            String encodedEncryptedAESText = new String(Base64.encode(encryptedAES, Base64.DEFAULT), StandardCharsets.UTF_8);
             storage.store(KEY_ALIAS, encodedEncryptedAESText);
             return aes;
         } catch (NoSuchAlgorithmException e) {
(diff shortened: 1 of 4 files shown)


lbalmaceda (approved):

lbalmaceda:
Thanks for your contribution!

PR #492, opened 2021-06-22
Always close request body InputStream when exception occurs

Marcono1234:
### Changes
Resolves concerns raised in https://github.com/auth0/Auth0.Android/issues/484#issuecomment-864298766
Depends on #486

Uses Kotlin's [`Closeable.use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) extension function to make sure the request body is always closed by `BaseRequest.execute()`, even when exceptions occur while reading.
Additionally removes `AwareInputStreamReader` since that was only needed for unit tests. Instead the tests now use custom `InputStream` implementations which track whether they were closed, and some more tests have been added to verify that for network errors the stream is closed as well (these tests are currently failing, partially related to #486).

(Note that the test code might not be as clean as it was before, let me know if you want it to be changed.)

### Testing

### Checklist


lbalmaceda:
@Marcono1234 Thanks for the work here. I'm waiting on the PR you linked as a dependency to have its comments resolved by @jeffdgr8. Once merged I'll leave you a review.

lbalmaceda:
@Marcono1234 I just merged the PR you based this on. Please, rebase and when ready, take it out of draft so I can review it. Thanks!

Marcono1234:
I have rebased this pull request now. The changes in `BaseRequest.kt` are mostly indentation changes (GitHub does not display that very well).
Please let me know if the modifications to `BaseRequestTest.kt` are intrusive, or if you have any suggestions.

When I build the project locally (`gradlew build`) there are a few test failures in unrelated classes:
- `InvalidKeyException: invalid key format`
Test classes:
  - `AsymmetricSignatureVerifierTest`
  - `IdTokenVerifierTest`
- `IllegalArgumentException: bad base-64`
Test classes:
  - `WebAuthProviderTest`

I have also added a commit for fixing test failures on Windows where the `-----BEGIN PRIVATE KEY-----` header was not removed from PEM certificates.

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
```diff
@@ -187,33 +200,28 @@ public class BaseRequestTest {
             resultAdapter,
             errorAdapter
         )
-        val networkError = mock<JsonIOException>()
+        val networkError = JsonIOException("Network error")
```
lbalmaceda, line 203: why is a real instance prefered vs a mock?

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
```diff
@@ -400,56 +450,143 @@ public class BaseRequestTest {
         )
     }
 
+    /**
+     * @return Function indicating whether the response body `InputStream` was closed
+     */
     @Throws(Exception::class)
-    private fun mockSuccessfulServerResponse() {
+    private fun mockSuccessfulServerResponse(): () -> Boolean {
```
lbalmaceda, line 457: I'm not 100% convinced about the signature changes. Now, these functions return a function that would yield a boolean, but the name of each of the 3 functions is not implying that. Do you think it could be refactored? Maybe make it return something else (i.e. the input stream) and wrap that with a separate function to check if it was closed? 

What are your thoughts?

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
Marcono1234: I think here for such a simple class using a mock might be excessive (but that is probably personal preference).
The advantage of using a real class here is that you can use a custom exception message which might make debugging test failures easier.

auth0/src/test/java/com/auth0/android/request/internal/BaseRequestTest.kt
Marcono1234: You are right, the function now becomes responsible for two tasks which is probably not a good design. But I am not sure if there is an easier solution which does not make the tests much more verbose.
Returning the `AwareInputStream` might have the same problem because the functions are not really producing the input stream as mocked response, but instead are modifying the mocks, so returning the input stream might be equally irritating.

Maybe an alternative would be to add a mutable `wasResponseBodyClosed: Boolean` property to the test class? What do you think? The test functions not using the mocking functions could just ignore that property.

lbalmaceda: @Marcono1234 Yes, I think that would be better. Not that I don't like the kotlin way shown in the current solution, but the double-task part is not convincing me. As long as we reset the value to "false" before each test, adding a property should be ok.

Alright, those should be the last set of changes. Thanks for sticking to this PR!

Marcono1234:
Sorry for the delay. I have rebased the changes onto `main` and added a commit addressing your review comment. It changes the mock methods to again not have a return value, and instead introduces the property `wasResponseStreamClosed` for the test class whose value is checked.

Note that commit 013dd29 might be obsolete with your addition of `.gitattributes` in 92e0f16c2c218031af85cc7b2c27b5b96386527f, but I think `.gitattributes` does not affect developers who have already checked out the files of this project locally, so at least for them that fix might still be necessary.

lbalmaceda (dismissed):

lbalmaceda:
Apologies for the delay. I had to close and re-open it to trigger the CI that was stuck. I like the changes so I'll be adding these to the next release. Thanks again for the contribution @Marcono1234.

lbalmaceda (approved):

Marcono1234:
No problem and thanks for accepting this contribution!

PR #500, opened 2021-07-16
Add support for Recovery Code multi-factor authentication

lbalmaceda:
### Changes

This endpoint is used to authenticate when MFA is required and the user has a saved recovery code. A new recovery code might be sent back by the server, which needs to be displayed to the end-user.

### References

- See https://auth0.com/docs/api/authentication#verify-with-recovery-code
- Also normalized the line endings following: https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

### Testing

### Checklist


auth0/src/main/java/com/auth0/android/result/Credentials.kt
```diff
@@ -62,4 +62,13 @@ public open class Credentials(
...
     internal open val currentTimeInMillis: Long
         get() = System.currentTimeMillis()
+
+    /**
+     * Getter for the new multi-factor authentication recovery code. Only available if these credentials are the result of logging in using an MFA recovery code.
+     *
+     * @return the new MFA recovery code.
+     */
+    @field:SerializedName("recovery_code")
+    public var recoveryCode: String? = null
```
lbalmaceda, line 72: having a setter separate from the constructor prevents introducing a breaking change, as verified with the tests

auth0/src/main/java/com/auth0/android/request/internal/CredentialsDeserializer.kt
```diff
@@ -0,0 +1,68 @@
...
+        val idToken = context.deserialize<String>(jsonObject.remove("id_token"), String::class.java)
+        val accessToken =
+            context.deserialize<String>(jsonObject.remove("access_token"), String::class.java)
+        val type = context.deserialize<String>(jsonObject.remove("token_type"), String::class.java)
+        val refreshToken =
+            context.deserialize<String>(jsonObject.remove("refresh_token"), String::class.java)
+        val expiresIn = context.deserialize<Long>(jsonObject.remove("expires_in"), Long::class.java)
+        val scope = context.deserialize<String>(jsonObject.remove("scope"), String::class.java)
+        val recoveryCode =
+            context.deserialize<String>(jsonObject.remove("recovery_code"), String::class.java)
```
lbalmaceda, line 32: lines added here, reading the `recovery_code` property. The rest is just java->kotlin conversion from the existing class

auth0/src/test/java/com/auth0/android/request/internal/CredentialsDeserializerMock.kt
```diff
@@ -0,0 +1,25 @@
...
+        accessToken: String,
+        type: String,
+        refreshToken: String?,
+        expiresAt: Date,
+        scope: String?,
+        recoveryCode: String?
+    ): Credentials {
+        val credentials =
+            CredentialsMock(idToken, accessToken, type, refreshToken, expiresAt, scope)
+        credentials.recoveryCode = recoveryCode
```
lbalmaceda, line 19: java->kotlin converstion to be able to call this internal setter

auth0/src/test/java/com/auth0/android/result/CredentialsTest.kt
```diff
@@ -0,0 +1,38 @@
...
+            Credentials("idToken", "accessToken", "type", "refreshToken", Date(), "openid profile")
+        MatcherAssert.assertThat(credentials.scope, Matchers.`is`("openid profile"))
+    }
+
+    @Test
+    public fun shouldGetRecoveryCode() {
+        val date = Date()
+        val credentials: Credentials =
+            CredentialsMock("idToken", "accessToken", "type", "refreshToken", date, "scope")
+        credentials.recoveryCode = "recoveryCode"
```
lbalmaceda, line 35: java->kotlin converstion to be able to call this internal setter

auth0/src/main/java/com/auth0/android/result/Credentials.kt
```diff
@@ -62,4 +62,13 @@ public open class Credentials(
...
+        internal set
 }
```
Widcket, line 74:
```suggestion
}

```

auth0/src/main/java/com/auth0/android/request/internal/CredentialsDeserializer.kt
```diff
@@ -0,0 +1,68 @@
...
+        scope: String?,
+        recoveryCode: String?
+    ): Credentials {
+        val credentials = Credentials(
+            idToken, accessToken, type, refreshToken, expiresAt, scope
+        )
+        credentials.recoveryCode = recoveryCode
+        return credentials
+    }
+}
```
Widcket, line 68:
```suggestion
}

```

auth0/src/main/java/com/auth0/android/result/Credentials.kt
Widcket, line 74:
```suggestion
}

```

auth0/src/test/java/com/auth0/android/result/CredentialsTest.kt
```diff
@@ -0,0 +1,38 @@
...
+        MatcherAssert.assertThat(credentials.recoveryCode, Matchers.`is`("recoveryCode"))
+    }
+}
```
Widcket, line 38:
```suggestion
}

```

auth0/src/main/java/com/auth0/android/request/internal/CredentialsDeserializer.kt
lbalmaceda: These files have been applied the default code formatting shortcut to match the same style we follow in the rest of the files.

Widcket: It's more about following the UNIX convention of ending files with a newline.

Widcket: (text files)

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -172,6 +172,39 @@ public class AuthenticationAPIClientTest {
...
+        val client = AuthenticationAPIClient(auth0)
+        client.loginWithRecoveryCode("ey30.the-mfa-token.value", "123456")
+            .start(callback)
+        ShadowLooper.idleMainLooper()
+        assertThat(
+            callback, AuthenticationCallbackMatcher.hasPayloadOfType(
+                Credentials::class.java
+            )
+        )
+        assertThat(callback.payload.recoveryCode, Matchers.`is`("654321"))
```
Widcket, line 189: If we're asserting the recovery code on the payload, should we be asserting the token value as well?

Widcket: Unless that is asserted in some other test, but the same could be said about the `Accept-Language` assert.

lbalmaceda: @Widcket This line checks for the content of the `response`. Here we only care about checking that if a new recovery code was sent, we can parse it. (see one is 123456 and the other is 654321).

A few lines below we assert the content of the `request`, which includes both mfa token and recovery code.

auth0/src/test/java/com/auth0/android/result/CredentialsTest.kt
```diff
@@ -0,0 +1,38 @@
...
+        val credentials =
+            Credentials("idToken", "accessToken", "type", "refreshToken", Date(), "openid profile")
+        MatcherAssert.assertThat(credentials.scope, Matchers.`is`("openid profile"))
+    }
+
+    @Test
+    public fun shouldGetRecoveryCode() {
+        val date = Date()
+        val credentials: Credentials =
+            CredentialsMock("idToken", "accessToken", "type", "refreshToken", date, "scope")
```
Widcket, line 34: Shouldn't `recoveryCode` be passed to the constructor?

Widcket (approved):
LGTM, only have a question left that is not a blocker.

auth0/src/test/java/com/auth0/android/result/CredentialsTest.kt
lbalmaceda: There is no new constructor that takes a recovery code. I didn't want to create a secondary constructor just for this one property, and changing the existing one would be a breaking change even when default values are used. Kotlin doesn't enforce the use of "named arguments" so anyone could be calling it the way we are calling it in this specific line and would be facing a breaking change due to a missing property.

Issue #511, opened 2021-08-23
Add parameters map to getCredentials


lustikuss:
Hello,
currently it is possible to add a map of parameters to the `login` and the `renewAuth` call. Unfortunately, this is not possible for the `getCredentials` function.

### Describe the ideal solution
To have a an uniform signature between the `AuthenticationApiClient` and the `CredentialsManager` add the parameter `parameters: Map<String, Any?>` to `BaseCredentialsManager.getCredentials`.

I would like to provide a pull request where I add the parameter to the aforementioned function.

## Alternatives and current work-arounds
None that I know of.

## PS

Addtionally, this would lead to feature equality with auth0.swift: https://github.com/auth0/Auth0.swift/issues/493

lbalmaceda:
@lustikuss Please expand a bit on your use case, so we can review the feature request. Thanks

lustikuss:
In our use case, the app is required to renew tokens on special events (user changes an internal state) and inform the authentification provider about changes the user made (via parameters). Additionally this set parameter is required to be send with every authentication call.

I.E. `fun getCredentials(scope: String?, minTtl: Int, callback: Callback<Credentials, CredentialsManagerException>)` gets called and the token hasExpired or willExpire or the scope changed.  
Therefore `val request = authenticationClient.renewAuth(refreshToken)` will be called. But without the possibility to add parameters.

So additionally to `if (scope != null) { request.addParameter("scope", scope) }` we would like to have something like `        parameters.forEach {            request.addParameter(it.key, it.value)        }` where parameters is `Map<String, String>`.

lustikuss:
Hi, I opened a pull request with the discussed changes. Can I assist with something else?

lbalmaceda:
I'm sorry for the delay in getting back to you but I still couldn't take a look at your PR. I'll try to review it tomorrow. Cheers

PR #514, opened 2021-09-15
Credentials Managers: renew tokens with extra parameters

lustikuss:
### Changes

This pull request adds the possibility to add custom parameters to the web request in getCredentials.
Similary to renewAuth and login it is possible to add custom parameters to the web request

- a Map<String, String> parameter called parameters has been added to com.auth0.android.authentication.storage.BaseCredentialsManager#getCredentials()

### References

- Android Issue: https://github.com/auth0/Auth0.Android/issues/511
- IOS Issue: https://github.com/auth0/Auth0.swift/issues/493

### Testing
Changes in this PR have been tested in a real-world Product.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,7 +585,7 @@ val manager = CredentialsManager(authentication, storage)
 ```
 
 2. **Save credentials:**
-The credentials to save **must have** `expires_in` and at least an `access_token` or `id_token` value. If one of the values is missing when trying to set the credentials, the method will throw a `CredentialsManagerException`. If you want the manager to successfully renew the credentials when expired you must also request the `offline_access` scope when logging in in order to receive a `refresh_token` value along with the rest of the tokens. i.e. Logging in with a database connection and saving the credentials:
+The credentials to save **must have** `expires_at` and at least an `access_token` or `id_token` value. If one of the values is missing when trying to set the credentials, the method will throw a `CredentialsManagerException`. If you want the manager to successfully renew the credentials when expired you must also request the `offline_access` scope when logging in in order to receive a `refresh_token` value along with the rest of the tokens. i.e. Logging in with a database connection and saving the credentials:
 
 ```kotlin
 authentication
@@ -613,7 +613,7 @@ val authenticated = manager.hasValidCredentials()
 ```
 
 4. **Retrieve credentials:**
-Existing credentials will be returned if they are still valid, otherwise the `refresh_token` will be used to attempt to renew them. If the `expires_in` or both the `access_token` and `id_token` values are missing, the method will throw a `CredentialsManagerException`. The same will happen if the credentials have expired and there's no `refresh_token` available.
+Existing credentials will be returned if they are still valid, otherwise the `refresh_token` will be used to attempt to renew them. If the `expires_at` or both the `access_token` and `id_token` values are missing, the method will throw a `CredentialsManagerException`. The same will happen if the credentials have expired and there's no `refresh_token` available.
 
 ```kotlin
 manager.getCredentials(object : Callback<Credentials, CredentialsManagerException> {
--- a/auth0/src/main/java/com/auth0/android/Auth0.kt
+++ b/auth0/src/main/java/com/auth0/android/Auth0.kt
@@ -1,12 +1,10 @@
 package com.auth0.android
 
 import android.content.Context
-import androidx.annotation.VisibleForTesting
 import com.auth0.android.request.DefaultClient
 import com.auth0.android.request.NetworkingClient
 import com.auth0.android.util.Auth0UserAgent
 import okhttp3.HttpUrl
-import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import java.util.*
 
@@ -93,30 +91,11 @@ public open class Auth0 @JvmOverloads constructor(
             .build()
             .toString()
 
-
-    private fun resolveConfiguration(configurationDomain: String?, domainUrl: HttpUrl): HttpUrl {
-        var url = ensureValidUrl(configurationDomain)
-        if (url == null) {
-            val host = domainUrl.host
-            url = if (host.endsWith(DOT_AUTH0_DOT_COM)) {
-                val parts = host.split(".").toTypedArray()
-                if (parts.size > 3) {
-                    ("https://cdn." + parts[parts.size - 3] + DOT_AUTH0_DOT_COM).toHttpUrl()
-                } else {
-                    AUTH0_US_CDN_URL.toHttpUrl()
-                }
-            } else {
-                domainUrl
-            }
-        }
-        return url
-    }
-
     private fun ensureValidUrl(url: String?): HttpUrl? {
         if (url == null) {
             return null
         }
-        val normalizedUrl = url.toLowerCase(Locale.ROOT)
+        val normalizedUrl = url.lowercase(Locale.ROOT)
         require(!normalizedUrl.startsWith("http://")) { "Invalid domain url: '$url'. Only HTTPS domain URLs are supported. If no scheme is passed, HTTPS will be used." }
         val safeUrl =
             if (normalizedUrl.startsWith("https://")) normalizedUrl else "https://$normalizedUrl"
@@ -124,8 +103,6 @@ public open class Auth0 @JvmOverloads constructor(
     }
 
     private companion object {
-        private const val AUTH0_US_CDN_URL = "https://cdn.auth0.com"
-        private const val DOT_AUTH0_DOT_COM = ".auth0.com"
         private fun getResourceFromContext(context: Context, resName: String): String {
             val stringRes = context.resources.getIdentifier(resName, "string", context.packageName)
             require(stringRes != 0) {
@@ -141,7 +118,7 @@ public open class Auth0 @JvmOverloads constructor(
     init {
         domainUrl = ensureValidUrl(domain)
         requireNotNull(domainUrl) { String.format("Invalid domain url: '%s'", domain) }
-        configurationUrl = resolveConfiguration(configurationDomain, domainUrl)
+        configurationUrl = ensureValidUrl(configurationDomain) ?: domainUrl
         auth0UserAgent = Auth0UserAgent()
     }
 }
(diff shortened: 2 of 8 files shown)


auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -113,6 +115,10 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         if (scope != null) {
             request.addParameter("scope", scope)
         }
+        parameters.forEach { (key, value) ->
```
lbalmaceda, line 118: all can be added at once. See https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/request/Request.kt#L35

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -340,6 +343,10 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
         if (scope != null) {
             request.addParameter("scope", scope)
         }
+        parameters.forEach { (key, value) ->
```
lbalmaceda, line 346: all can be added at once. See https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/request/Request.kt#L35

lbalmaceda (changes requested):
@lustikuss Thanks for the contribution. 

The reason you had to fix the existing tests is that the change you introduced is a breaking change. This library still supports java users, so just relying on Kotlin's default parameters is not enough, unfortunately. That's why you have to pass either named parameters or all of the parameters (4 now vs 3 before).

I think this feature can still be achieved using the `@JvmOverloads` annotation in the abstract class. It will create a new signature for the method combining each parameter that has a default value set. That should keep the old signatures in place while adding the new one you are introducing. You can read more about it here: https://kotlinlang.org/docs/java-to-kotlin-interop.html#overloads-generation. 

Once that's fixed I can give it another pass and focus on the test. Cheers

lustikuss:
Hi, thanks for the feedback! I totally forgot about the backwards compatibility, sorry! Unfortunetaly @JvmOverloads is not possible for abstract functions, so I just re-created the old getCredentials function which calls the new getCredentials fun with an empty map.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -202,6 +202,30 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
...
+     * If a LockScreen is setup and [.requireAuthentication] was called, the user will be asked to authenticate before accessing
+     * the credentials. Your activity must override the [Activity.onActivityResult] method and call
+     * [.checkAuthenticationResult] with the received values.
+     *
+     * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
+     * @param minTtl   the minimum time in seconds that the access token should last before expiration.
+     * @param parameters to send with the request as a map of string with the keys as string
+     * @param callback the callback to receive the result in.
+     */
+    override fun getCredentials(
```
lbalmaceda, line 224: remove override

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -72,6 +72,25 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
...
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with [CredentialsManagerException] if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
+     * @param minTtl   the minimum time in seconds that the access token should last before expiration.
+     * @param parameters to send with the request as a map of string with the keys as string
+     * @param callback the callback that will receive a valid [Credentials] or the [CredentialsManagerException].
+     */
+    override fun getCredentials(
```
lbalmaceda, line 89: remove override

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
```diff
@@ -914,6 +914,34 @@ public class CredentialsManagerTest {
...
+            client.renewAuth("refreshToken")
+        ).thenReturn(request)
+
+        manager.getCredentials(
+            scope = "some changed scope to trigger refresh",
+            minTtl = 0,
+            parameters = parameters,
+            callback = callback
+        )
+        verify(request).start(requestCallbackCaptor.capture())
```
lbalmaceda, line 942: you should as well verify that the request was created with the additional parameters. See for reference what you did in the other test https://github.com/auth0/Auth0.Android/pull/514/files#diff-bd364d0aa7aaac49b5c2c42fc4de0c065f7b9df5c6f8aa8fc50ae665cd057d92R1411

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -202,6 +202,30 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
...
+     * at all with the cryptographic implementation and will have [CredentialsManagerException.isDeviceIncompatible] return true.
+     *
+     *
+     * If a LockScreen is setup and [.requireAuthentication] was called, the user will be asked to authenticate before accessing
+     * the credentials. Your activity must override the [Activity.onActivityResult] method and call
+     * [.checkAuthenticationResult] with the received values.
+     *
+     * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
+     * @param minTtl   the minimum time in seconds that the access token should last before expiration.
+     * @param parameters to send with the request as a map of string with the keys as string
```
lbalmaceda, line 221:
```suggestion
     * @param parameters additional parameters to send in the request to refresh expired credentials
```

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -72,6 +72,25 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
...
+    }
+
+    /**
+     * Retrieves the credentials from the storage and refresh them if they have already expired.
+     * It will fail with [CredentialsManagerException] if the saved access_token or id_token is null,
+     * or if the tokens have already expired and the refresh_token is null.
+     *
+     * @param scope    the scope to request for the access token. If null is passed, the previous scope will be kept.
+     * @param minTtl   the minimum time in seconds that the access token should last before expiration.
+     * @param parameters to send with the request as a map of string with the keys as string
```
lbalmaceda, line 86:
```suggestion
     * @param parameters additional parameters to send in the request to refresh expired credentials
```

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -340,6 +365,9 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
         if (scope != null) {
             request.addParameter("scope", scope)
         }
+
+        request.addParameters(parameters)
```
lbalmaceda, line 369: might make sense to move this line above the "add scope" line, so that particular parameter always has precedence

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -113,6 +132,9 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         if (scope != null) {
             request.addParameter("scope", scope)
         }
+
+        request.addParameters(parameters)
```
lbalmaceda, line 136: might make sense to move this line above the "add scope" line, so that particular parameter always has precedence

lbalmaceda (changes requested):

lbalmaceda:
@lustikuss Thanks for the changes. It didn't come through with my review but the reason I'm asking you to remove the `override` is that the `BaseCredentialsManager` class was made public in the last major release (because the Kotlin compiler enforces that subclasses are scoped similarly) and so we cannot just add a new method to an abstract class as someone might be extending that today and be required to implement the new method now, which will be a breaking change. 

My suggestion is to add that new method to each of the manager subclasses and remove it from the "base interface". 

Please, remember to rebase the PR. Cheers

lustikuss:
Hi, thank you for your input. I rebased my branch and commited the fixes to your findings.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
lustikuss: I've removed the override in 0d28d3164799d560a5a422f7295ceea4ee573aed and added public explicitly because auf '-Xexplicit-api=strict'

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
lustikuss: I've removed the override in 0d28d3164799d560a5a422f7295ceea4ee573aed and added public explicitly because auf '-Xexplicit-api=strict'

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
lustikuss: I've made the changes in 0d28d3164799d560a5a422f7295ceea4ee573aed  because I was not sure if applying the suggestion and then rebasing the branch would maybe break it.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
lustikuss: I've made the changes in 0d28d3164799d560a5a422f7295ceea4ee573aed  because I was not sure if applying the suggestion and then rebasing the branch would maybe break it.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
lustikuss: done

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
lustikuss: and done

PR #539, opened 2022-02-03
Add Intent.FLAG_ACTIVITY_NEW_TASK flag for AuthenticationActivity

MichaelPersson:
### Changes

Add `Intent.FLAG_ACTIVITY_NEW_TASK` flag to intent for opening `AuthenticationActivity`. With this flag it is possible to open the login flow with an application context.
In current solution you can pass a context parameter but it needs to be an activity so maybe it's a bit misleading today. 

Using DI framework this would be really nice to be able to use application context. 

Another potential change would be to expose parameters so intent flags could be added depending on use case.

### References

Info about the flag.
https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_NEW_TASK

### Testing

I have done successful manual testing in our app where we're using the web login flow.

Could not find any Junit test that needed to change or be added.

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
@@ -99,6 +99,7 @@ public open class AuthenticationActivity : Activity() {
             intent.putExtra(EXTRA_AUTHORIZE_URI, authorizeUri)
             intent.putExtra(EXTRA_CT_OPTIONS, options)
             intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
             context.startActivity(intent)
         }
     }


MichaelPersson:
I guess that depends on how the architecture should look like.
In our case we can choose to either provide activity each time invoking the login flow or we could inject an application context. We're using Hilt where we could provide a class that use `@ActivityContext` parameter but unfortunately that will not work for us today. 
With that said I guess this change would be nice to have ;)

If applications context should not be supported then some restrictions could be added since this will be a runtime error.

poovamraj:
Rather than it being about Application Context, The major concern would be adding the intent flag that would affect all other users of the library. We will take into consideration on runtime error happening. Even we would like to move to Activity Context but haven't because of the backward compatibility issue it would cause. You can expect that in our next major.

I would rather recommend the initial suggestion you mentioned where the required Intents could be passed. It would be great to do that without complicating the current API. If it is possible could you make the PR changes more oriented towards it?

MichaelPersson:
I'll close this one and see if I can create a PR that adds more flexibility.

PR #563, opened 2022-06-01
[SDK-3346] Implemented coroutine support

poovamraj:
### Changes
We have added 2 new APIs
- `await()` for all requests
- `awaitCredentials()` for credential manager

The `await()` requests are called over IO thread and switched using `switchRequestContext`
`awaitCredentials` use `suspendCancellableCoroutine` to continue, this is especially useful when waiting for device unlock in `SecureCredentialManager`

### Reference
Credential Manager is an abstract class and hence the added APIs are directly added to the subclass similar to this PR to avoid breaking changes - https://github.com/auth0/Auth0.Android/pull/514

This will be cleaned up in the next major release.

### Testing
We have added unit tests, these tests are marked with `@ExperimentalCoroutineApi`, this shouldn't be an issue and they are marked only for test methods.


sample/src/main/java/com/auth0/sample/DatabaseLoginFragment.kt
```diff
@@ -40,6 +51,10 @@ class DatabaseLoginFragment : Fragment() {
         savedInstanceState: Bundle?
     ): View {
         val binding = FragmentDatabaseLoginBinding.inflate(inflater, container, false)
+        val authentication = AuthenticationAPIClient(account)
+        val storage = SharedPreferencesStorage(requireContext())
+        manager = SecureCredentialsManager(requireContext(), authentication, storage)
+        manager.requireAuthentication(requireActivity(), 201, "Hi Hello", "Poovam");
```
adamjmcgrath, line 57: 👋 

You've left some testing code in here

poovamraj: My bad 🤦  Great catch though. I have fixed it now, there should be zero change in that file

adamjmcgrath (dismissed):

adamjmcgrath (approved):

PR #564, opened 2022-06-06
createdAt should be deserialized as ISO8601 UTC (not local time)

### LINKED ISSUE #553: UserProfile createdAt wrong date - does not respect the GMT
### Describe the problem

I have created an account at "Fri Apr 29 10:09:23 GMT+03:00 2022", but the UserProfile returns createdAt "Fri Apr 29 07:09:23 GMT+03:00 2022". I assume it is the time without the GMT. For the Credentials expiresAt, the time is correct "Sat Apr 30 10:09:25 GMT+03:00 2022" - current date plus one day I assume.

From the comments on the code, I only see comment about Credentials.expiresA, but none for UserProfile.createdAt. For me at least, it would have no logic in returning dates in multiple formats.
`

    /**
     * Getter for the expiration date of the Access Token.
     * Once expired, the Access Token can no longer be used to access an API and a new Access Token needs to be obtained.
     *
     * @return the expiration date of this Access Token
     */
    @field:SerializedName("expires_at") public val expiresAt: Date
`

### What was the expected behavior?

I would expect for the UserProfile.createdAt date to be correct and respect the GMT, as the Credentials.expiresAt does. 

### Reproduction

We are using Auth0 Universal login. To reproduce this issue:
1. Create a new account using "WebAuthProvider.login(auth0)" with "params["screen_hint"] = "signup""
2. Get the user profile using "client.userInfo(credentials.accessToken)"

### Environment

- **Version of this library used:2.6.0**
- **Which framework are you using, if applicable: Android Studio Bumblebee | 2021.1.1 Patch 3**
- **Other modules/plugins/libraries that might be involved: No**
- **Any other relevant information you think
 would be useful:**

Screen Shot 2022-04-29 at 10 17 20 AM
Screen Shot 2022-04-29 at 10 16 44 AM

adamjmcgrath:
### Changes

Fix issue where `createdAt` from the `UserProfile` was being interpreted as a local date rather than UTC.

`/api/v2/users` (and `/userinfo` for old clients that have OIDC conformant disabled) will issue User Profiles with `created_at` and `updated_at` dates in ISO8601 format.

image

see [get_users_by_id](https://auth0.com/docs/api/management/v2#!/Users/get_users_by_id)

We used `SimpleDateFormat` through Gson by supplying a `dateFormat`, but `SimpleDateFormat`  doesn't support ISO8601 until API v24. And since the API put a `Z` at the end of the date to indicate UTC following ISO8601, we ignored the `Z` (timezone indicator) by putting it in quotes `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` (here https://github.com/google/gson/blob/gson-parent-2.8.9/gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java#L85).

With no timezone information Gson deserialized it as a date time local to the device. If the device is in GMT + 3, it assumes the date is GMT + 3, when it's actually UTC (`Z` = "UTC")

Gson supports ISO8601 and it's the fallback when no suitable date formats are supplied (see https://github.com/google/gson/blob/gson-parent-2.8.9/gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java#L85). So I've changed the default Gson builder to not use `SimpleDateFormat` and instead fallback to its `ISO8601Utils`.

I had to leave the serialization/de-serialization of dates in `SecureCredentialsManager` the same, because user's devices will have the local time stored with the ISO8601 UTC timezone indicator at the end (the `Z`) and these still need to be interpreted as local times since this is how they were stored.

### References

fixes #553

### Testing

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)

- [ ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)


auth0/src/test/java/com/auth0/android/request/internal/UserProfileGsonTest.kt
```diff
@@ -235,14 +234,28 @@ public class UserProfileGsonTest : GsonBaseTest() {
...
+                getUTCDate(2014, 6, 6, 18, 33, 49, 5)
             )
         )
     }
 
+    @Test
+    @Throws(Exception::class)
+    public fun shouldCreateProfileFromISO8601CreatedAt() {
+        val profileUtc = pojoFrom(
+            StringReader("""{ "created_at": "2022-04-29T07:09:23.000Z" }"""),
```
adamjmcgrath, line 246: Used the example in #553

poovamraj (approved):
+1

PR #575, opened 2022-06-16
[3353] Validate claims for ID Token received in Authentication API Client

poovamraj:
### Changes

- We have added a new method `validateClaims()` for Authentication Request and Signup Request
- `withIdTokenVerificationIssuer` and `withIdTokenVerificationLeeway` are added to customize the claim validation
- README and documentation are updated
- IdTokenVerifier now has an option where signature verification is optional
- IdTokenVerificationOptions is converted to Kotlin

### References
We have internal discovery regarding this linked in the ticket mentioned in title

### Testing
We have added unit tests and tested it manually as well


README.md
```diff
@@ -528,9 +518,38 @@ users
...
+```
+
+For Authentication Client, the method `validateClaims()` has to be called to enable it.
+
+```kotlin
+val auth0 = Auth0("YOUR_CLIENT_ID", "YOUR_DOMAIN")
+val client = AuthenticationAPIClient(auth0)
+client
+     .login("{username or email}", "{password}", "{database connection name}")
+     .validateClaims()
```
Widcket, line 542: What is the rationale behind adding a method for this instead of having it happen automatically and transparently, while having methods to customize issuer, maxAge, etc?

poovamraj: @Widcket that could break users who are still using Auth0 Private Cloud with ["Custom Domains"](https://auth0.com/docs/custom-domains) still on the [legacy behavior](https://auth0.com/docs/private-cloud/private-cloud-migrations/migrate-private-cloud-custom-domains#background)

That is why we have not enabled this by default.

Widcket: But is it worth it compromising the API design over that, instead of pushing the feature to the next major? Because if you make it optional, not many customers will use it. Not even with lots of documentation.

poovamraj: This method will be made default and removed in the next major which can handle breaking changes. Until then we wanted to provide an option where the validation can still be made without breaking the current major version.

This is mentioned in AuthenticationRequest.kt Line 54. We have followed a similar behaviour in Java SDKs to provide a feature as optional to make it mandatory in the next release.

Widcket: Thanks for the clarification 👍🏼

PR #579, opened 2022-06-22
[SDK-3358] Improve Android README

adamjmcgrath:
### Changes

- Added Java examples
- Added coroutine examples
- Added TOC
- Other small updates, typos/reformatting etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,20 @@ WebAuthProvider.login(account)
 ```
 
 <details>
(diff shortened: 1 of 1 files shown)


poovamraj:
@adamjmcgrath I removed few code that is very similar in sample. What do you think about it? I guess there will be lesser consistency but just adding `;` would be too much right?

poovamraj (approved):

PR #610, opened 2022-10-13
[SDK-3676] Readme redesign

poovamraj:
### Changes

- Changed to new README design
- Moved FAQ and EXAMPLES to separate page to keep the main README concise and also have enough headers to parse through information easier in respective pages


README.md
```diff
@@ -6,32 +6,19 @@
 [![Maven Central](https://img.shields.io/maven-central/v/com.auth0.android/auth0.svg?style=flat-square)](https://search.maven.org/artifact/com.auth0.android/auth0)
 [![javadoc](https://javadoc.io/badge2/com.auth0.android/auth0/javadoc.svg)](https://javadoc.io/doc/com.auth0.android/auth0)
```
jimmyjames, line 8: Any reason the other badges from the figma design (Circle status, license, coverage, etc.) are not added here?

README.md
```diff
@@ -6,32 +6,19 @@
...
-   * [Encryption enforced](#encryption-enforced)
-- [Networking client customization](#networking-client-customization)
-- [FAQ](#faq)
-- [Proguard](#proguard)
-- [What is Auth0?](#what-is-auth0)
-- [Issue Reporting](#issue-reporting)
-- [License](#license)
-
-## Requirements
+📚 [Documentation](#documentation) • 🚀 [Getting Started](#getting-started) • ⏭️ [Next Steps](#next-steps) • 💬 [Feedback](#feedback)
```
jimmyjames, line 9: Might be best to use the the emoji shortcuts (e.g., `:books:`) instead of pasting the emoji directly in, for MD parsers.

jimmyjames (commented):
I left a couple questions inline. I also noticed the `README` is still fairly lengthy. Do we have some opportunity to move some of the non-basic info to the examples or FAQ? Just at a glance, stuff like encryption, Proguard, etc. Just wondering if we can further tighten the README a bit more and move details to examples/faq/api reference.

jimmyjames:
It looks like the build is failing due to the ship orb config. You can see `java-jwt` for the ship config there, should be fairly similar.

stevehobbsdev:
I second @jimmyjames' comment about the length, according to the guide everything that's basically not login should be moved out to `EXAMPLES` or `FAQ` where possible.

Also this PR does not include the "Configure Auth0" section.

README.md
poovamraj: Excepts for downloads which is not available through maven, I guess we have covered everything @jimmyjames. And we have added `javadoc` badge to the mix

Design - 
Screenshot 2022-10-19 at 16 48 25

README preview -
Screenshot 2022-10-19 at 16 48 17

poovamraj:
@jimmyjames @stevehobbsdev I have updated the README from the comments. I haven't changed a few things and providing context here

1. Still using emojis instead of Github markdown syntax as emojis are unicode encoded and can be rendered in more places than github syntax. We already had issues in other repos where doc generators didn't work well with Github syntax

2. We have moved Proguard, Unit testing and Secure credential manager details to EXAMPLES but Credentials Manager details are still in README as I feel it is a basic feature of the SDK which should be noted in the front.

3. We haven't added "Configure Auth0" section as they can be found in Quickstart and README can go into the nuances of the SDK.

README.md
```diff
@@ -1,37 +1,24 @@
...
-   * [Encryption enforced](#encryption-enforced)
-- [Networking client customization](#networking-client-customization)
-- [FAQ](#faq)
-- [Proguard](#proguard)
-- [What is Auth0?](#what-is-auth0)
-- [Issue Reporting](#issue-reporting)
-- [License](#license)
-
-## Requirements
+📚 [Documentation](#documentation) • 🚀 [Getting Started](#getting-started) • ⏩ [Next Steps](#next-steps) • ❓ [FAQs](https://github.com/auth0/auth0.android/blob/main/FAQ.md) • 💬 [Feedback](#feedback)
```
ewanharris, line 9: Leftover next steps link I think, also I don't think the FAQs needs to be linked here if it's included in the `Documentation` section below

```suggestion
📚 [Documentation](#documentation) • 🚀 [Getting Started](#getting-started) • 💬 [Feedback](#feedback)
```

poovamraj: Hey @ewanharris I was trying to be consistent with the Figma design. We do have FAQ and API Reference. Since we didn't have API Reference in the same page as Next, I added "Next Steps". What do you think about that?

image

poovamraj: Looks like what I used is an older design on the same page. Committing this suggestion :)

ewanharris (approved):

Issue #618, opened 2022-11-10
Retry of SecureCredentialsManager.saveCredentials after CredentialsManagerException fails again if old keys are present


paulyoungtt:
### Describe the problem

Users who had been using our app for a while, and then upgraded their device or otherwise modified their lock screen settings, were no longer able to sign in.

We traced this to an issue with the way cleanup is done when SecureCredentialsManager.saveCredentials encounters an error. Newer RSA/AES keys are cleaned up, allowing for retry, but older keys are not. As a result, the app's attempt to retry the call to saveCredentials resulted in the same exception being thrown.

### What was the expected behavior?

The retry would succeed & the user would be able to sign in.

### Reproduction

It's a pretty heavy lift to repro to be honest. You'd have to start with a client using an older version of the Auth0 SDK that stores credentials using the old key alias. You then upgrade the client to use the current version of the Auth0 SDK. You then also update the device so that the key store security changes enough to throw one of the corresponding exceptions leading to the attempt at cleaning up the stored keys.

Rather than trying to build up a repro case, we verified it by adding logging in our app and providing it to specific customers who were experiencing the failure, confirming that they had the older key alias values in their credentials storage, and then validated the potential fix by performing the removal of the offending older keys from storage prior to our retry of the saveCredentials call. So while this is an available workaround, this seems like it should be fixed in the library itself.

### Environment

Auth0.Android 2.8.0

poovamraj:
Hi @paulyoungtt that is a pretty detailed issue report, thanks for that. Is it anyway possible to figure out what was the old version of Auth0 that the user was using?

paulyoungtt:
> Hi @paulyoungtt that is a pretty detailed issue report, thanks for that. Is it anyway possible to figure out what was the old version of Auth0 that the user was using?

Hi @poovamraj, thanks for following up. It looks like we started at version 1.15.2.

poovamraj:
Hey @paulyoungtt Thanks a lot for this great catch and even fantastic PR. We have included it in our next release and we will keep you updated about the release.

PR #619, opened 2022-11-10
Remove old encryption keys from KeyStore & storage when cleaning up on failure.

paulyoungtt:
* Remove old encryption keys from KeyStore & storage when cleaning up on failure.
* Update tests, including various tests that were failing to verify what they thought they were verifying.

The background here is we were getting reports of users being unable to sign in with valid credentials, and we traced it to the scenario of the lock screen settings change causing the KeyStore needing to be reset. However these users also had had our app long enough that their keys were stored using the older key aliases, and the logic to clean up and allow for retry after the first CredentialsManagerException was failing to account for this combination, and the retry would fail in the same way because those older keys were still present.

In updating the tests to include verification of the new desired behavior, I identified that many of the CryptoUtil tests used a pattern of verifying within an exception verification block after the exception was thrown, causing those verify steps to be skipped. I updated those tests to move the verify steps outside of the exception verification block. This resulted in one test now failing unrelated to my changes. I broke that test into 2 tests to account for the failure. I also updated the names of a few tests that were claiming to validate one outcome but were actually validating a different outcome.

### References

See issue #618 

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CryptoUtil.java
@@ -238,6 +238,7 @@ class CryptoUtil {
             KeyStore keyStore = KeyStore.getInstance(ANDROID_KEY_STORE);
             keyStore.load(null);
             keyStore.deleteEntry(KEY_ALIAS);
+            keyStore.deleteEntry(OLD_KEY_ALIAS);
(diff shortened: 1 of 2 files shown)


poovamraj (approved):
This is a great PR! Thanks a lot for the contribution @paulyoungtt. We will add this fix to our next release

Issue #617, opened 2022-11-02
Open class Auth0 is effectively final since 2.0.0


julia-martemianova:
### Describe the problem you'd like to have solved
The problem was previously described here: https://github.com/auth0/Auth0.Android/issues/611. Creating this issue to reopen it.
The scenario in which the URL is changing: for the social login connection we have the following path which we cannot change: `/connect/authorize`. To make it work we have to maintain a fork of the library with a small customisation.

### Describe the ideal solution
Add the open keyword to each fun/val/var in Auth0.kt.

### Alternatives and current work-arounds
N/A

### Additional information, if any
Minimal repro to show what's the problem:
```kotlin
open class Auth0 {
    public val authorizeUrl: String get() = "..."
}
class MyAuth0 : Auth0() {
    // [OVERRIDING_FINAL_MEMBER] 'authorizeUrl' in 'Auth0' is final and cannot be overridden
    override val authorizeUrl: String get() = "/somethingelse/..."
}
```

poovamraj:
Hi @julia-martemianova 👋 

This seems to be a popular feature request. I guess instead of opening the Auth0 class, we can make the `authorizeUrl` mutable similar to `networkingClient` (Need to verify this doesn't have any unwanted side effects though)

But before that can you confirm which Social Login requires such scenario? As it is used only in Web Authentication and currently in Auth0 universal login we can't change the Authorization URL.

julia-martemianova:
Hi @poovamraj  thanks for getting back to me!
I double checked. We need to test our identity server integration for social login in production before releasing it. For that we need a different path to initiate the social login connection. I you could make the `authorizeUrl` mutable, that would be great. Because apart from the pain of maintaining a fork we are also not able to use latest versions of the library.

poovamraj:
@julia-martemianova Just to confirm, this " identity server" is not Auth0?

julia-martemianova:
> Just to confirm, this "identity server" is not Auth0?
@poovamraj yes

poovamraj:
@julia-martemianova since this SDK is designed for working with Auth0 the requested feature is missing in the newer versions. Due to other priorities in the pipeline, we couldn't concentrate on this request right away. However, we wouldn't mind merging a PR with the requested feature and can help you with it from our end.

poovamraj:
Hi @julia-martemianova, we have created a PR to fix this - https://github.com/auth0/Auth0.Android/pull/622. Do check it out

[...]

Issue #611, opened 2022-10-14
open class Auth0 is effectively final since 2.0.0


TWiStErRob:
### Describe the problem you'd like to have solved
We need to override the `Auth0.authorizeUrl` from `/authorize` to `/somethingelse/authorize`.
In the current implementation (old 1.25) we can use `class MyAuth0 extends Auth { @Override public String getAuthorizeUrl() { ... } }`.
Sadly during the [Kotlin konversion](https://github.com/auth0/Auth0.Android/commit/ba9ef068d644df0d9a5d8d11d236613ef5273d46) in 2.0.0, `class Auth0` stayed `open`, but there are no overridable funs/vals/vars in it.

### Describe the ideal solution
Add the `open` keyword to each `fun`/`val`/`var` in `Auth0.kt`.

## Alternatives and current work-arounds
N/A

## Minimal repro to show what's the problem:
```kotlin
open class Auth0 {
    public val authorizeUrl: String get() = "..."
}
class MyAuth0 : Auth0() {
    // [OVERRIDING_FINAL_MEMBER] 'authorizeUrl' in 'Auth0' is final and cannot be overridden
    override val authorizeUrl: String get() = "/somethingelse/..."
}
```

cc @aeramli

poovamraj:
Hi @TWiStErRob, we will close this for now since we couldn't really see a scenario where this needs to be supported.  Please feel free to reopen this issue with more details and we will be happy to help.

poovamraj:
Hey @TWiStErRob, we have fixed this here - https://github.com/auth0/Auth0.Android/pull/622. We will notify once this is released.

PR #622, opened 2022-12-02
Allow `authorizeUrl` and `logoutUrl` customisation

poovamraj:
### Changes

We will open up `authorizeUrl` and `logoutUrl`. This has been a request from the community who use this SDK for other IdP.

(Also fixed another related test)

### References

https://github.com/auth0/Auth0.Android/issues/617
https://github.com/auth0/Auth0.Android/issues/611

### Testing
We have added unit tests to ensure these methods are non final and accessible publically

### Checklist

### DIFF
--- a/auth0/src/test/java/com/auth0/android/Auth0Test.java
+++ b/auth0/src/test/java/com/auth0/android/Auth0Test.java
@@ -203,6 +203,20 @@ public class Auth0Test {
     }
 
     @Test
+    public void shouldAllowOverridingAuthorizeUrl() {
+        class MyAuth extends Auth0 {
+            public MyAuth() {
+                super(CLIENT_ID, DOMAIN);
+            }
+            @Override
+            public String getAuthorizeUrl() {
+                // Calling super to make sure it's fully visible as well as overridable.
+                return super.getAuthorizeUrl() + "something";
+            }
+        }
+    }
+
+    @Test
     public void shouldEnsureLogoutUrlIsOpen() throws NoSuchMethodException {
         Method method = Auth0.class.getMethod("getLogoutUrl");
         Assert.assertTrue(Modifier.isPublic(method.getModifiers()));
--- a/auth0/src/test/java/com/auth0/android/Auth0Test.java
+++ b/auth0/src/test/java/com/auth0/android/Auth0Test.java
@@ -26,6 +26,9 @@ import static org.hamcrest.Matchers.notNullValue;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.when;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
 @RunWith(RobolectricTestRunner.class)
 public class Auth0Test {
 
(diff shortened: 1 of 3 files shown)


auth0/src/test/java/com/auth0/android/Auth0Test.java
```diff
@@ -191,4 +194,18 @@ public void shouldThrowWhenHttpUppercaseDomainUsed() {
     public void shouldThrowWhenConfigDomainIsHttp() {
         Assert.assertThrows("Invalid domain url: 'http://" + OTHER_DOMAIN + "'. Only HTTPS domain URLs are supported. If no scheme is passed, HTTPS will be used.", IllegalArgumentException.class, () -> new Auth0(CLIENT_ID, DOMAIN, "HTTP://" + OTHER_DOMAIN));
     }
+
+    @Test
+    public void shouldEnsureAuthorizeUrlIsOpen() throws NoSuchMethodException {
+        Method method = Auth0.class.getMethod("getAuthorizeUrl");
+        Assert.assertTrue(Modifier.isPublic(method.getModifiers()));
+        Assert.assertFalse(Modifier.isFinal(method.getModifiers()));
+    }
```
TWiStErRob, line 203: This tests that the `val` is `open`, but it can be very easily broken to trigger a false positive:
 * if someone makes the `Auth0` class non-`open` in the future, it's not a compile error, and this test will still pass, but in essence inheritance is forbidden and therefore the functionality the test is verifying is broken.
 * if someone makes the `Auth0` class have no protected/public constructors, the class becomes effectively final again. Adding a private constructor like `open class Auth0 private constructor() {` will not even emit a warning, it's all good from `kotlinc`'s point of view, but the class becomes non-extensible.
 * Removing `JvmOverloads` might also have a bad effect.

So I would propose to write this test this way and let the compiler verify it:
```java
@Test
public void shouldAllowOverridingAuthorizeUrl() {
	class MyAuth extends Auth0 {
		public MyAuth() {
			super(CLIENT_ID, DOMAIN);
		}
		@Override
		public @NotNull String getAuthorizeUrl() {
			// Calling super to make sure it's fully visible as well as overridable.
			return super.getAuthorizeUrl() + "something";
		}
	}
}
```
alternatively:
```java
@Test
public void shouldAllowOverridingAuthorizeUrl() {
	// Creating an instance as well, as opposed to just `class MyAuth extends Auth0`
	// to make sure that the constructor is not doing funny business.
	new Auth0(CLIENT_ID, DOMAIN) {
		@Override
		public @NotNull String getAuthorizeUrl() {
			// Calling super to make sure it's fully visible as well as overridable.
			return super.getAuthorizeUrl() + "something";
		}
	}
}
```

TWiStErRob (approved):
LGTM, thanks for the ping in #611.

julia-martemianova (dismissed):
Thanks a lot for the changes!

auth0/src/test/java/com/auth0/android/Auth0Test.java
poovamraj: Thanks @TWiStErRob, that is a great catch. I have added this to our test suite

jimmyjames (approved):

Issue #640, opened 2023-03-21
Android 12+ Crash - Missing Pending Intent Flag


SeanoScribd:
### Describe the problem

Crash when attempting to view login redirect with Trusted Web Activity on Android 12+

Regular redirect via:

```
    WebAuthProvider.login(account)
        .withScheme("https")
        .start(context, callback)
```

works fine.

Gradle SDK levels:
```
    compileSdk 33
    minSdk 21
    targetSdk 33
```

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.auth0test, PID: 29631
    java.lang.IllegalArgumentException: com.example.auth0test: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
    Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
        at android.app.PendingIntent.checkFlags(PendingIntent.java:401)
        at android.app.PendingIntent.getActivityAsUser(PendingIntent.java:484)
        at android.app.PendingIntent.getActivity(PendingIntent.java:470)
        at android.app.PendingIntent.getActivity(PendingIntent.java:434)
        at androidx.browser.customtabs.CustomTabsClient.createSessionId(CustomTabsClient.java:231)
        at androidx.browser.customtabs.CustomTabsClient.newSession(CustomTabsClient.java:269)
        at com.google.androidbrowserhelper.trusted.TwaLauncher$TwaCustomTabsServiceConnection.onCustomTabsServiceConnected(TwaLauncher.java:324)
        at androidx.browser.customtabs.CustomTabsServiceConnection.onServiceConnected(CustomTabsServiceConnection.java:57)
        at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:2188)
        at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:2221)
        at android.os.Handler.handleCallback(Handler.java:942)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
        at android.os.Looper.loop(Looper.java:288)
        at android.app.ActivityThread.main(ActivityThread.java:7898)
        at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #641, opened 2023-03-27
Update dependencies

poovamraj:
### Changes

- We have updated dependencies. This is due to an [issue](https://github.com/auth0/Auth0.Android/issues/640) in  `androidx.browser` that made us update to 1.4.0. 
- Updated other dependencies as well as cleanup

### ⚠️ Breaking Change
This version of Auth0.Android enables developers to use the library when building projects that target the API level 31. 

Developers using this version of Auth0.Android will need to update the the compileSdkVersion in their build.gradle to version 31. It's still possible to use targetSdkVersion lower than 31, but it's recommended to migrate as soon as possible.

When migrating to targetSdkVersion 31 and targeting Android 12 or higher, activities, services, or broadcast receivers that use intent filters, [must now explicitly declare the android:exported ](https://developer.android.com/about/versions/12/behavior-changes-12#exported)attribute for these app components.

### References

https://github.com/auth0/Auth0.Android/issues/640

### Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 
 Android API version 31 or later and Java 8+.
 
-> :warning: For Apps compiling and targeting using Android SDK version 30 and below can use Auth0.Android version 2.9.0
+> :warning: Applications targeting Android SDK version 30 and below should use version 2.9.0.
 
 Here’s what you need in `build.gradle` to target Java 8 byte code for Android and Kotlin plugins respectively.
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,9 @@
 
 ### Requirements
 
-Android API version 21 or later and Java 8+.
+Android API version 31 or later and Java 8+.
+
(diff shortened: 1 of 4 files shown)


ewanharris (dismissed):

README.md
```diff
@@ -25,7 +25,9 @@
 
 ### Requirements
 
-Android API version 21 or later and Java 8+.
+Android API version 31 or later and Java 8+.
+
+> :warning: For Apps compiling and targeting using Android SDK version 30 and below can use Auth0.Android version 2.9.0
```
jimmyjames, line 30:
```suggestion
> :warning: Applications targeting Android SDK version 30 and below should use version 2.9.0.
```

jimmyjames (approved):
Just a suggestion for the documentation, otherwise looks good.

jimmyjames (approved):

Issue #648, opened 2023-04-13
idToken not being provided if not manually added via scope property result in exception thrown


ryandary:
https://github.com/auth0/Auth0.Android/blob/a0e5d7464fb3f64d65def18585335de6087e5f4c/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt#L539

The language of the documentation suggests that the scope that are used when performing the initial auth will be applied on the renew, however it seems that this is not always the case.  We found that when calling renew without specifically adding the scope parameter, the result of the call is an exception because the payload returned by the server did not contain the idToken.

poovamraj:
Hi @ryandary 

> the response will include an id_token and an access_token if 'openid' scope was requested when the refresh_token was obtained

The documentation does say "if" openid scope is present. Have any suggestions to make this better?

ryandary:
That's clear.  In my case, it is being provided in auth, so my expectation is that it will be provided in refresh.

poovamraj:
@ryandary the `renewAuth` method will work only based on the parameters provided. It doesn't know of any previous state. You can consider it as a pure function that isn't influenced by previous states.

However what you can do is use our `SecureCredentialsManager` to manage your token and it will renew the token depending on the scope and does the management internally and securely. Hope this answers your questions. Feel free to let me know if you have more doubts and I'd be happy to help.

ryandary:
Perhaps the phrase that is confusing is "if 'openid' scope was requested when the refresh_token was obtained" as this suggests some prior state that is being considered.  The "was requested... when.. was obtained" implies that some state is being remembered either internally on device, in the cloud or possibly in the token itself.

It is unclear to the caller (me in this case) that scope needed to be passed explicitly and that it needed to match what was passed at initial auth.  The sentence that I started this thread with implies that as long as the initial auth included that scope that this function would also return the tokens.

ryandary:
The fact that the exception is thrown seems to suggest that the internal implementation also makes expectations that are based on state rather than on simply passed-in values.

poovamraj:
Can you share the exception that you are referring to here?

ryandary:
Here is a snippet.  This is caused by calling renewAuth on AuthenticationAPIClient without using the addParameter and passing the scope and the initial auth was called by passing the scope "openid email offline_access".

Caused by: com.auth0.android.Auth0Exception: Something went wrong
[...]

PR #654, opened 2023-05-02
Add required scopes to token and renewAuth requests

poovamraj:
### Changes
We have added required scope to `renewAuth` and `token` requests in `AuthenticationAPIClient`. Up until now these methods wouldn't have worked without adding scope "openid" to the request. This didn't have effect on internal usage as we always added the default scopes. But usage outside through the API client will fail always until the scope is added.

### References
https://github.com/auth0/Auth0.Android/issues/648

### Testing

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.kt
@@ -181,6 +181,12 @@ public class ParameterBuilder private constructor(parameters: Map<String, String
                 .setScope(OidcUtils.DEFAULT_SCOPE)
         }
 
+        @JvmStatic
+        public fun newBuilderWithRequiredScope(): ParameterBuilder {
+            return newBuilder()
+                .setScope(OidcUtils.REQUIRED_SCOPE)
+        }
+
         /**
          * Creates a new instance of the builder.
          *
--- a/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
@@ -555,7 +555,7 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
      * @return a request to start
      */
     public fun renewAuth(refreshToken: String): Request<Credentials, AuthenticationException> {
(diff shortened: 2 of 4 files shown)


auth0/src/main/java/com/auth0/android/authentication/ParameterBuilder.kt
```diff
@@ -181,6 +181,12 @@ public class ParameterBuilder private constructor(parameters: Map<String, String
                 .setScope(OidcUtils.DEFAULT_SCOPE)
         }
 
+        @JvmStatic
+        public fun newBuilderWithRequiredScope(): ParameterBuilder {
+            return newBuilder()
+                .setScope(OidcUtils.REQUIRED_SCOPE)
```
Widcket, line 187: If the developer is already adding the `openid` scope, will this result in it being duplicated?

poovamraj: If you see the `setScope` method, it will use `includeRequiredScope` which avoids repetition as you mentioned. There are existing internal mechanisms to avoid this.

d4rken:
@poovamraj 

This change caused unexpected behavior in our app.

Previously we specified `openid profile email offline_access` during login, When the accessToken expired and we refreshed, we do so with scope specified as `null`. According to docs this means that the previous scope is kept (scope specified during our login).


When using `getCredentials` and renewing the accessToken:

Before this change `newBuilder` is used for `renew`, it uses an empty parameter map, no scope is specified in the request, so the request doesn't contain a scope parameter and the server returns the previous scope.

With this change the `newBuilderWithRequiredScope` function is used and a default scope of `openid` is used, scope is no longer `null` and the server returns a new token with the scope `openid` and it's no longer the old scope.

Is this intended or a bug?

d4rken:
I compared `2.7.0` against `2.9.3` in the debugger.


```Kotlin
            if (scope != null) {
                request.addParameter("scope", scope)
            }
```

This re-adds the scope if one was specified in `getCredentials(scope = ...` but if none is specified? :thinking: Previously it used `storedScope`, but now it uses the scope from the new default constructor?

poovamraj:
@d4rken I agree, we have to revert this commit. We will cut a release soon (today or tomorrow)

d4rken:
Thanks for the quick response.

Issue #661, opened 2023-05-24
Crypto exception raised on save after refresh could be retried ?


bennycao:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

When saving credentials with `SecureCredentialsManager` a `CryptoException` could be thrown.
This could happen in the scenario of saving credentials after token refresh here https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt#L565. 

Would it be possible to `retry` and save the credentials as described here https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt#L190 ? For this given scenario ? This would prevent user being signed out, as there is no way to resave the credentials on the client side.

### Describe the ideal solution

Retry to save credentials in a token refresh scenario where CryptoException is thrown

poovamraj:
@bennycao I guess this isn't an ideal solution we can adopt. It is best to leave it to the users to decide how to handle such exceptions. Is there any specific scenario that is stopping you from achieving this?

bennycao:
> @bennycao I guess this isn't an ideal solution we can adopt. It is best to leave it to the users to decide how to handle such exceptions. Is there any specific scenario that is stopping you from achieving this?

@poovamraj this is in the scenario where it is a token refresh, which means we the client do not control or know about the refreshed credentials

poovamraj:
[...]

PR #666, opened 2023-06-25
Return refreshed Credentials in CredentialsManagerException to avoid logout

poovamraj:
### Changes
To avoid Logout because of not storing the refreshed credentials. We are providing the Credentials back to the user through our Exception.

Since Exceptions are highly logged we are masking sensitive data from it to avoid exposing them.

### References
https://github.com/auth0/Auth0.Android/issues/661

### Testing

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManagerException.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManagerException.kt
@@ -1,6 +1,7 @@
 package com.auth0.android.authentication.storage
 
 import com.auth0.android.Auth0Exception
+import com.auth0.android.result.Credentials
 
 /**
  * Represents an error raised by the [CredentialsManager].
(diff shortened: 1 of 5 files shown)


bennycao:
Hi @poovamraj , just want to understand what it means by the comment here https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt#L191C51-L191C51

it says clearcredentials so hasValidCredentials returns a true value. Doesn't clearing credentials mean hasValidCredentials will return false. And does the retry statement mean retrying of saveCredentials or something else ? It would be good to understand what the comment is referring too, and not go down this path if re-saving refreshed credentials won't help

poovamraj:
@bennycao the comment is mentioned to note _why_ we are calling the `clearCredentials` method. It is done so that `hasValidCredentials` won't return true when we can't fetch valid credentials due to corrupted keys.

jimmyjames (approved):

Issue #662, opened 2023-06-14
NullPointerException when errorDescription is null


Fraid:
### Checklist

- [X] The issue can be reproduced in the [Auth0.Android sample app](https://github.com/auth0-samples/auth0-android-sample/tree/master/00-Login-Kt) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Auth0 will send an unauthorized error but the errorDescription will be null causing a NullPointerException inside of the expected AuthenticationException.
`throw AuthenticationException(ERROR_VALUE_UNAUTHORIZED, errorDescription!!)`

I look in the code and it's null because:
`CallbackHelper.asMap` testing for `value==2` but for errorDescription it's `value==3 `
"error_description=Please verify your email before logging in.e=foo+t2@gmail.com"
So the key "error_description" is never set and then "OAuthManager.assertNoError" will crash with a NullPointerException.

Original app trace:
`java.lang.RuntimeException: Unable to resume activity {com.MY_PACKAGE.app/com.auth0.android.provider.AuthenticationActivity}: java.lang.NullPointerException
                                                                                                  at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4770)
                                                                                                  at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:4803)
                                                                                                  at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:57)
                                                                                                  at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
                                                                                                  at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:179)
[...]

PR #667, opened 2023-06-25
Avoid null pointer exception because of error description

poovamraj:
### Changes
To avoid causing an exception in case error description is null, we will pass unknown error description

### References
https://github.com/auth0/Auth0.Android/issues/662

### Testing

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
@@ -54,10 +54,8 @@ abstract class CallbackHelper {
         final String[] entries = valueString.length() > 0 ? valueString.split("&") : new String[]{};
         Map<String, String> values = new HashMap<>(entries.length);
         for (String entry : entries) {
-            final String[] value = entry.split("=");
-            if (value.length == 2) {
-                values.put(value[0], value[1]);
-            }
+            int idx = entry.indexOf("=");
+            values.put(entry.substring(0, idx), entry.substring(idx + 1));
         }
         return values;
     }
--- a/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java
@@ -55,7 +55,11 @@ abstract class CallbackHelper {
         Map<String, String> values = new HashMap<>(entries.length);
         for (String entry : entries) {
             int idx = entry.indexOf("=");
-            values.put(entry.substring(0, idx), entry.substring(idx + 1));
+            final String key = idx > 0 ? entry.substring(0, idx) : entry;
+            final String value = idx > 0 && entry.length() > idx + 1 ? entry.substring(idx + 1) : null;
+            if (value != null) {
+                values.put(key, value);
+            }
         }
         return values;
     }
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
@@ -192,6 +192,7 @@ internal class OAuthManager(
             TAG,
             "Error, access denied. Check that the required Permissions are granted and that the Application has this Connection configured in Auth0 Dashboard."
         )
+        val unknownErrorDescription = "An unexpected error occurred."
         when {
             ERROR_VALUE_ACCESS_DENIED.equals(errorValue, ignoreCase = true) -> {
                 throw AuthenticationException(
@@ -200,16 +201,16 @@ internal class OAuthManager(
                 )
             }
             ERROR_VALUE_UNAUTHORIZED.equals(errorValue, ignoreCase = true) -> {
-                throw AuthenticationException(ERROR_VALUE_UNAUTHORIZED, errorDescription!!)
+                throw AuthenticationException(ERROR_VALUE_UNAUTHORIZED, errorDescription ?: unknownErrorDescription)
             }
             ERROR_VALUE_LOGIN_REQUIRED == errorValue -> {
                 //Whitelist to allow SSO errors go through
-                throw AuthenticationException(errorValue, errorDescription!!)
+                throw AuthenticationException(errorValue, errorDescription ?: unknownErrorDescription)
             }
             else -> {
                 throw AuthenticationException(
                     errorValue,
-                    errorDescription ?: "An unexpected error occurred."
+                    errorDescription ?: unknownErrorDescription
                 )
             }
         }
--- a/auth0/src/test/java/com/auth0/android/provider/CallbackHelperTest.java
+++ b/auth0/src/test/java/com/auth0/android/provider/CallbackHelperTest.java
@@ -78,6 +78,19 @@ public class CallbackHelperTest {
     }
 
     @Test
+    public void shouldParseQueryValuesWithEqual() {
+        String uriString = "testandroid://auth.testandroid.com/android/com.testandroid.app/callback?error=unauthorized&error_description=Please%20verify%20your%20email%20before%20logging%20in.e%3Dfoo%2Bt2%40gmail.com&state=abscefg-QWERTYUIOPasdfgHJKLMNBVCXZdd";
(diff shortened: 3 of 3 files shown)


Fraid:
@poovamraj , thank this PR will prevent any other futur exception, as quick solution should be fine.

But for my case, it's not an unknown error. User need to verify the email, app should inform the user.

Knowing that I don't have access to auth0 admin, I only did the integration.
If this error `"error_description=Please verify your email before logging in.e=foo+t2@gmail.com"` come directly from auth0 API, I think this need to be handled ?
But If it's a custom message from auth0 admin, maybe prevent any admin to use custom "="?

For me an additional fix is required,  as an example I was thinking about something like:
```
if(values.lenght==3){
//It's an example! merging description content
values.put(value[0],String.format("%s %s",value[1],value[2]);
}
Result: "error_description":"Please verify your email before logging in.e foo+t2@gmail.com"
```
But I don't know your codebase, so carefully understand the use of `asMap()`, I don't know why there is filtering with "values.lenght=2",  must be a reason for that.

I have old screenshot when I did the debugging:
[And value of "values"](https://drive.google.com/file/d/1JWviw4mafuSiZT8In_P-FpPNoFfJj2k-/view?usp=sharing)
[breakpoint ](https://drive.google.com/file/d/1dY4SPGWgDNVNQ5DIXMThWjrPS9UuVzQ_/view?usp=sharing)

poovamraj:
Hi @Fraid, that makes sense. I am wondering the callback URL returned should be URL encoded. Are you saying it is not encoded? Then it wouldn't be possible for any parser to seperate this value without such delimiters

AndDev:
Hi @poovamraj, I'm not sure to understand your question, but I created the following test case (Don't know if it's compiling, but I think you'll get the idea).

Add following method [here](https://github.com/auth0/Auth0.Android/blob/ef24ca571335ca4095253044bbbb95879164e3d6/auth0/src/test/java/com/auth0/android/provider/CallbackHelperTest.java#L91):
   ```
@Test
    public void shouldParseQueryValuesWithEmail() {

//This url is returned as it is
        String uriString = "testandroid://auth.testandroid.com/android/com.testandroid.app/callback?error=unauthorized&error_description=Please%20verify%20your%20email%20before%20logging%20in.e%3Dfoo%2Bt2%40gmail.com&state=abscefg-QWERTYUIOPasdfgHJKLMNBVCXZdd";
        Uri uri = Uri.parse(uriString);
        final Map<String, String> values = CallbackHelper.getValuesFromUri(uri);

        assertThat(values, is(notNullValue()));
        assertThat(values, aMapWithSize(3));
        assertThat(values, hasEntry("error", "unauthorized"));
        assertThat(values, hasEntry("state", "abscefg-QWERTYUIOPasdfgHJKLMNBVCXZdd"));

//This will normally failed.
        assertThat(values, hasEntry("error_description", "Please verify your email before logging in.e=foo+t2@gmail.com"));
    }
```
With this temporary change for [asMap](https://github.com/auth0/Auth0.Android/blob/ef24ca571335ca4095253044bbbb95879164e3d6/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java#L50C5-L63C6) you should pass the test (temporary because I don't know every message that this method is handling) 
```
private static Map<String, String> asMap(@Nullable String valueString) {
        if (valueString == null) {
            return new HashMap<>();
        }
        final String[] entries = valueString.length() > 0 ? valueString.split("&") : new String[]{};
        Map<String, String> values = new HashMap<>(entries.length);
        for (String entry : entries) {
    final String[] value = entry.split("=");
    if (value.length == 2) {
        values.put(value[0], value[1]);
    } else if (value.length == 3) {
        values.put(value[0], String.format("%s=%s", value[1], value[2]));
    }
}
        return values;
    }
```

Hope this help.

**Comment by Fraid** [comment]
Hi @poovamraj, I'm not sure to understand your question, but I created the following test case (Don't know if it's compiling, but I think you'll get the idea).

Add following method [here](https://github.com/auth0/Auth0.Android/blob/ef24ca571335ca4095253044bbbb95879164e3d6/auth0/src/test/java/com/auth0/android/provider/CallbackHelperTest.java#L91):
   ```
@Test
    public void shouldParseQueryValuesWithEmail() {

//This url is returned as it is
        String uriString = "testandroid://auth.testandroid.com/android/com.testandroid.app/callback?error=unauthorized&error_description=Please%20verify%20your%20email%20before%20logging%20in.e%3Dfoo%2Bt2%40gmail.com&state=abscefg-QWERTYUIOPasdfgHJKLMNBVCXZdd";
        Uri uri = Uri.parse(uriString);
        final Map<String, String> values = CallbackHelper.getValuesFromUri(uri);

        assertThat(values, is(notNullValue()));
        assertThat(values, aMapWithSize(3));
        assertThat(values, hasEntry("error", "unauthorized"));
        assertThat(values, hasEntry("state", "abscefg-QWERTYUIOPasdfgHJKLMNBVCXZdd"));

//This will normally failed.
        assertThat(values, hasEntry("error_description", "Please verify your email before logging in.e=foo+t2@gmail.com"));
    }
```
With this temporary change for [asMap](https://github.com/auth0/Auth0.Android/blob/ef24ca571335ca4095253044bbbb95879164e3d6/auth0/src/main/java/com/auth0/android/provider/CallbackHelper.java#L50C5-L63C6) you should pass the test (temporary because I don't know every message that this method is handling) 
```
private static Map<String, String> asMap(@Nullable String valueString) {
        if (valueString == null) {
            return new HashMap<>();
        }
        final String[] entries = valueString.length() > 0 ? valueString.split("&") : new String[]{};
        Map<String, String> values = new HashMap<>(entries.length);
        for (String entry : entries) {
    final String[] value = entry.split("=");
    if (value.length == 2) {
        values.put(value[0], value[1]);
    } else if (value.length == 3) {
        values.put(value[0], String.format("%s=%s", value[1], value[2]));
    }
}
        return values;
    }
```

Hope this help.

poovamraj:
@Fraid how is that you are getting both the `state` which is retured when login is succesful and `error`/`error_description` which is returned when login fails. 

Are you using custom actions to achieve this work flow?

If so, you can login and then use the `email` and `isEmailVerified` returned from the `UserProfile` from `user` in `Credentials` to check for this.

It is interesting because I have never seen Auth0 return both error and state in the same callback URL.

Fraid:
@poovamraj I think I get what you mean, when I received the token, I'll try to get the profile first. 
I'll let you know soon...

Fraid:
@poovamraj I think we may not talking about the same thing, this is how I log. 
User will be prompt and at the return, app will crash. 
So I don't even have any callback nor any token:

```
 Map<String, Object> params = new HashMap<>();
        params.put("prompt", "login");
        WebAuthProvider.login(auth0)
                .withParameters(params)
                .withScheme(activity.getString(R.string.com_auth0_scheme))
                .withScope("openid email")
                .start(activity,.....)
```

poovamraj:
@Fraid yes I agree, the login wouldn't work in this case. The current way to parse values from callback URL is flaky. I am trying to find a way where we can dynamically get all the values from the callback URL instead of parsing using hardcoded numbers. With the suggested code, the same error will happen if we have 3 "=". So using "=" as delimiter is the real reason for the issue here

poovamraj:
@Fraid the current change should make it work

jimmyjames (approved):

fpitters:
Hi, on this matter, we just noticed that using
```public constructor(message: String, cause: Auth0Exception? = null) : super(message, cause)```
and calling `getDescription` raises an NPE.

Let me know if a new issue is worth it :)

poovamraj:
@fpitters description is a nullable property inside `AuthenticationException`. is the crash happening on the SDK or on your code?

fpitters:
Ops, it's actually my bad.

The NPE is raised only on unit tests and while using `returnDefaultValues`
Then `AuthenticationException.getDescription` which uses android `TextUtils.isEmpty` is trying to access `description` when actually null. Using a static mock solves the issue.

Sorry for the noise :( 

Maybe it would be good to move to Kotlin String.isEmpty :)

PR #672, opened 2023-07-18
Migrate deprecated AGP dsl

Goooler:
Something like `compileSdkVersion` is deprecated, we should migrate these API usages in both project and docs.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -11,7 +11,7 @@ When building the project with `build`, an error appeared regarding an `invalid 
 ```gradle
 android {
     //...
-    lintOptions {
+    lint {
        warning 'InvalidPackage'
     }
 }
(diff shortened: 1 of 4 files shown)


poovamraj:
@Goooler this PR doesn't seem to change gradle-wrapper.properties and other required things like adding `namespace`. Are you sure this is building locally?

Goooler:
Added `namespace`, I didn't update the wrapper files.

Goooler:
The wrapper validation job need to be approved, see https://github.com/auth0/Auth0.Android/actions/runs/5745829757

poovamraj:
Hey @Goooler looks like the build is failing and since we have are planning to do this migration ourselves in an upcoming initiative. We would like to close this PR. Thanks for the contribution though.

Issue #663, opened 2023-06-15
java.lang.SecurityException: Permission Denial: starting Intent  being thrown when starting auth page


bennycao:
### Checklist

- [X] The issue can be reproduced in the [Auth0.Android sample app](https://github.com/auth0-samples/auth0-android-sample/tree/master/00-Login-Kt) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

In our bug monitoring we are seeing uses receiving `java.lang.SecurityException: Permission Denial: starting Intent` exceptions which causes a crash. Full stack trace below
```
java.lang.SecurityException: Permission Denial: starting Intent { act=android.intent.action.VIEW dat=https://login..../... cmp=team.opay.pay/com.opay.webview.WebFoundationActivity (has extras) } from ProcessRecord{79a1d4e 32017:au..../u0a127} (pid=32017, uid=10127) not exported from uid 10099
        at android.os.Parcel.readException(Parcel.java:2013)
        at android.os.Parcel.readException(Parcel.java:1959)
        at android.app.IActivityManager$Stub$Proxy.startActivity(IActivityManager.java:4492)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at leakcanary.ServiceWatcher$install$4$2.invoke(ServiceWatcher.kt:93)
        at java.lang.reflect.Proxy.invoke(Proxy.java:913)
        at $Proxy2.startActivity(Unknown:-1)
        at android.app.Instrumentation.execStartActivity(Instrumentation.java:1611)
        at android.app.Activity.startActivityForResult(Activity.java:4502)
[...]

PR #677, opened 2023-07-31
Handle SecurityException thrown while launching the browser

poovamraj:
### Changes
This change handles exception thrown while launching browser instance for authentication.

One such error we wanted to handle is SecurityException: Permission Denial happening because of using browsers (mostly misconfigured applications) that have not `exported` their activity

### References
https://github.com/auth0/Auth0.Android/issues/663

### Testing

### DIFF
--- a/sample/src/main/res/values/strings.xml
+++ b/sample/src/main/res/values/strings.xml
@@ -1,6 +1,6 @@
 <resources>
     <string name="app_name">Auth0 SDK Sample</string>
-    <string name="com_auth0_domain">lbalmaceda.auth0.com</string>
-    <string name="com_auth0_client_id">PX2vJhpALUNT66NHdCdD30XWqlIR4oEZ</string>
+    <string name="com_auth0_domain">poovamraj.eu.auth0.com</string>
+    <string name="com_auth0_client_id">swarLtLAd0aW55ajPUPzYPXjtB3UtZki</string>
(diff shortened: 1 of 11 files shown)


sample/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,6 @@
 <resources>
     <string name="app_name">Auth0 SDK Sample</string>
-    <string name="com_auth0_domain">lbalmaceda.auth0.com</string>
-    <string name="com_auth0_client_id">PX2vJhpALUNT66NHdCdD30XWqlIR4oEZ</string>
+    <string name="com_auth0_domain">poovamraj.eu.auth0.com</string>
+    <string name="com_auth0_client_id">swarLtLAd0aW55ajPUPzYPXjtB3UtZki</string>
```
Widcket, line 4: Maybe we should replace these with placeholders?

Widcket:
CI is failing

sample/src/main/res/values/strings.xml
poovamraj: This would be the cleanest and straight forward way @Widcket.

Considering this is a public client this shouldn't be an issue. We use this in auth0-spa-js as well - [Link](https://github.com/search?q=repo%3Aauth0%2Fauth0-spa-js%20brucke.auth0.com&type=code)

Widcket (approved):

PR #687, opened 2023-09-20
Update Okio to resolve CVE-2023-3635

jimmyjames:
Excludes transient dependency okio from OkHttp and include 3.5.0 which resolves CVE-2023-3635. Once OkHttp 4.12.0 is released we can remove the manual okio dependency and exclude rule.

This change also updates the Kotlin version to 1.8, as Okio requires it beyond 3.2.0, as well as the Gradle version to 6.9.3


evansims (approved):

jimmyjames:
@evansims any idea why the snyk check is failing? It's unable to find a JAVA_HOME when executing gradle, and I notice the snyk runner is set to use the php one, is that correct? Also I don't see that workflow having run as part of the initial PR to add GH actions. Looks like the same issue is also blocking #690

poovamraj (approved):

PR #689, opened 2023-09-22
Use ThreadPool to launch browser for authentication

poovamraj:
### Changes
Previously we created a new instance of `Thread` to launch the browser for authentication. This caused flakiness on the CI as we had to rely on the execution order of the VM. We have used our own Thread pool now. This optimizes the resources used and also helps us avoid the flakiness in the CI.

### References
Failure of CI here and such in the past - https://github.com/auth0/Auth0.Android/pull/688

### Testing
We have added more verification now to ensure background switch is made

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
@@ -119,7 +119,7 @@ class CustomTabsController extends CustomTabsServiceConnection {
             return;
         }
 
-        new Thread(() -> {
+        threadSwitcher.backgroundThread(() -> {
             try {
                 if (launchAsTwa) {
                     this.launchedAsTwa = true;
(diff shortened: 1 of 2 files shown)


auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -119,7 +119,7 @@ public void launchUri(@NonNull final Uri uri, final boolean launchAsTwa, ThreadS
             return;
         }
 
-        new Thread(() -> {
+        threadSwitcher.backgroundThread(() -> {
```
jimmyjames, line 122: As the threadSwitcher param can be null, do we need a check here? And what do we do if it is null?

poovamraj: Hey @jimmyjames `CustomTabsController` is not public and used only internally only from one place. So this check wouldn't be highly required

Issue #676, opened 2023-07-25
Is there missing proguard rules ?


bennycao:
### Checklist

- [X] The issue can be reproduced in the [Auth0.Android sample app](https://github.com/auth0-samples/auth0-android-sample/tree/master/00-Login-Kt) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Hi, we are seeing in our error logs exception being thrown

`
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.annotation.Annotation java.lang.reflect.Field.getAnnotation(java.lang.Class)' on a null object reference
        at com.auth0.android.request.internal.JsonRequiredTypeAdapterFactory$1.read(JsonRequiredTypeAdapterFactory.java:35)
        at com.google.gson.TypeAdapter$1.read(TypeAdapter.java:199)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
        at com.auth0.android.request.internal.JsonRequiredTypeAdapterFactory$1.read(JsonRequiredTypeAdapterFactory.java:31)
        at com.google.gson.TypeAdapter$1.read(TypeAdapter.java:199)
        at com.google.gson.Gson.fromJson(Gson.java:963)
        at com.google.gson.Gson.fromJson(Gson.java:928)
        at com.google.gson.Gson.fromJson(Gson.java:877)
        at com.google.gson.Gson.fromJson(Gson.java:848)
[...]

PR #701, opened 2023-12-06
Guard against NullPointerException when getting Credentials from Json

bennycao:
### Changes

This change adds code to guard against possible NullPointerException thrown as a result of `f` being null in the JsonRequiredTypeAdaptor

Solves https://github.com/auth0/Auth0.Android/issues/676 

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)

- [ ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)


poovamraj:
@bennycao I'd rather prefer us to check if the field is null [here](https://github.com/auth0/Auth0.Android/blob/c17b87b4f67ddec78e5e6be9276c6f84a7bcd402/auth0/src/main/java/com/auth0/android/request/internal/JsonRequiredTypeAdapterFactory.java#L35) like below

```
if (f!= null && f.getAnnotation(JsonRequired.class) != null) {
```

Instead of catching the exception. It feels much cleaner and the check makes more sense. Can you try this solution out?

bennycao:
> @bennycao yes if the required value like `expiresAt` is not present for constructing the object, there will be an error thrown, but since this is fixing an issue that no one is able to reproduce, this is the only change that will introduce no new changes to the public API but at the same time solve your issue. If making this change introduces the new issue in your error logs, atleast you will be able to further debug it.

@poovamraj i've made the update as requested.

bennycao:
@poovamraj Any chance we can get this looked thanks ?

poovamraj (approved):

bennycao:
@poovamraj thanks for the approval. When do we think a new release with this fix be available ?
In terms of build status, i'm not sure what is happening there. But let me know if i need to do anything. Thanks

bennycao:
@poovamraj any movement on getting this into a release ? we would like to close this out asap. Thanks

poovamraj:
@bennycao you can go ahead with merging the PR after ensuring all checks are passed. We can cut a patch release once this is merged.

bennycao:
> @bennycao you can go ahead with merging the PR after ensuring all checks are passed. We can cut a patch release once this is merged.

@poovamraj it seems some build steps require authorizations to run. And i don't have rights to merge. Can you please authorize these ?

auth0/src/main/java/com/auth0/android/request/internal/JsonRequiredTypeAdapterFactory.java
```diff
@@ -32,7 +32,7 @@ public T read(JsonReader in) throws IOException {
 
                 Field[] fields = pojo.getClass().getDeclaredFields();
                 for (Field f : fields) {
-                    if (f.getAnnotation(JsonRequired.class) != null) {
```
RMSPORTS1902, line 35:
```suggestion
                    if (f.getAnnotation(JsonRequired.class) != null) {
```

RMSPORTS1902: Yes

PR #745, opened 2024-06-10
feat: implemented biometrics authentication for SecureCredentialsManager using androidx.biometrics package

desusai7:
### Migrated to use [BioMetricManager](https://developer.android.com/reference/androidx/biometric/BiometricManager) for performing Authentication before accessing credentials via SecureCredentialsManager

Breaking Changes: 

* Removed the `requireAuthentication` API which used to be a pre-requisite previously for performing authentication while retrieving Credentials as it is no longer needed.
* Constructor of Auth0 class is made private, and now you will have to use Auth0.getInstance(clientId, domain) to get an instance of Auth0, it would check if the instance with the given configuration exists already, if yes returns it, if not would create one and returns it.
* Added multiple overloads of getCredentials() & awaitCredentials() to the BaseCredentialsManager interface, now all the implementations of this interface would have to implemented these new methods using override as previously they used to define these methods directly in their implementations.
* await function of Request interface has been made abstract now and all the implementations of the interface should implement the method.
* Credentials class has been converted to a data class now this class can longer be extended and also the currentTimeInMillis property of Credentials class has been removed.

Other changes: 

* We've added a dependency on `androidx.biometrics` package to use the `BiometricsManager` for authentication purposes.
* Updated the constructor of `SecureCredentialsManager` to accept `fragmentActivity` & `authenticationOptions` which are utilised while performing authentication using `BiometricPrompt` before retrieving credentials.  It's no longer needed to call `requireAuthentication` before retrieving credentials.
* Added class [LocalAuthenticationManager](https://github.com/auth0/Auth0.Android/blob/feat/biometric/auth0/src/main/java/com/auth0/android/authentication/storage/LocalAuthenticationManager.kt) which consolidates all the logic for interacting with the `BiometricsManager`
* Added class [LocalAuthenticationOptions](https://github.com/auth0/Auth0.Android/blob/feat/biometric/auth0/src/main/java/com/auth0/android/authentication/storage/LocalAuthenticationOptions.kt) which helps to configure the behaviour of the `BiometricsManager` like `title`, `description` displayed within in the biometric prompt and the level of authentication required.
* Updated `CredentialsManagerException` to contain enum `Code` describing the reason for exception and allowing users to check all the cases of `CredentialsManagerException` and handle accordingly

### Testing

### Checklist


auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -90,4 +85,30 @@ public abstract class BaseCredentialsManager internal constructor(
...
+}
+
+public abstract class SecuredCredentialsManager internal constructor(
+    authenticationClient: AuthenticationAPIClient,
+    storage: Storage,
+    jwtDecoder: JWTDecoder
+) : BaseCredentialsManager(
+    authenticationClient, storage, jwtDecoder
+) {
+    public abstract fun getCredentials(fragmentActivity: FragmentActivity, authenticationOptions: LocalAuthenticationOptions, callback: Callback<Credentials, CredentialsManagerException>)
```
poovamraj, line 112: If FragmentActivity is required in both method, can't we have this as a class level dependency, that way you can have same abstract method for both cred managers.

We just have to ensure there is no context leak by having fragment activity at a class level

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManagerException.kt
```diff
@@ -6,10 +6,182 @@ import com.auth0.android.result.Credentials
...
  * Represents an error raised by the [CredentialsManager].
  */
-public class CredentialsManagerException internal constructor(
-    message: String,
-    cause: Throwable? = null
-) : Auth0Exception(message, cause) {
+public class CredentialsManagerException :
+    Auth0Exception {
+
+    internal enum class Code {
```
poovamraj, line 12: Nice cleanup 👍

poovamraj (commented):
Please take not of all the breaking changes in each PR so that it can be collected together for the major release. We don't want to lose track of them

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
desusai7: Agreed, updated the code accordingly, please take a look again.

sgammon:
Hi, we are an auth0 customer building for Android, and we are interested in seeing this ship. Is there any update?

poovamraj (dismissed):

poovamraj (approved):

desusai7:
> Hi, we are an auth0 customer building for Android, and we are interested in seeing this ship. Is there any update?

@sgammon, we will be releasing it shortly.

sgammon:
@desusai7 Awesome! Thank you so much, we're very excited! 😄

desusai7:
@sgammon, we've released it as part of version `3.0.0-beta.0`.

sgammon:
We are already using it! I haven't figured out how to use the new biometrics yet, but that is next on the list. I plan to check samples, test suites, etc., but if there is already documentation, please let me know 😄

desusai7:
> We are already using it! I haven't figured out how to use the new biometrics yet, but that is next on the list. I plan to check samples, test suites, etc., but if there is already documentation, please let me know 😄

documentation exists for it [here](https://github.com/auth0/auth0.android/blob/main/EXAMPLES.md#requiring-authentication)

PR #802, opened 2025-02-20
API support for fetching session token

pmathew92:
### Changes

This PR adds support to fetch a session token, in exchange for a refresh token, to be used for SSO

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -65,6 +65,10 @@ public abstract class BaseCredentialsManager internal constructor(
 
     @JvmSynthetic
     @Throws(CredentialsManagerException::class)
+    public abstract suspend fun awaitSsoCredentials(): SSOCredentials
+
+    @JvmSynthetic
+    @Throws(CredentialsManagerException::class)
     public abstract suspend fun awaitCredentials(): Credentials
 
     @JvmSynthetic
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -4,6 +4,7 @@ import androidx.annotation.VisibleForTesting
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.callback.Callback
 import com.auth0.android.result.Credentials
+import com.auth0.android.result.SSOCredentials
 import com.auth0.android.util.Clock
 import java.util.*
 
@@ -29,7 +30,9 @@ public abstract class BaseCredentialsManager internal constructor(
 
     @Throws(CredentialsManagerException::class)
     public abstract fun saveCredentials(credentials: Credentials)
+    public abstract fun saveSsoCredentials(ssoCredentials: SSOCredentials)
     public abstract fun getCredentials(callback: Callback<Credentials, CredentialsManagerException>)
+    public abstract fun getSsoCredentials(callback: Callback<SSOCredentials, CredentialsManagerException>)
     public abstract fun getCredentials(
         scope: String?,
         minTtl: Int,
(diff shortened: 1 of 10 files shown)


auth0/src/main/java/com/auth0/android/result/SSOCredentials.kt
```diff
@@ -0,0 +1,54 @@
+package com.auth0.android.result
+
+import com.google.gson.annotations.SerializedName
+
+/**
+ * Holds the session token credentials required for SSO .
+ *
+ *  * *sessionToken*: Session Token for SSO
```
Widcket, line 8:
```suggestion
 *  * *sessionToken*: Session Token for web SSO
```

auth0/src/main/java/com/auth0/android/result/SSOCredentials.kt
```diff
@@ -0,0 +1,54 @@
...
+ *  * *refreshToken*: Refresh Token that can be used to request new tokens without signing in again
+ *  * *tokenType*: Contains information about how the token should be used.
+ *  * *expiresIn*: The token expiration duration.
+ *  * *issuedTokenType*: Type of the token issued.
+ *
+ */
+public data class SSOCredentials(
+    /**
+     * The Session Token used for SSO .
```
Widcket, line 17:
```suggestion
     * The Session Token used for web SSO.
```

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -29,7 +30,9 @@ public abstract class BaseCredentialsManager internal constructor(
 
     @Throws(CredentialsManagerException::class)
     public abstract fun saveCredentials(credentials: Credentials)
+    public abstract fun saveSsoCredentials(ssoCredentials: SSOCredentials)
```
Widcket, line 33: I'm not sure this method is needed. We're already providing a method for getting fresh `SSOCredentials`.

Widcket: The same functionality can also be achieved by simply retrieving the credentials, replacing the refresh token, and saving them back.

Widcket: In the case we have it for internal code reuse, then we should not have it be public.

pmathew92: Added this method for the scenario where we get a new refresh token when calling the API directly from the AuthenticationAPI class. The user would have to call this to replace the existing one else the following getSsoCredentials call will fail.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
```diff
@@ -197,12 +214,153 @@ public class CredentialsManagerTest {
...
+    public fun shouldSaveTheRefreshTokenIfTheNewSsoRefreshTokenIsNotSameAsTheOldOne() {
+        verifyNoMoreInteractions(storage)
+        val ssoCredentials = SSOCredentials(
+            "accessToken",
+            "issuedTokenType", "tokenType", 60, "refresh_token"
+        )
+        Mockito.`when`(storage.retrieveString("com.auth0.refresh_token"))
+            .thenReturn("refresh-token")
+        manager.saveSsoCredentials(ssoCredentials)
+        verify(storage).store("com.auth0.refresh_token", "refresh_token")
```
Widcket, line 257: What is the difference between this test and the previous one?

pmathew92: The previous one checks for scenario when the existing refresh_token is the same as the one received in the session token credentials and doesn't save it. This one checks for when we have separate refresh_token than the existing one and save it

Widcket: Yes, but they seem to be identical, line-by-line (except for the assert, of course).

pmathew92: Agreed... everything is identical except the mocked value returned. which is what the test case is checking for . How the method will behave when the existing refresh_token is same and different from the new one

Widcket (approved):

PR #821, opened 2025-04-23
Added new error type to CredentialsManagerException class

pmathew92:
### Changes
This PR adds a new error type `UNKNOWN_ERROR` to the CredentialsManagerException class. This error is triggered during any unexpected error occurs when token refresh happens in the CredentialsManager classes

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
@@ -1,6 +1,7 @@
 package com.auth0.android.authentication.storage
 
 import android.text.TextUtils
+import android.util.Log
 import androidx.annotation.VisibleForTesting
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.authentication.AuthenticationException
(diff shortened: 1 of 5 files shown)


Widcket:
Should runtime exceptions also be caught when doing a token exchange for SSO credentials?

pmathew92:
> Should runtime exceptions also be caught when doing a token exchange for SSO credentials?

This is added to handle a specific use case when the customer is doing google auth and it crashes with NPE when doing token exchange for certain number of users . Both them and we are not able to reproduce this locally and the customer was also not able to handle this due the crash occuring on a background thread in the SDK. Hence added this check

Widcket:
> This is added to handle a specific use case when the customer is doing google auth and it crashes with NPE when doing token exchange for certain number of users .

For all we know, it could also happen when doing a token exchange for SSO credentials; for example, by adding a custom claim (using an action) that the SDK doesn't know how to decode. We don't know for sure that it is exclusive to the code path for regular refresh token exchanges. Since we are now catching runtime exceptions in regular refresh token exchanges, it stands to reason to do the same in that one other place where we're performing that same request.

pmathew92:
> For all we know, it could also happen when doing a token exchange for SSO credentials; for example, by adding a custom claim (using an action) that the SDK doesn't know how to decode. We don't know for sure that it is exclusive to the code path for regular refresh token exchanges. Since we are now catching runtime exceptions in regular refresh token exchanges, it stands to reason to do the same in that one other place where we're performing that same request.

Added to sso token exchange

Widcket (approved):

Issue #835, opened 2025-05-30
The SDK doesn't handle configuration changes correctly and causes a memory leak


francescocervone:
### Checklist

- [x] The issue can be reproduced in the [Auth0.Android sample app](https://github.com/auth0-samples/auth0-android-sample/tree/master/00-Login-Kt) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

If the user performs a [configuration change](https://developer.android.com/guide/topics/resources/runtime-changes) (e.g. rotate the device, change font or display size, change locale, enable / disable dark mode, ...) during the login flow, the Auth0 SDK doesn't deliver the login result correctly to the caller app.

Since the root cause of this issue is the same as https://github.com/auth0/Auth0.Android/issues/833, and since if you follow the steps below with the demo app on the `main` branch you get a crash (root cause explained below), you can reproduce this issue by using the demo app from the branch of this PR https://github.com/auth0/Auth0.Android/pull/832.

Basically, when a configuration change occurs, all the activities in the stack get recreated and the old instances thrown away.
So, the instance that started the login flow with the `WebAuthProvider.login` is destroyed. When the callback returns it notifies the destroyed activity.

Indeed this also actually leads to a **memory leak**, since the SDK is referencing a destroyed activity. 

This is also the reason why the demo on the `main` branch is crashing, because the `Fragment` that was detached from its activity after it was destroyed is receiving the callback and calls the `requireContext()`.

### Suggested solution

Also in this case, like https://github.com/auth0/Auth0.Android/issues/833, the suggested solution is to use the Android's [activity result API](https://developer.android.com/training/basics/intents/result).

### Reproduction

[...]

PR #846, opened 2025-07-15
Fix: Handle configuration changes correctly to prevent memory leak

utkrishtsahu:
### Changes

Please describe both what is changing and why this is important. Include:
In the SDK (WebAuthProvider.kt): modified the provider to manage a static list of active listeners instead of holding a direct reference to a single callback.
In the sample app (DatabaseLoginFragment.kt): updated the fragment to actively manage its listener status

### References

https://github.com/auth0/Auth0.Android/issues/835

### Testing
All existing Unit tests are passed

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.kt
@@ -29,21 +29,30 @@ public object WebAuthProvider {
     private val TAG: String? = WebAuthProvider::class.simpleName
     private const val KEY_BUNDLE_OAUTH_MANAGER_STATE = "oauth_manager_state"
 
-    private val callbacks = CopyOnWriteArraySet<Callback<Credentials, AuthenticationException>>()
+    private val loginCallbacks = CopyOnWriteArraySet<Callback<Credentials, AuthenticationException>>()
+    private val logoutCallbacks = CopyOnWriteArraySet<Callback<Void?, AuthenticationException>>()
 
     @JvmStatic
     @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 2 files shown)


auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.kt
```diff
@@ -511,7 +490,6 @@ public object WebAuthProvider {
             ctOptions = options
             return this
         }
-
```
pmathew92, line 514: Please don't remove the extra lines between methods

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.kt
```diff
@@ -112,27 +122,10 @@ public object WebAuthProvider {
     }
 
     internal fun onRestoreInstanceState(bundle: Bundle) {
-        if (managerInstance == null) {
```
pmathew92, line 115: Isn't this check required now?

pmathew92:
Code formatting has been distorted. Please fix them
Add tests for the new changes added

francescocervone:
Hi. 
I am the author of the mentioned issue.

How does this change fix the issue? 
In both the `start` and `await` cases, login and logout.

If the listener passed to the `start` method is coming from the caller Activity and a configuration change happens, we are basically storing a reference to the destroyed Activity in a global variable (because the `WebAuthProvider` is a Kotlin `object`).

So, when the result comes from the browser, it is dispatched to the destroyed Activity.

Could you please explain how this works?

Thanks!

pmathew92:
Hi @francescocervone , This PR was created with the intention of being a temporary fix for the memory leak issue, but as you mentioned it seems we won't be able to handle all the scenarios with this . Hence we will be closing this

PR #850, opened 2025-07-28
feat: Add support for DPoP

pmathew92:
### Changes

This PR implements DPoP (RFC 9449) support for Auth0.Android, enabling sender-constrained OAuth 2.0 access and refresh tokens using cryptographic proof of possession.

🔧 **API Changes**

- Added `DPoPProvider` object with static methods for key management and header generation
- Added `enableDPoP()` method to `WebAuthProvider` and `AuthenticationAPIClient` class
- New `DPoPException` type for DPoP-specific error handling
- The `userInfo` API in the AuthenticationAPIClient class now takes an `tokenType` parameter with default value `Bearer`

### References

[RFC9449](https://www.rfc-editor.org/rfc/rfc9449.html)

> [!NOTE]
DPoP is currently in Early Access. Contact Auth0 support to enable it for your tenant.

### Testing

### Checklist


auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -1017,6 +1073,13 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
             factory.post(url.toString(), credentialsAdapter), clientId, baseURL
         )
         request.addParameters(requestParameters)
+        try {
```
Widcket: Should we extract this logic into a helper to avoid copy-pasting the same block many times?

pmathew92: Done

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -561,9 +578,26 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
...
+                    accessToken,
+                    tokenType,
+                    DPoPProvider.auth0Nonce
+                )
+                addHeader(HEADER_AUTHORIZATION, headerData.authorizationHeader)
+                headerData.dpopProof?.let {
+                    addHeader(DPoPProvider.DPOP_HEADER, it)
+                }
+            } catch (exception: DPoPException) {
+                Log.e(TAG, "Error generating DPoP proof: ${exception.stackTraceToString()}")
```
Widcket: Could we return an `AuthenticationException` if this fails?

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -1046,6 +1085,20 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
...
+    /**
+     * Helper method to add DPoP proof to all the [Request]
+     */
+    private fun <T> Request<T, AuthenticationException>.addDPoPHeader(): Request<T, AuthenticationException> {
+        try {
+            DPoPProvider.generateProof(getUrl(), getHttpMethod().toString())?.let {
+                addHeader(DPoPProvider.DPOP_HEADER, it)
+            }
+        } catch (exception: DPoPException) {
+            Log.e(TAG, "Error generating DPoP proof: ${exception.stackTraceToString()}")
```
Widcket: Could we return an AuthenticationException if this fails?

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+            put("alg", "ES256")
+            put("jwk", createJWK(publicKey as ECPublicKey))
+        }
+        val headerEncoded = encodeBase64Url(headerJson.toString().toByteArray(Charsets.UTF_8))
+
+        //2. Construct the Payload
+        val payloadJson = JSONObject().apply {
+            put("jti", UUID.randomUUID().toString())
+            put("htm", httpMethod.uppercase())
+            put("htu", httpUrl)
```
Widcket: The URL must have any fragments and query params removed.

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+     * ```kotlin
+     *
+     *  try {
+     *      DPoPProvider.clearKeyPair()
+     *     } catch (exception: DPoPException) {
+     *          Log.e(TAG,"Error clearing  the key pair from the keystore: ${exception.stackTraceToString()}")
+     *     }
+     *
+     * ```
+     * **Note** : It is the developers responsibility to invoke this method to clear the keystore when logging out  a session.
```
Widcket:
```suggestion
     * **Note** : It is the developer's responsibility to invoke this method to clear the keystore when logging out.
```

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+
+        val signatureInput = "$headerEncoded.$payloadEncoded".toByteArray(Charsets.UTF_8)
+
+        //4. Sign the data
+        val signature = signData(signatureInput, privateKey)
+        return "$headerEncoded.$payloadEncoded.${signature}"
+    }
+
+    /**
+     * Method to clear the DPoP key pair from the keystore. It must be called when the user logs out from a session
```
Widcket:
```suggestion
     * Method to clear the DPoP key pair from the keystore. It must be called when the user logs out
```

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+        if (publicKey !is ECPublicKey) {
+            Log.e(TAG, "Key is not a ECPublicKey: ${publicKey.javaClass.name}")
+            return null
+        }
+        val jwkJson = createJWK(publicKey)
+        return createSHA256Hash(jwkJson.toString())
+    }
+
+    /**
+     * Generates a new key pair for DPoP if it does not already exist. This should be called before making any requests that require DPoP proof.
```
Widcket:
```suggestion
     * Generates a new key pair for DPoP if it does not already exist. This should be called before making any requests that require a DPoP proof.
```

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+    public fun generateKeyPair(context: Context) {
+        if (keyStore.hasKeyPair()) {
+            return
+        }
+        keyStore.generateKeyPair(context)
+    }
+
+    /**
+     * Generates the header data for a request that requires DPoP proof of possession. The `Authorization` header value is created
+     * using the access token and token type. The `DPoP` header value contains the generated DPoP proof
```
Widcket:
```suggestion
     * using the access token and token type. The `DPoP` header value contains the generated DPoP proof.
```

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+     *      Log.e(TAG, "Error storing nonce: ${exception.stackTraceToString()}")
+     *  }
+     *
+     * ```
+     *
+     * @param response The HTTP response containing the nonce header.
+     */
+    @JvmStatic
+    public fun storeNonce(response: Response) {
+        auth0Nonce = response.headers[NONCE_HEADER]
```
Widcket: Will this overwrite a stored nonce with `null` when the next response does not have a nonce?

Widcket: Also, could concurrent requests be problematic?

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+        var r = derSignature.copyOfRange(offset, offset + rLength)
+        offset += rLength
+
+        if (derSignature[offset++] != 0x02.toByte()) throw SignatureException("Invalid DER signature: Expected INTEGER for S")
+        val sLength = decodeLength(derSignature, offset).also { offset += it.second }.first
+        var s = derSignature.copyOfRange(offset, offset + sLength)
+        offset += sLength
+
+        // Remove leading zero if present
+        if (r.size > 1 && r[0] == 0x00.toByte() && r[1].toInt() and 0x80 == 0x80) r =
```
Widcket: Please address these warnings.

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+            r.copyOfRange(1, r.size)
+        if (s.size > 1 && s[0] == 0x00.toByte() && s[1].toInt() and 0x80 == 0x80) s =
+            s.copyOfRange(1, s.size)
+
+        // Pad with leading zeros to 32 bytes for P-256
```
Widcket: Is it necessary to remove the leading zero if we're going to add leading zeroes here?

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
```diff
@@ -0,0 +1,400 @@
...
+        val rawS = ByteArray(32)
+        System.arraycopy(s, 0, rawS, 32 - s.size, s.size)
+
+        return rawR + rawS
+    }
+
+    private fun decodeLength(data: ByteArray, offset: Int): Pair<Int, Int> {
+        var len = data[offset].toInt() and 0xFF
+        var bytesConsumed = 1
+        if (len and 0x80 != 0) {
```
Widcket: Please address this warning.

auth0/src/main/java/com/auth0/android/dpop/SenderConstraining.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import android.content.Context
+
+/**
+ * Interface for SenderConstraining
+ */
+public interface SenderConstraining<T> {
+
+    /**
+     * Enables DPoP for authentication requests.
```
Widcket: This text will no longer be relevant once DPoP support is added to the My Account API client and other API clients.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -790,8 +824,10 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
         val credentialsAdapter = GsonAdapter(
             Credentials::class.java, gson
         )
-        return factory.post(url.toString(), credentialsAdapter)
+        val request = factory.post(url.toString(), credentialsAdapter)
             .addParameters(parameters)
+            .addDPoPHeader()
```
Widcket: The DPoP proof should not be added for existing sessions. This seems to be adding it every time.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -927,7 +963,8 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
             Credentials::class.java, gson
         )
         val request = factory.post(url.toString(), credentialsAdapter)
-        request.addParameters(parameters)
+            .addParameters(parameters)
+            .addDPoPHeader()
         return request
```
Widcket, line 968: Shouldn't the method `ssoExchange()` also get the DPoP proof?

auth0/src/main/java/com/auth0/android/request/RetryInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+internal class RetryInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val retryCountHeader = request.header(RETRY_COUNT_HEADER)
+        val currentRetryCount = retryCountHeader?.toIntOrNull() ?: 0
+        val response = chain.proceed(request)
+
+        //Handling DPoP Nonce retry
+        if (DPoPProvider.isNonceRequiredError(response) && currentRetryCount < DPoPProvider.MAX_RETRY_COUNT) {
+            DPoPProvider.storeNonce(response)
```
Widcket: The nonce can also be sent ahead of time in a 200 OK response (https://www.rfc-editor.org/rfc/rfc9449.html#name-providing-a-new-nonce-value), for example, in the `/userinfo` response.

auth0/src/main/java/com/auth0/android/request/RetryInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+internal class RetryInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val retryCountHeader = request.header(RETRY_COUNT_HEADER)
+        val currentRetryCount = retryCountHeader?.toIntOrNull() ?: 0
+        val response = chain.proceed(request)
+
+        //Handling DPoP Nonce retry
+        if (DPoPProvider.isNonceRequiredError(response) && currentRetryCount < DPoPProvider.MAX_RETRY_COUNT) {
```
Widcket, line 18: Should this be using `currentRetryCount <= DPoPProvider.MAX_RETRY_COUNT`?

auth0/src/test/java/com/auth0/android/dpop/DPoPProviderTest.kt
```diff
@@ -0,0 +1,421 @@
...
+
+        val result = DPoPProvider.isNonceRequiredError(mockResponse)
+        assertThat(result, `is`(true))
+    }
+
+    @Test
+    public fun `isNonceRequiredError should return true for 401 response with resource server nonce error`() {
+        whenever(mockResponse.code).thenReturn(401)
+        whenever(mockResponse.headers).thenReturn(
+            Headers.Builder().add("WWW-Authenticate", "DPoP error=\"use_dpop_nonce\"").build()
```
Widcket: There should be another test case for when `use_dpop_nonce` is not quoted.

auth0/src/test/java/com/auth0/android/dpop/DPoPProviderTest.kt
```diff
@@ -0,0 +1,421 @@
...
+        whenever(mockResponse.code).thenReturn(401)
+
+        val result = DPoPProvider.isNonceRequiredError(mockResponse)
+        assertThat(result, `is`(false))
+    }
+
+    @Test
+    public fun `isNonceRequiredError should return false for 401 response with different WWW-Authenticate error`() {
+        whenever(mockResponse.headers).thenReturn(
+            Headers.Builder().add("WWW-Authenticate", "error=\"different_error\"").build()
```
Widcket: This seems to be missing the `DPoP` scheme.

auth0/src/test/java/com/auth0/android/dpop/DPoPProviderTest.kt
Widcket: Also, another test case with a `WWW-Authenticate` header containing both `Bearer` and `DPoP` challenges.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -306,6 +319,52 @@ public class WebAuthProviderTest {
         )
     }
 
+    //jwk
+
+    @Test
+    public fun enablingDPoPWillGenerateNEwKEyPairIfOneDoesNotExist() {
```
Widcket:
```suggestion
    public fun enablingDPoPWillGenerateNewKeyPairIfOneDoesNotExist() {
```

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -306,6 +319,52 @@ public class WebAuthProviderTest {
...
+        assertThat(
+            uri,
+            not(
+                UriMatchers.hasParamWithName("dpop_jkt")
+            )
+        )
+    }
+
+    @Test
+    public fun shouldNotHaveDpopJwkOnLoginIfDPoPIsEnabled() {
```
Widcket:
```suggestion
    public fun shouldHaveDpopJwkOnLoginIfDPoPIsEnabled() {
```

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
```diff
@@ -82,15 +82,25 @@ public class DefaultClientTest {
 
     @Test
     public fun shouldHaveLoggingDisabledByDefault() {
-        assertThat(DefaultClient().okHttpClient.interceptors, empty())
+        val netClient = DefaultClient(enableLogging = false)
+        assertThat(DefaultClient().okHttpClient.interceptors, hasSize(1))
+        val interceptor: Interceptor = netClient.okHttpClient.interceptors[0]
+        assert(
+            interceptor is RetryInterceptor,
+        )
```
Widcket, line 90: Should these assertions be in the test case below?

EXAMPLES.md
```diff
@@ -208,6 +210,76 @@ WebAuthProvider.login(account)
...
+                    accessToken, tokenType
+                )
+httpRequest.apply{
+    addHeader("Authorization", headerData.authorizationHeader)
+    headerData.dpopProof?.let {
+        addHeader("DPoP", it)
+    }
+}
+```
+If your API is issuing DPoP nonce's to prevent replay attacks, you can pass the nonce value to the `getHeaderData()` method to include it in the DPoP proof. Use the `DPoPProvider.isNonceRequiredError(response: Response)` method to check if a particular API response failed because a nonce is required.
```
Widcket:
```suggestion
If your API is issuing DPoP nonces to prevent replay attacks, you can pass the nonce value to the `getHeaderData()` method to include it in the DPoP proof. Use the `DPoPProvider.isNonceRequiredError(response: Response)` method to check if a particular API response failed because a nonce is required.
```

EXAMPLES.md
```diff
@@ -651,6 +723,62 @@ authentication
...
```
Widcket:
```suggestion
If your API is issuing DPoP nonces to prevent replay attacks, you can pass the nonce value to the `getHeaderData()` method to include it in the DPoP proof. Use the `DPoPProvider.isNonceRequiredError(response: Response)` method to check if a particular API response failed because a nonce is required.
```

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Let me check if this can be done without introducing much rework on the Request classes

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: The `addDPoPHeader` will only add it if there is an existing key pair, which will only be added when the useDPoP method is called as part of the login flow

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
pmathew92: Only the URL path is passed . No fragments or query params are passed

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
pmathew92: This is only called when the response has the `use_dpop_nonce` error. As per the protocol ,such errors should have the nonce value in the header

Widcket: Nonces can also come in 200 OK responses.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
Widcket: Do we have a test case for this?

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
Widcket: Can we have a test case for this?

auth0/src/test/java/com/auth0/android/dpop/DPoPProviderTest.kt
pmathew92: Yes that is the intention the check should fail for non DPoP error

Widcket: It could have the DPoP scheme, but have a different error code. Otherwise the `WWW-Authenticate` header is malformed, as it's missing the authentication scheme.

auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
pmathew92: Added

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Done

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Done

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
pmathew92: Added

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Done

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Done

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
pmathew92: Added `synchronized` within the nonce check to handle concurrent calss

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Added test case for removing the both fragments and query params from the url

auth0/src/main/java/com/auth0/android/dpop/DPoPProvider.kt
pmathew92: We first remove the DER encoding padding and replace with the raw signature 32 bytes padding

auth0/src/main/java/com/auth0/android/request/RetryInterceptor.kt
pmathew92: Added the required to change to handle this

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Added cases for the same

pmathew92:
> The latest changes (as per the replies to the PR feedback) have not yet been pushed.

@Widcket Have updated the PR with the latest changes

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,238 @@
...
+     *          Log.e(TAG,"Error generating key pair: ${exception.stackTraceToString()}")
+     *     }
+     *
+     * ```
+     *
+     * @param context The application context used to access the keystore.
+     * @throws DPoPException if there is an error generating the key pair or accessing the keystore.
+     */
+    @Throws(DPoPException::class)
+    public fun generateKeyPair(context: Context) {
```
Widcket: Is thus public method still needed?

pmathew92: Kept this method to give developers the option to generate key pair if needed.

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,238 @@
...
+     *          Log.e(TAG,"Error getting public key JWK: ${exception.stackTraceToString()}")
+     *     }
+     *
+     * ```
+     *
+     * @return The public key in JWK format or null if the key pair is not present.
+     * @throws DPoPException if there is an error accessing the key pair.
+     */
+    @Throws(DPoPException::class)
+    public fun getPublicKeyJWK(context: Context): String? {
```
Widcket: Does this method need to be public?

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,238 @@
...
+         *  } catch (exception: Exception) {
+         *      Log.e(TAG, "Error storing nonce: ${exception.stackTraceToString()}")
+         *  }
+         *
+         * ```
+         *
+         * @param response The HTTP response containing the nonce header.
+         */
+        @JvmStatic
+        public fun storeNonce(response: Response) {
```
Widcket: Does this method need to be public?

pmathew92: Can be used by developers to store the nonce value when making calls to their resource servers

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: Can be made internal

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
Widcket: Creating a key pair manually has implications in that it can affect the rollout of DPoP (all the checks that rely on the existence of a key pair will now return `true`). This should be documented in the method's API docs, or the method should be made internal.

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
Widcket: Unless the SDK stores the nonce in a map keyed by the audience, it will overwrite any Auth0 nonce currently stored, which will, in turn, be included in proofs to Auth0 endpoints, causing problems in concurrent scenarios.

Widcket: Which is why handling and storing nonces for non-Auth0 APIs is left to the developer, and why the SPA SDK will include a basic but general-purpose fetcher.

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: Understood. Will make this internal

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: Will make this internal

auth0/src/main/java/com/auth0/android/dpop/DPoPUtil.kt
```diff
@@ -0,0 +1,259 @@
...
+            put("kty", "EC")
+            put("x", encodeBase64Url(xBytes))
+            put("y", encodeBase64Url(yBytes))
+        }
+    }
+
+    private fun createSHA256Hash(input: String): String {
+        val digest = MessageDigest.getInstance("SHA-256")
+        val hash = digest.digest(input.toByteArray(Charsets.UTF_8))
+        return Base64.encodeToString(hash, Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP)
```
Widcket: Maybe use `encodeBase64Url()` here?

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,238 @@
...
+         * @param httpMethod Method type of the request
+         * @param httpUrl Url of the request
+         * @param accessToken Access token to be included in the `Authorization` header
+         * @param tokenType Either `DPoP` or `Bearer`
+         * @param nonce Optional nonce value to be used in the proof
+         * @throws DPoPException if there is an error generating the DPoP proof or accessing the key pair
+         */
+        @Throws(DPoPException::class)
+        @JvmStatic
+        public fun getHeaderData(
```
Widcket, line 188: Will this method end up creating a new key pair if one does not exist?

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,238 @@
...
+     *     }
+     *
+     * ```
+     *
+     * @return The public key in JWK format or null if the key pair is not present.
+     * @throws DPoPException if there is an error accessing the key pair.
+     */
+    @Throws(DPoPException::class)
+    public fun getPublicKeyJWK(context: Context): String? {
+        generateKeyPair(context)
```
Widcket: This method will create a new key pair if one does not yet exist when using Web Auth. But if the developer is using embedded login (that is, just the Authentication API client), will a new key pair be created if one does not yet exist?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+        ShadowLooper.idleMainLooper()
+        verify(authCallback).onSuccess(credentialsCaptor.capture())
+        val credentials = credentialsCaptor.firstValue
+        assertThat(credentials, `is`(notNullValue()))
+        assertThat(credentials.type, `is`("DPoP"))
+        mockAPI.shutdown()
+    }
+
+    // Update the existing test that checks DPoP key generation
+//    @Test
```
Widcket: There is commented out code here.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+//        assertThat(uri, UriMatchers.hasParamWithValue("dpop_jkt", "KQ-r0YQMCm0yVnGippcsZK4zO7oGIjOkNRbvILjjBAo"))
+//    }
+
+    @Test
+    public fun shouldResetDPoPStateWhenManagerInstanceIsReset() {
+        WebAuthProvider.useDPoP()
+
+        // Verify DPoP is enabled
+        val provider1 = WebAuthProvider.useDPoP()
+        assertThat(provider1, `is`(notNullValue()))
```
Widcket: The assertions in this test are just checking that the `WebAuthProvider` instance returned by `useDPoP()` is not null. Is this intended?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+        `when`(mockKeyStore.hasKeyPair()).thenReturn(true)
+        `when`(mockKeyStore.getKeyPair()).thenReturn(Pair(mock(), FakeECPublicKey()))
+
+        WebAuthProvider.useDPoP()
+            .login(account)
+            .start(activity, callback)
+
+        val managerInstance = WebAuthProvider.managerInstance as OAuthManager
+        // Verify that the manager has DPoP configured (this would require exposing the dPoP field in OAuthManager for testing)
+        assertThat(managerInstance, `is`(notNullValue()))
```
Widcket: This seems to be missing the assert over the dPoP field in OAuthManager.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+    }
+
+    @Test
+    public fun shouldNotPassDPoPInstanceToOAuthManagerWhenDPoPIsNotEnabled() {
+        login(account)
+            .start(activity, callback)
+
+        val managerInstance = WebAuthProvider.managerInstance as OAuthManager
+        assertThat(managerInstance, `is`(notNullValue()))
```
Widcket: This seems to be missing the assert over the dPoP field in OAuthManager.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+        assertThat(uri, `is`(notNullValue()))
+        assertThat(uri, UriMatchers.hasParamWithName("dpop_jkt"))
+        assertThat(
+            uri,
+            UriMatchers.hasParamWithValue("dpop_jkt", "KQ-r0YQMCm0yVnGippcsZK4zO7oGIjOkNRbvILjjBAo")
+        )
+    }
+
+    @Test
+    public fun shouldNotIncludeDPoPJWKThumbprintWhenDPoPIsEnabledButNoKeyPair() {
```
Widcket: Isn't the method that creates the `dpop_jkt` calling `generateKeyPair(context)`?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2724,6 +2727,299 @@ public class WebAuthProviderTest {
...
+            intentCaptor.firstValue.getParcelableExtra<Uri>(AuthenticationActivity.EXTRA_AUTHORIZE_URI)
+        assertThat(uri, `is`(notNullValue()))
+        assertThat(uri, UriMatchers.hasParamWithValue("connection", "test-connection"))
+        assertThat(uri, UriMatchers.hasParamWithValue("scope", "openid profile email custom"))
+        assertThat(uri, UriMatchers.hasParamWithValue("state", "custom-state"))
+        assertThat(uri, UriMatchers.hasParamWithName("dpop_jkt"))
+    }
+
+    @Test
+    public fun shouldWorkWithLoginBuilderPatternWhenDPoPIsEnabled() {
```
Widcket, line 2835: Isn't this testing the same thing as the previous test case?

auth0/src/main/java/com/auth0/android/dpop/DPoPUtil.kt
```diff
@@ -0,0 +1,259 @@
...
+                uri.host,
+                uri.port,
+                uri.path,
+                null, // Remove query
+                null  // Remove fragment
+            )
+            cleanedUri.toString()
+        } catch (e: Exception) {
+            Log.w(TAG, "Failed to parse URL, using original: $url", e)
+            url
```
Widcket: Should this method just throw? After all, the server will reject a DPoP proof with an `htu` that does not match the request's URL.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -2772,6 +2738,283 @@ public class AuthenticationAPIClientTest {
...
+        assertThat(request.path, Matchers.equalTo("/oauth/token"))
+        assertThat(
+            callback, AuthenticationCallbackMatcher.hasPayloadOfType(
+                Credentials::class.java
+            )
+        )
+    }
+
+    @Test
+    public fun shouldAddDpopHeaderToTokenExchangeWhenEnabled() {
```
Widcket: A test case is missing for when DPoP is enabled but no key pair exists (when making an RTE request). This is the case when DPoP is being rolled out, and the user hasn't logged in again yet.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -2772,6 +2738,283 @@ public class AuthenticationAPIClientTest {
...
+        assertThat(request.path, Matchers.equalTo("/oauth/token"))
+        assertThat(
+            callback, AuthenticationCallbackMatcher.hasPayloadOfType(
+                Credentials::class.java
+            )
+        )
+    }
+
+    @Test
+    public fun shouldNotAddDpopHeaderWhenDpopEnabledButNoKeyPair() {
```
Widcket: For the login case (that is, requests to /token when the `grant_type` is not `refresh_token`), it should create a new key pair if one does not already exist.

Widcket: It should not create a key pair when the `grant_type` is `refresh_token`.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
Widcket: In this case (when the `grant_type` is `refresh_token`), a key pair should not be created.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -2772,6 +2738,283 @@ public class AuthenticationAPIClientTest {
...
+        assertThat(request.path, Matchers.equalTo("/oauth/token"))
+        assertThat(
+            callback, AuthenticationCallbackMatcher.hasPayloadOfType(
+                SSOCredentials::class.java
+            )
+        )
+    }
+
+    @Test
+    public fun shouldNotAddDpopHeaderWhenKeyPairGenerationFails() {
```
Widcket: This should be called `shouldNotAddDpopHeaderWhenKeyPairRetrievalFails()`.

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -2772,6 +2738,283 @@ public class AuthenticationAPIClientTest {
...
+
+        mockAPI.willReturnSuccessfulLogin()
+        val callback = MockAuthenticationCallback<Credentials>()
+
+        client.useDPoP().login(SUPPORT_AUTH0_COM, "some-password", MY_CONNECTION)
+            .start(callback)
+        ShadowLooper.idleMainLooper()
+
+        val request = mockAPI.takeRequest()
+        // Should not have DPoP header when key pair retrieval fails
```
Widcket: Shouldn't this result in an error being thrown?

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
```diff
@@ -2772,6 +2738,283 @@ public class AuthenticationAPIClientTest {
...
+        assertThat(request.path, Matchers.equalTo("/userinfo"))
+        assertThat(
+            callback, AuthenticationCallbackMatcher.hasPayloadOfType(
+                UserProfile::class.java
+            )
+        )
+    }
+
+    @Test
+    public fun shouldNotAddDpopHeaderToNonTokenEndpoints() {
```
Widcket: This should be called `shouldNotAddDpopHeaderToSignupEndpoint()`.

auth0/src/test/java/com/auth0/android/dpop/DPoPTest.kt
```diff
@@ -0,0 +1,518 @@
...
+                    )
+                    DPoP.storeNonce(response)
+                }
+            })
+        }
+
+        threads.forEach { it.start() }
+        threads.forEach { it.join() }
+
+        // Should not crash and should have some nonce value
```
Widcket, line 473: The problem with non-thread-safe nonce storage is that one could end up with an old (and invalid) nonce stored, while the newer one gets overwritten.

auth0/src/test/java/com/auth0/android/dpop/DPoPTest.kt
```diff
@@ -0,0 +1,518 @@
...
+
+    @Test
+    public fun `storeNonce should handle missing DPoP-Nonce header`() {
+        whenever(mockResponse.headers).thenReturn(Headers.Builder().build())
+        DPoP.storeNonce(mockResponse)
+        assertThat(DPoP.auth0Nonce, `is`(nullValue()))
+    }
+
+    @Test
+    public fun `storeNonce should overwrite existing nonce`() {
```
Widcket, line 243: Missing test case when the second response does not include a nonce, and the stored nonce (from the first response) is not cleared.

auth0/src/test/java/com/auth0/android/dpop/DPoPTest.kt
```diff
@@ -0,0 +1,518 @@
...
+        whenever(mockKeyStore.hasKeyPair()).thenReturn(true)
+
+        val parameters = mapOf("grant_type" to "authorization_code")
+        val result = dPoP.shouldGenerateProof(testHttpUrl, parameters)
+
+        assertThat(result, `is`(true))
+    }
+
+    @Test
+    public fun `shouldGenerateProof should return false for token endpoint with refresh grant type`() {
```
Widcket: For the /token endpoint with `grant_type` of `refresh_token`, it should return the result of calling `hasKeyPair()`. That is, for RTE requests, it should depend on whether or not a key pair exists.

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -0,0 +1,216 @@
...
+         *  } catch (exception: Exception) {
+         *      Log.e(TAG, "Error storing nonce: ${exception.stackTraceToString()}")
+         *  }
+         *
+         * ```
+         *
+         * @param response The HTTP response containing the nonce header.
+         */
+        @JvmStatic
+        internal fun storeNonce(response: Response) {
```
Widcket, line 126: The nonce storage operation does not seem to be synchronized (serialized).

Widcket: Also, it should handle the case when `response.headers[NONCE_HEADER]` is null, to avoid clearing out the nonce currently stored.

Widcket:
The test cases that assert that an error is thrown do not check that its `cause` property is not null (where applicable; when it should be null, there should be an assertion for this).

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: No. This method will only check if an existing key pair exist or not

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: No, currently generateKeyPair is called only from the  getPublicKeyJWK(context) method which is being called as part of the WebAuth flow. So using the Embedded login wouldn't  generate a key pair . Isn't that how the flow is supposed to happen ?

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: This case is not needed as it doesn't validate anything meaningful

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
Widcket: > So using the Embedded login wouldn't generate a key pair .

DPoP is to be supported in all login cases. There is nothing in the requirements doc that says that DPoP is just for Web Auth. That's why the `shouldGenerateProof()` on Auth0.swift checks the `grant_type`, so that login calls always result in a key pair being created if one doesn't already exist.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: yes, that would create the `generateKeyPair(context)`.  This case was handling the scenario if keypair fetching returns null even if key pair exists. Have updated the title to convey the same

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Removed redundant case

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
Widcket: That is because on the Auth0.swift implementation, the key fetcher also creates a key pair if one does not already exist.

auth0/src/main/java/com/auth0/android/dpop/DPoPUtil.kt
pmathew92: Throwing an exception

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: Done

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Removed

auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
pmathew92: Added code to create a key-pair in embedded login flow if the key-pair doesn't exist

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Added the dPoP assertion in this

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Added

auth0/src/test/java/com/auth0/android/dpop/DPoPTest.kt
pmathew92: Done

auth0/src/test/java/com/auth0/android/dpop/DPoPTest.kt
pmathew92: Done

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
pmathew92: Done

auth0/src/test/java/com/auth0/android/authentication/AuthenticationAPIClientTest.kt
pmathew92: Done

Widcket (approved):

PR #885, opened 2025-12-09
feat: Add support for `organization` to custom token exchange

pmathew92:
### Changes

This PR adds support for organization support in `customTokenExchange` in the `AuthenticationAPIClient` class 

### Testing

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,4 +29,4 @@ jobs:
 
       - run: ./gradlew clean test jacocoTestReport lint --continue --console=plain --max-workers=1 --no-daemon
 
-      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # pin@5.5.1
+      - uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # pin@5.5.2


auth0/src/main/java/com/auth0/android/request/internal/validator/CustomTokenExchangeValidator.kt
```diff
@@ -0,0 +1,52 @@
...
+        "https://auth0.com",
+        "http://okta.com",
+        "https://okta.com",
+        "urn:ietf",
+        "urn:auth0",
+        "urn:okta"
+    )
+
+    override fun validate(options: RequestOptions) {
+        val subjectTokenType = options.parameters["subject_token_type"] as String
```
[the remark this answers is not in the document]
pmathew92: The `subject_token_type` will always be a string

auth0/src/main/java/com/auth0/android/request/internal/validator/CustomTokenExchangeValidator.kt
```diff
@@ -0,0 +1,60 @@
...
+        if(subjectTokenType.isEmpty()){
+            throw AuthenticationException(
+                "Invalid URI", IllegalArgumentException(
+                    "The passed URI must not be an empty String"
+                )
+            )
+        }
+
+        // Check if it's a reserved namespace
+        if (reservedNamespaces.contains(subjectTokenType)) {
```
sanchitmehtagit: https://auth0.com/x should also be considered invalid
instead we can add startWith check here

auth0/src/main/java/com/auth0/android/request/internal/validator/CustomTokenExchangeValidator.kt
```diff
@@ -0,0 +1,60 @@
...
+                    "The passed URI is a reserved namespace and cannot be used"
+                )
+            )
+        }
+
+        // If it starts with http:// or https://, validate it as a URI
+        if (subjectTokenType.startsWith(
+                "http://",
+                ignoreCase = true
+            ) || subjectTokenType.startsWith("https://", ignoreCase = true)
```
sanchitmehtagit: subjectTokenType.startsWith("urn:", ignoreCase = true)
should we validate urn too?

auth0/src/main/java/com/auth0/android/request/internal/validator/CustomTokenExchangeValidator.kt
pmathew92: subjectTokenType is not actual URI link but something the user defines themselves
we have list of reserved keyword thats defined in 
https://auth0.com/docs/authenticate/custom-token-exchange
But let me check this and get back on the same. Thanks for pointing this out

auth0/src/main/java/com/auth0/android/request/internal/validator/CustomTokenExchangeValidator.kt
pmathew92: urn is not required as it can fall under user custom schemes and those are allowed

sanchitmehtagit: Thank you! I was thinking more in terms of validating a URN. for example, if we provide URI("urn:ietf: params:oauth"), this API should be able to check whether it’s a valid URI.

sanchitmehta94 (approved):
LGTM

PR #891, opened 2026-01-06
fix: Fixing the memory leak introduced in the DPoP flow

pmathew92:
### Changes
This Pr fixes the memory leak introduced as part of the DPoP flow. The `dpop` property in the `WebAuthProvider` class will retain the instance of the activity/fragment context preventing it from being garbage collected. Updated the `context` in the DPoP class to use an `activityContext` instead.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
+++ b/auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
@@ -22,7 +22,9 @@ public data class HeaderData(val authorizationHeader: String, val dpopProof: Str
  * Class for securing requests with DPoP (Demonstrating Proof of Possession) as described in
  * [RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449).
  */
-public class DPoP(private val context: Context) {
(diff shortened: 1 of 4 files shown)


auth0/src/main/java/com/auth0/android/dpop/DPoP.kt
```diff
@@ -22,7 +22,9 @@ public data class HeaderData(val authorizationHeader: String, val dpopProof: Str
  * Class for securing requests with DPoP (Demonstrating Proof of Possession) as described in
  * [RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449).
  */
-public class DPoP(private val context: Context) {
+public class DPoP(context: Context) {
+
+    private val applicationContext: Context = context.applicationContext
```
sanchitmehtagit, line 28: For my understanding . holding activity context would cause issue DPoP instance lives longer than activity and causing activity to not GC?

pmathew92: Yes

PR #902, opened 2026-02-06
fix: Prevent DPoP replay protection error due to OkHttp retry

pmathew92:
## Description

This PR introduces a dual-client architecture in `DefaultClient` to prevent OkHttp from automatically retrying network requests that use DPoP (Demonstrating Proof-of-Possession). DPoP proofs contain unique, non-replayable tokens (`jti` claims and timestamps), and OkHttp's automatic retry mechanism can cause "proof already used" errors when the same proof is sent twice.

### Core Implementation (`DefaultClient.kt`)
- Added `nonRetryableOkHttpClient` property with `retryOnConnectionFailure = false`

## Testing
All tests pass successfully
```bash
./gradlew auth0:testDebugUnitTest
```

### DIFF
--- a/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
+++ b/auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
@@ -103,7 +103,7 @@ public class DefaultClient @VisibleForTesting(otherwise = VisibleForTesting.PRIV
             .headers(headers)
             .build()
 
-        // Use non-retryable client for DPoP requests or token refresh requests
+        // Use non-retryable client for DPoP requests
         val client = if (shouldUseNonRetryableClient(headers)) {
             nonRetryableOkHttpClient
         } else {


auth0/src/test/java/com/auth0/android/request/DefaultClientTest.kt
```diff
@@ -230,6 +230,43 @@ public class DefaultClientTest {
...
+        )
+        assertThat(
+            client.okHttpClient.connectTimeoutMillis,
+            equalTo(client.nonRetryableOkHttpClient.connectTimeoutMillis)
+        )
+        assertThat(
+            client.okHttpClient.readTimeoutMillis,
+            equalTo(client.nonRetryableOkHttpClient.readTimeoutMillis)
+        )
+    }
```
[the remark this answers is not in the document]
pmathew92: Ignore

auth0/src/main/java/com/auth0/android/request/DefaultClient.kt
```diff
@@ -90,12 +102,31 @@ public class DefaultClient @VisibleForTesting(otherwise = VisibleForTesting.PRIV
...
+            okHttpClient
+        }
+
+        return client.newCall(request)
+    }
+
+    /**
+     * Determines if the request should use the non-retryable OkHttpClient.
+     * Returns true for:
+     * 1. Requests with DPoP header
```
NandanPrabhu, line 119: do we have other  conditions too for which non retryable client must be used?

pmathew92: Currently Dpop scenario is where this is causing major error. We can update these cases on need to basis like refresh token exchange etc

PR #949, opened 2026-04-06
feat: Add three new DPoP errors to `CredentialsManagerException`

pmathew92:
### Changes

Added 3 new error type to the `CredentialsManagerException` class when DPoP is used:

- DPOP_KEY_MISSING : Thrown when the stored credentials are Dpop bound but the key-pair is no longer available in KeyStore
- DPOP_NOT_CONFIGURED : Thrown when credentials are Dpop bound but the AuthenticationAPIClient used by the CredentialsManager was not configured to use DpoP
- DPOP_KEY_MISMATCH : Thrown when the key pair used while creating the token and refresh token is different from the key pair used while trying to refresh the token.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -195,6 +195,63 @@ public abstract class BaseCredentialsManager internal constructor(
     }
 
     /**
+     * Validates DPoP key/token alignment before attempting a refresh.
+     *
+     * Uses two signals to detect DPoP-bound credentials:
+     * - tokenType == "DPoP"
+     * - KEY_DPOP_THUMBPRINT exists
+     *
+     * @param tokenType the token_type value from storage (or decrypted credentials for migration)
+     * @return null if validation passes, or a CredentialsManagerException if it fails
+     */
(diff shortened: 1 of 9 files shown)


NandanPrabhu (approved):

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -155,6 +165,92 @@ public abstract class BaseCredentialsManager internal constructor(
...
+        // Check 1: Does the DPoP key still exist in KeyStore?
+        val hasKey = try {
+            DPoPUtil.hasKeyPair()
+        } catch (e: DPoPException) {
+            Log.e(this::class.java.simpleName, "Failed to check DPoP key existence", e)
+            false
+        }
+        if (!hasKey) {
+            Log.w(this::class.java.simpleName, "DPoP key missing from KeyStore. Clearing stale credentials.")
+            clearCredentials()
```
NandanPrabhu, line 222: @pmathew92 do we need clearCredentials() here or expect the client to clear it on returning DPOP_KEY_MISSING exception

pmathew92: Since the key is missing and the the token is dpop bound, user has to be re-authenticated . So clearing the stored credentials here to avoid any stale data

PR #951, opened 2026-04-15
feat: Add clearAll() API to credentials manager

pmathew92:
### Changes

Adds a new `clearAll()` method to `BaseCredentialsManager` , `CredentialsManager`, and `SecureCredentialsManager` that performs a complete cleanup of all stored credentials and cryptographic key pairs.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
@@ -10,6 +10,8 @@ import com.auth0.android.Auth0
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.authentication.AuthenticationException
 import com.auth0.android.callback.Callback
+import com.auth0.android.dpop.DPoP
+import com.auth0.android.dpop.DPoPException
 import com.auth0.android.request.internal.GsonProvider
 import com.auth0.android.result.APICredentials
 import com.auth0.android.result.Credentials
@@ -737,6 +739,22 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
     }
 
     /**
+     * Removes all credentials, API credentials, and cryptographic key pairs.
+     * This calls [Storage.removeAll] to clear all stored data
+     */
+    override fun clearAll() {
+        storage.removeAll()
+        crypto.deleteAllKeys()
+        clearBiometricSession()
+        try {
+            DPoP.clearKeyPair()
+        } catch (e: DPoPException) {
+            Log.e(TAG, "Failed to clear DPoP key pair ${e.stackTraceToString()}")
+        }
+        Log.d(TAG, "All credentials and key pairs were removed")
+    }
+
+    /**
      * Removes the credentials for the given audience from the storage if present.
      * @param audience Audience for which the [APICredentials] are stored
      * @param scope Optional scope for which the [APICredentials] are stored. If the credentials were initially fetched/stored with scope,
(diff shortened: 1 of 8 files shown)


NandanPrabhu (dismissed):

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
utkrishtsahu, line 1: the test name says "AndDPoPKeyPair" but we're only verifying storage.removeAll() here. If someone accidentally removes the DPoP.clearKeyPair() call from the implementation, this test would still pass. Can we add a verify for the DPoP cleanup too?

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2164,6 +2164,13 @@ public class SecureCredentialsManagerTest {
         verifyNoMoreInteractions(storage)
     }
 
+    @Test
+    public fun shouldClearAllCredentialsKeyPairsAndDPoPKeyPair() {
```
utkrishtsahu: the test name mentions "AndDPoPKeyPair" but we're not verifying DPoP.clearKeyPair() was called. Also missing a verify for clearBiometricSession(). Would be good to cover both so the test actually validates everything the method does.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -736,6 +738,22 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
         Log.d(TAG, "Credentials were just removed from the storage")
     }
 
+    /**
+     * Removes all credentials, API credentials, and cryptographic key pairs.
+     * This calls [Storage.removeAll] to clear all stored data
+     */
+    override fun clearAll() {
+        storage.removeAll()
+        crypto.deleteAllKeys()
```
utkrishtsahu, line 747: Small thought — if crypto.deleteAllKeys() throws something unexpected, then clearBiometricSession() and DPoP.clearKeyPair() get skipped. Since this is a "clean up everything" method,can we wrap each step independently so one failure doesn't prevent the others from running?

utkrishtsahu (changes requested):

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
pmathew92: CryptoUtils , deleteAllKeys internally calls the existing `deleteRSAKeys` and `deleteAESKeys` method which internally handles if any exception being thrown. So any failure here wouldn't affect the other APIs

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
pmathew92: The name is misleading. We can't directly verify the `DPoP.clearKeyPair() ` as Mockito can't validate the call for Kotlin static calls. Hence didn't add that check

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
pmathew92: The name is misleading. We can't directly verify the DPoP.clearKeyPair()  as Mockito can't validate the call for Kotlin static calls. Hence didn't add that check

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
utkrishtsahu: agree

utkrishtsahu (approved):

PR #954, opened 2026-04-16
feat : Add Partial Custom Tabs support (bottom sheet & side sheet)

subhankarmaiti:
Adds support for [Partial Custom Tabs](https://developer.chrome.com/docs/android/custom-tabs/guide-partial-custom-tabs), allowing the browser-based authentication flow to render as a bottom sheet or side sheet instead of taking over the full screen. Requires Chrome 107+; gracefully falls back to full-screen Custom Tabs on unsupported browsers.

## API Changes

Six new methods on `CustomTabsOptions.Builder`:

| Method | Description |
|---|---|
| `withInitialHeight(int height)` | Sets the bottom sheet height in **dp**. Minimum enforced by Chrome is 50% of screen. |
| `withInitialWidth(int width)` | Sets the side sheet width in **dp** (for wider screens). |
| `withResizable(boolean resizable)` | Locks or unlocks user-resizing of the sheet. Default: resizable. |
| `withToolbarCornerRadius(int cornerRadius)` | Sets toolbar top corner radius in **dp**. Only applies in bottom sheet mode. |
| `withSideSheetBreakpoint(int breakpoint)` | Breakpoint in **dp** to toggle between bottom sheet and side sheet. Browser default is typically 840dp. |
| `withBackgroundInteractionEnabled(boolean enabled)` | Allows tapping the background app while the partial tab is visible. Default: disabled. |

All dimension parameters accept **dp** values. Height and width are converted to pixels internally via `dpToPx()`.

### Usage — Bottom Sheet

```java
CustomTabsOptions options = CustomTabsOptions.newBuilder()
    .withInitialHeight(500)
    .withToolbarCornerRadius(16)
    .withResizable(false)
    .build();

WebAuthProvider.login(account)
    .withCustomTabsOptions(options)
    .start(context, callback);
```

### Usage — Side Sheet on larger screens

```java
CustomTabsOptions options = CustomTabsOptions.newBuilder()
    .withInitialHeight(500)
    .withInitialWidth(400)
    .withSideSheetBreakpoint(800)
    .withToolbarCornerRadius(10)
    .withResizable(false)
    .build();

WebAuthProvider.login(account)
    .withCustomTabsOptions(options)
    .start(context, callback);
```

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
@@ -9,6 +9,7 @@ import android.os.Parcel;
 import android.os.Parcelable;
 
 import androidx.annotation.ColorRes;
+import androidx.annotation.Dimension;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.browser.customtabs.CustomTabColorSchemeParams;
@@ -20,6 +21,8 @@ import androidx.core.content.ContextCompat;
 import com.auth0.android.authentication.AuthenticationException;
 
 import java.util.List;
+
+import android.util.DisplayMetrics;
 
 /**
  * Holder for Custom Tabs customization options. Use {@link CustomTabsOptions#newBuilder()} to begin.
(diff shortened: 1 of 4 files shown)


auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
```diff
@@ -34,11 +37,38 @@ public class CustomTabsOptions implements Parcelable {
...
         this.disabledCustomTabsPackages = disabledCustomTabsPackages;
+        this.initialHeight = initialHeight;
+        this.activityHeightResizeBehavior = activityHeightResizeBehavior;
+        this.toolbarCornerRadius = toolbarCornerRadius;
+        this.initialWidth = initialWidth;
+        this.sideSheetBreakpoint = sideSheetBreakpoint;
+        this.backgroundInteractionEnabled = backgroundInteractionEnabled;
+    }
+
+    private static int dpToPx(@NonNull Context context, int dp) {
```
pmathew92: Why is this private method static ?  Also lets move this below the public methods

auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
```diff
@@ -147,11 +211,24 @@ public static class Builder {
...
         Builder() {
             this.showTitle = false;
             this.toolbarColor = 0;
             this.browserPicker = BrowserPicker.newBuilder().build();
             this.disabledCustomTabsPackages = null;
+            this.initialHeight = 0;
+            this.activityHeightResizeBehavior = CustomTabsIntent.ACTIVITY_HEIGHT_DEFAULT;
+            this.toolbarCornerRadius = 0;
+            this.initialWidth = 0;
+            this.sideSheetBreakpoint = 0;
```
pmathew92, line 224: if `sideSheetBreakpoint` is 0 by default, wouldn't this cause all CCTs to be opened as side sheet including on small width devices ?

auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
```diff
@@ -88,6 +118,28 @@ Intent toIntent(@NonNull Context context, @Nullable CustomTabsSession session) {
                     .setToolbarColor(ContextCompat.getColor(context, toolbarColor));
             builder.setDefaultColorSchemeParams(colorBuilder.build());
         }
+
+        // Partial Custom Tabs - Bottom Sheet
```
pmathew92, line 116: Lets check if we need to have an upper limit to these properties

subhankarmaiti: as per https://developer.chrome.com/docs/android/custom-tabs/guide-partial-custom-tabs chrome/androidX automatically clamp the height from 50% to max height, No need any checks in sdk side

auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
subhankarmaiti: When `sideSheetBreakpoint` is 0 (the default), we are not calling `setActivitySideSheetBreakpointDp`. Updated the doc

auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
```diff
@@ -212,15 +283,126 @@ public Builder withDisabledCustomTabsPackages(List<String> disabledCustomTabsPac
...
+         * Pass the size in dp. The underlying
+         * {@link CustomTabsIntent.Builder#setToolbarCornerRadiusDp(int)} currently accepts
+         * values in the range {@code 0}–{@code 16} (inclusive) and will throw an
+         * {@link IllegalArgumentException} for values outside that range.
+         *
+         * @param cornerRadius the toolbar's top corner radius in dp.
+         * @return this same builder instance.
+         */
+        @NonNull
+        public Builder withToolbarCornerRadius(@Dimension(unit = Dimension.DP) int cornerRadius) {
```
pmathew92, line 336: Since this throws an exception when a value above that 16, which will crash  should we limit the max value to 16 . We should also handle case for values less than 0
```
if (cornerRadius < 0 ) {                                                                                                                                                                                
   cornerRadius   = 0                                                                                                               
}    
if( cornerRadius > 16) {
   cornerRadius   = 16
}

this.toolbarCornerRadius = cornerRadius;
 return this;
```

pmathew92 (approved):

PR #961, opened 2026-04-27
breaking : Remove redundant constructors from SecureCredentialsManager

pmathew92:
### Changes
This PR removes two redundant constructors from `SecureCredentialsManager` class making it less confusing for users while instantiating the class .  
`AuthenticationAPIClient` already holds a reference to the Auth0 object, so passing both was always duplication. 
Removing `Auth0` also makes advanced client configuration like DPoP a natural part of setup: callers configure the client before passing it in, and the manager inherits that configuration automatically.

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
@@ -5,7 +5,6 @@ import android.util.Log
 import androidx.annotation.VisibleForTesting
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.authentication.AuthenticationException
-import com.auth0.android.authentication.storage.BaseCredentialsManager.Companion.DEFAULT_MIN_TTL
 import com.auth0.android.callback.Callback
 import com.auth0.android.dpop.DPoP
 import com.auth0.android.dpop.DPoPException
@@ -21,7 +20,6 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 import java.util.Date
 import java.util.Locale
 import java.util.concurrent.Executor
-import java.util.concurrent.Executors
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 
@@ -47,7 +45,7 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         authenticationClient,
         storage,
         JWTDecoder(),
-        Executors.newSingleThreadExecutor()
+        authenticationClient.executor
     )
 
     public override val userProfile: UserProfile?
--- a/auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
+++ b/auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
@@ -5,7 +5,6 @@ import android.app.KeyguardManager
 import android.content.Context
 import android.util.Base64
 import androidx.fragment.app.FragmentActivity
-import com.auth0.android.Auth0
 import com.auth0.android.NetworkErrorException
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.authentication.AuthenticationException
@@ -51,6 +50,7 @@ import org.mockito.Mockito
 import org.mockito.MockitoAnnotations
 import org.mockito.kotlin.KArgumentCaptor
 import org.mockito.kotlin.any
+import org.mockito.kotlin.anyOrNull
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.eq
 import org.mockito.kotlin.mock
@@ -128,7 +128,6 @@ public class SecureCredentialsManagerTest {
     public val exception: ExpectedException = ExpectedException.none()
     private lateinit var manager: SecureCredentialsManager
     private lateinit var gson: Gson
-    private lateinit var auth0: Auth0
 
     @Before
     public fun setUp() {
@@ -154,8 +153,6 @@ public class SecureCredentialsManagerTest {
                     .get()
             )
         weakFragmentActivity = WeakReference(fragmentActivity)
-        auth0 = Mockito.spy(Auth0.getInstance("clientId", "domain"))
-        Mockito.`when`(auth0.executor).thenReturn(serialExecutor)
         Mockito.`when`(client.executor).thenReturn(serialExecutor)
 
         val secureCredentialsManager =
@@ -164,7 +161,7 @@ public class SecureCredentialsManagerTest {
                 storage,
                 crypto,
                 jwtDecoder,
-                auth0.executor,
+                serialExecutor,
                 weakFragmentActivity,
                 getAuthenticationOptions(),
                 factory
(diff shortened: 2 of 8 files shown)


utkrishtsahu (commented):
## PR Review: Remove redundant constructors from SecureCredentialsManager

### Summary

This PR removes the two `Auth0`-based constructors from `SecureCredentialsManager`, leaving only the `AuthenticationAPIClient`-based ones. The motivation is sound — `AuthenticationAPIClient` already holds an `Auth0` reference, so accepting both was pure duplication, and this change makes DPoP configuration via `useDPoP()` a natural part of setup flow.

The documentation updates (migration guide, examples) are thorough and the new concurrency tests are a welcome addition.

---

### Issues

#### 1. Bundled behavioral change in `CredentialsManager` (medium concern)

The `CredentialsManager` public constructor (`CredentialsManager.kt` line ~45) was changed from:

```kotlin
Executors.newSingleThreadExecutor()  // each instance gets its own executor
```

to:

```kotlin
authenticationClient.executor  // shared executor from Auth0 instance
```

This is a **separate behavioral change** from the stated constructor removal. Previously, each `CredentialsManager` instance had an independent executor. Now, all `CredentialsManager` and `SecureCredentialsManager` instances sharing the same `Auth0` object also share a single-thread executor, **serializing all credential operations globally**.

While the new test `shouldSerializeGetCredentialsCallsAcrossCredentialsManagerAndSecureCredentialsManager` validates this is intentional, this is a meaningful behavior change that:
- Could degrade performance if multiple credential managers are used concurrently (operations queue behind each other)
- Should be called out explicitly in the PR description and migration guide since it changes runtime behavior

Consider documenting this behavioral change separately, or splitting it into its own commit/PR so it's reviewable on its own merit.

#### 2. `** Imp:**` KDoc formatting (`SecureCredentialsManager.kt`)

The new doc comments use `** Imp:** This method is not thread safe` (lines ~117, ~162). This uses markdown bold syntax inside KDoc, which may not render correctly in all documentation generators (e.g., Dokka). Consider using `**Important:** This method is not thread safe` or `@implNote This method is not thread safe` for better cross-tool compatibility.

#### 3. `@Throws(CredentialsManagerException::class)` added to `saveApiCredentials`

A new `@Throws` annotation was added to `saveApiCredentials` (line ~171). This seems correct since the method uses `CryptoUtil.encrypt()` which can throw, but this annotation addition is unrelated to the constructor refactoring and should be called out.

---

### Nits

- **Line 991-992**: The `?: existingCredentials.type` was moved to the next line. While fine stylistically, this is an unrelated formatting change.
- **PR title**: `breaking :` has a space before the colon — minor convention issue, should be `breaking:`.
- **Import ordering** (`AuthenticationAPIClient.kt`): `java.util.concurrent.Executor` was added between `androidx` and `com.auth0` imports. Verify this matches the project's import ordering convention (typically `java.*` imports go after other imports in Kotlin Android projects, but this varies).

---

### What looks good

- **Migration guide section** is comprehensive with clear before/after examples for both Kotlin and Java
- **EXAMPLES.md** updates are thorough — all code samples correctly updated
- **4 new concurrency tests** covering: multi-instance renewal, single-instance renewal, valid credential access, and cross-manager serialization — well thought out
- **Sample app** updated consistently
- Clean removal of the `auth0` field from test setup (`SecureCredentialsManagerTest`)
- The `internal val executor` on `AuthenticationAPIClient` is appropriately scoped

### Verdict

The constructor removal itself is clean and well-documented. The main concern is the bundled executor-sharing change in `CredentialsManager` which changes runtime behavior beyond what the PR title suggests. Consider either:
1. Splitting the executor change into a separate PR, or
2. Updating the PR description and migration guide to explicitly document the new shared-executor behavior and its implications.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -47,7 +45,7 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         authenticationClient,
         storage,
         JWTDecoder(),
-        Executors.newSingleThreadExecutor()
+        authenticationClient.executor
```
utkrishtsahu, line 48: Can we add this change in migration guide as :Previously each instance created its own Executors.newSingleThreadExecutor(). Now all managers sharing the same Auth0 object share one executor.

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2138,67 +2132,316 @@ public class SecureCredentialsManagerTest {
...
+                        latch.countDown()
+                    }
+                })
+            }
         }
-        Mockito.`when`(cryptoMock.decrypt(any())).thenAnswer {
-            val input = it.arguments[0] as ByteArray
-            input
+
+        latch.await(10, TimeUnit.SECONDS)
```
utkrishtsahu: The return value of latch.await() is never asserted. If callbacks never fire  await returns false, the test passes silently, and the verify calls may also pass.

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2138,67 +2132,316 @@ public class SecureCredentialsManagerTest {
...
-                    SecureCredentialsManager(
-                        client,
-                        storage,
-                        cryptoMock,
-                        jwtDecoder,
-                        auth0.executor,
-                    )
-                secureCredsManager.getCredentials(object :
+
+        latch.await(10, TimeUnit.SECONDS)
```
utkrishtsahu: same as above

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2138,67 +2132,316 @@ public class SecureCredentialsManagerTest {
...
+                            result.accessToken,
+                            Is.`is`(validCredentials.accessToken)
                         )
+                        latch.countDown()
+                    }
+                })
+            }
+        }
+
+        latch.await(10, TimeUnit.SECONDS)
```
utkrishtsahu: same as above

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2138,67 +2132,316 @@ public class SecureCredentialsManagerTest {
...
+                            Is.`is`(renewedCredentials.refreshToken)
                         )
-                        MatcherAssert.assertThat(renewedCredentials.type, Is.`is`(result.type))
+                        latch.countDown()
+                    }
+                })
+            }
+        }
+
+        latch.await(10, TimeUnit.SECONDS)
```
utkrishtsahu: same as above

auth0/src/test/java/com/auth0/android/authentication/storage/SecureCredentialsManagerTest.kt
```diff
@@ -2138,67 +2132,316 @@ public class SecureCredentialsManagerTest {
...
             }
         }
-        latch.await() // Wait for all threads to finish
-        Mockito.verify(client, Mockito.times(1))
-            .renewAuth(
-                refreshToken = "refreshToken"
-            ) // verify that api client's renewAuth is called only once
-        Mockito.verify(request, Mockito.times(1)).execute() // Verify single network request
+
+        latch.await(10, TimeUnit.SECONDS)
```
utkrishtsahu: same as above

utkrishtsahu (changes requested):

utkrishtsahu (approved):

PR #962, opened 2026-04-29
feat: Added Auth tab support

pmathew92:
### Changes

This PT adds support for [Android Auth TabIntent](https://developer.android.com/reference/androidx/browser/auth/AuthTabIntent) to the `WebAuthentication` login and logout flow. 

### References

https://developer.android.com/reference/androidx/browser/auth/AuthTabIntent

### Testing

### Checklist

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
@@ -45,6 +45,10 @@ public open class AuthenticationActivity : ComponentActivity() {
 
     @Suppress("DEPRECATION")
     public override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        // If the Activity Result API (e.g. auth tab) already handled this result and called
+        // finish(), skip our legacy delivery to prevent a second delivery.
+        if (isFinishing) return
         val resultData = if (resultCode == RESULT_CANCELED) Intent() else data
         deliverAuthenticationResult(resultData)
         finish()
--- a/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
@@ -1,11 +1,13 @@
 package com.auth0.android.provider
 
-import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.VisibleForTesting
+import androidx.browser.auth.AuthTabIntent
 import com.auth0.android.authentication.AuthenticationException
 import com.auth0.android.authentication.AuthenticationException.Companion.ERROR_KEY_CT_OPTIONS_NULL
 import com.auth0.android.authentication.AuthenticationException.Companion.ERROR_KEY_URI_NULL
@@ -17,14 +19,31 @@ import com.auth0.android.provider.WebAuthProvider.resume
 import com.auth0.android.request.internal.CommonThreadSwitcher.Companion.getInstance
 import com.google.androidbrowserhelper.trusted.TwaLauncher
 
-public open class AuthenticationActivity : Activity() {
+public open class AuthenticationActivity : ComponentActivity() {
     private var intentLaunched = false
+    internal val authTabResultHandler = AuthTabResultHandler(
+        onSuccess = { uri ->
+            deliverAuthenticationResult(uri?.let { Intent().setData(it) } ?: Intent())
+            finish()
+        },
+        onFailure = { ex -> deliverAuthenticationFailure(ex) },
+        onCancel = {
+            deliverAuthenticationResult(Intent())
+            finish()
+        }
+    )
+
+    private val authTabLauncher: ActivityResultLauncher<Intent> =
+        AuthTabIntent.registerActivityResultLauncher(this) { result ->
+            authTabResultHandler.handle(result.resultCode, result.resultUri)
+        }
     private var customTabsController: CustomTabsController? = null
-    override fun onNewIntent(intent: Intent?) {
+    override fun onNewIntent(intent: Intent) {
         super.onNewIntent(intent)
         setIntent(intent)
     }
 
+    @Suppress("DEPRECATION")
     public override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         val resultData = if (resultCode == RESULT_CANCELED) Intent() else data
         deliverAuthenticationResult(resultData)
@@ -47,6 +66,10 @@ public open class AuthenticationActivity : Activity() {
 
     override fun onResume() {
         super.onResume()
+        // Auth Tab results are delivered via the Activity Result API callback (onAuthTabResult)
+        // before onResume, which calls finish(). Without this guard, onResume would treat the
+        // missing intent data as a cancellation and deliver a second, spurious result.
+        if (isFinishing) return
         val authenticationIntent = intent
         if (!intentLaunched && authenticationIntent.extras == null) {
             //Activity was launched in an unexpected way
@@ -113,7 +136,7 @@ public open class AuthenticationActivity : Activity() {
     internal open fun createCustomTabsController(
         context: Context, options: CustomTabsOptions
     ): CustomTabsController {
-        return CustomTabsController(context, options, TwaLauncher(context))
+        return CustomTabsController(context, options, TwaLauncher(context), authTabLauncher)
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
--- a/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
+++ b/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
@@ -143,7 +143,7 @@ class CustomTabsController extends CustomTabsServiceConnection {
                             TwaLauncher.CCT_FALLBACK_STRATEGY
                     );
                 } else if (customTabsOptions.isAuthTab()) {
-                    launchAsAuthTab(context, uri);
+                    launchAsAuthTab(context, uri, threadSwitcher, failureCallback);
                 } else {
                     launchAsDefault(context, uri);
                 }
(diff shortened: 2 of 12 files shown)


auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
```diff
@@ -17,15 +19,36 @@ import com.auth0.android.provider.WebAuthProvider.resume
...
-    override fun onNewIntent(intent: Intent?) {
+    override fun onNewIntent(intent: Intent) {
         super.onNewIntent(intent)
         setIntent(intent)
     }
 
+    @Suppress("DEPRECATION")
     public override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        // If the Activity Result API (e.g. auth tab) already handled this result and called
```
utkrishtsahu, line 49: remove this generated comment

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
```diff
@@ -47,6 +70,10 @@ public open class AuthenticationActivity : Activity() {
 
     override fun onResume() {
         super.onResume()
+        // Auth Tab results are delivered via the Activity Result API callback (onAuthTabResult)
```
utkrishtsahu, line 73: remove this generated comment

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
pmathew92: This is added intentionally to highlight why this if check is needed

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
pmathew92: This is added intentionally to highlight why this if check is needed

auth0/src/test/java/com/auth0/android/provider/CustomTabsControllerTest.java
```diff
@@ -339,6 +355,106 @@ public void shouldThrowExceptionIfFailedToLaunchBecauseOfException() {
...
+    }
+
+    @Test
+    public void shouldFallbackToCustomTabWhenRedirectUriHasNoScheme() {
+        @SuppressWarnings("unchecked")
+        ActivityResultLauncher<Intent> mockAuthTabLauncher = mock(ActivityResultLauncher.class);
+        customTabsClientMock = Mockito.mockStatic(CustomTabsClient.class);
+        customTabsClientMock.when(() -> CustomTabsClient.isAuthTabSupported(eq(context), any(String.class))).thenReturn(true);
+
+        CustomTabsOptions ctOptions = CustomTabsOptions.newBuilder()
```
utkrishtsahu, line 508: missing .withAuthTab() — test never enters the auth tab code path, passes for the wrong reason

auth0/src/main/java/com/auth0/android/provider/CustomTabsOptions.java
```diff
@@ -315,27 +361,16 @@ public Builder withDisabledCustomTabsPackages(List<String> disabledCustomTabsPac
...
-         * <p><b>Warning:</b> Ephemeral browsing support in Auth0.Android is still experimental
-         * and can change in the future. Please test it thoroughly in all the targeted browsers
-         * and OS variants and let us know your feedback.</p>
-         *
-         * @return this same builder instance.
-         */
-        @ExperimentalAuth0Api
-        @NonNull
-        public Builder withEphemeralBrowsing() {
+        Builder withEphemeralBrowsing() {
```
utkrishtsahu, line 364: visibility reduced from public to package-private without callout; this is a breaking API change

pmathew92: No.. this is inside the CustomTabsOptions class. Not the public API

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
```diff
@@ -17,15 +19,36 @@ import com.auth0.android.provider.WebAuthProvider.resume
 import com.auth0.android.request.internal.CommonThreadSwitcher.Companion.getInstance
 import com.google.androidbrowserhelper.trusted.TwaLauncher
 
-public open class AuthenticationActivity : Activity() {
+public open class AuthenticationActivity : ComponentActivity() {
```
utkrishtsahu, line 22: ComponentActivity() — base class change from Activity is public API-breaking for subclassers;
Can we document in V4 migration guide

auth0/src/test/java/com/auth0/android/provider/CustomTabsControllerTest.java
pmathew92: Good catch. Fixed

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
```diff
@@ -17,15 +19,36 @@ import com.auth0.android.provider.WebAuthProvider.resume
...
+        }
+    )
+
+    private val authTabLauncher: ActivityResultLauncher<Intent> =
+        AuthTabIntent.registerActivityResultLauncher(this) { result ->
+            authTabResultHandler.handle(result.resultCode, result.resultUri)
+        }
     private var customTabsController: CustomTabsController? = null
-    override fun onNewIntent(intent: Intent?) {
+    override fun onNewIntent(intent: Intent) {
```
utkrishtsahu, line 41: signature changed from nullable to non-null; subclass overrides of the old signature will break

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
pmathew92: this is within the SDK. Not for public APIs. It is open for testing purpose. Do you see scenario where this can fail for users updating to V4 ?

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
pmathew92: Again an internal implementation. No effect for consuming clients

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -128,6 +143,8 @@ public void launchUri(@NonNull final Uri uri, final boolean launchAsTwa, ThreadS
                             null,
                             TwaLauncher.CCT_FALLBACK_STRATEGY
                     );
+                } else if (customTabsOptions.isAuthTab()) {
```
utkrishtsahu, line 145: else if (customTabsOptions.isAuthTab()) — TWA silently takes priority over Auth Tab; consider logging a warning or documenting this precedence

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
```diff
@@ -17,15 +19,36 @@ import com.auth0.android.provider.WebAuthProvider.resume
 import com.auth0.android.request.internal.CommonThreadSwitcher.Companion.getInstance
 import com.google.androidbrowserhelper.trusted.TwaLauncher
 
-public open class AuthenticationActivity : Activity() {
+public open class AuthenticationActivity : ComponentActivity() {
     private var intentLaunched = false
+    internal val authTabResultHandler = AuthTabResultHandler(
```
utkrishtsahu, line 24: internal val authTabResultHandler — internal exposes to entire module, not just tests; consider private + @VisibleForTesting(PRIVATE)

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
```diff
@@ -141,6 +158,52 @@ public void launchUri(@NonNull final Uri uri, final boolean launchAsTwa, ThreadS
...
+            return;
+        }
+        String redirectUri = uri.getQueryParameter("redirect_uri");
+        if (redirectUri == null) {
+            Log.w(TAG, "Could not determine redirect URI from authorize URL. Falling back to Custom Tab.");
+            launchAsDefault(context, uri);
+            return;
+        }
+
+        bindService();
```
utkrishtsahu, line 197: bindService() called before scheme null-check at line 198; if scheme is null, this work is wasted before falling back to launchAsDefault which calls bindService again

utkrishtsahu (changes requested):

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
pmathew92: Added comments and document to highlight this

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
pmathew92: We intentionally chose internal over private + @VisibleForTesting as part of the collaborator pattern                                                                                                                                                                                                                 
  - AuthTabResultHandler is internal itself, so the combination is module-scoped regardless                                                                                                                                   
  - The collaborator pattern was specifically chosen to avoid @VisibleForTesting on the activity — adding it back would contradict the design                                                                                 
  - Tests in the same module access it cleanly without annotation

auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java
pmathew92: Good point. Moved  scheme check above bindService()

auth0/src/main/java/com/auth0/android/provider/AuthenticationActivity.kt
utkrishtsahu: since this targets v4 and subclassing AuthenticationActivity isn't a documented use case, this is fine. No action needed.

auth0/src/test/java/com/auth0/android/provider/CustomTabsControllerTest.java
utkrishtsahu: the test now has .withAuthTab() but still uses the default BrowserPicker (no mock returning a package). This means preferredPackage will be null, so the method falls back at the first guard (preferredPackage == null) — not at the scheme check it's supposed to test. The test still passes for the wrong reason, just at a different fallback point.

utkrishtsahu (changes requested):

utkrishtsahu (approved):

PR #967, opened 2026-05-14
feat :Add partial support for PAR auth flow

pmathew92:
### Changes
adds support for Pushed Authorization Request (PAR) flows where the backend-for-frontend (BFF) handles the /par and /token endpoints while the SDK manages the browser-based authorization.

####Usage
```
// Callback style
WebAuthProvider.authorizeWithRequestUri(account)
    .start(context, requestUri, object : Callback<AuthorizationCode, AuthenticationException> {
        override fun onSuccess(result: AuthorizationCode) {
            // Exchange result.code for tokens via BFF
        }
        override fun onFailure(error: AuthenticationException) {
            // Handle error
        }
    })

// Coroutines
val authCode = WebAuthProvider.authorizeWithRequestUri(account)
    .await(context, requestUri)
// Exchange authCode.code for tokens via BFF
```

### Testing

### Checklist


auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
```diff
@@ -0,0 +1,65 @@
+package com.auth0.android.provider
+
+import android.content.Context
+import android.net.Uri
+import android.util.Log
```
utkrishtsahu: Uri and Log can be removed not used.

auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val callback: Callback<AuthorizationCode, AuthenticationException>,
+    private val requestUri: String,
+    private val sessionTransferToken: String? = null,
+    private val ctOptions: CustomTabsOptions = CustomTabsOptions.newBuilder().build()
+) : ResumableManager() {
+
+    private var requestCode = 0
+
+    private companion object {
+        private val TAG = "PARCodeManager"
```
utkrishtsahu: TAG can be removed not used.

auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
```diff
@@ -0,0 +1,65 @@
+package com.auth0.android.provider
```
utkrishtsahu, line 1: This manager doesn't survive Activity recreation. If the user rotates the device while the browser is open, managerInstance is lost and the callback silently fails.
OAuthManager handles this via toState() / onRestoreInstanceState(). PARCodeManager has no equivalent.

auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun startAuthentication(context: Context, requestCode: Int) {
+        this.requestCode = requestCode
+        val additionalParams = buildMap {
+            sessionTransferToken?.let { put("session_transfer_token", it) }
+        }
+        val uri = PARUtils.buildAuthorizeUri(account, requestUri, additionalParams)
+        AuthenticationActivity.authenticateUsingBrowser(context, uri, false, ctOptions)
+    }
+
+    override fun resume(result: AuthorizeResult): Boolean {
```
utkrishtsahu, line 33: There's no state validation here. The existing OAuth flow generates a state and validates it on callback.
 The PAR flow passes state through in the result but never checks it.

auth0/src/main/java/com/auth0/android/provider/WebAuthProvider.kt
```diff
@@ -647,4 +664,139 @@ public object WebAuthProvider : SenderConstraining<WebAuthProvider> {
...
+            val manager = PARCodeManager(
+                account = account,
+                callback = callback,
+                requestUri = requestUri,
+                sessionTransferToken = sessionTransferToken,
+                ctOptions = ctOptions
+            )
+
+            managerInstance = manager
+            manager.startAuthentication(context, 110)
```
utkrishtsahu: private const val AUTH_REQUEST_CODE = 110
also used in line 607

utkrishtsahu (changes requested):

auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
pmathew92: Added

auth0/src/main/java/com/auth0/android/provider/PARCodeManager.kt
pmathew92: SDK doesn't know the value of the state to validate against. The state and code_verifier is handled by the consuming application via the BFF approach. The SDK in this case just authorize the user and returns the code and state back to the user.

pmathew92: One thing we can do in this case is take an input state value from the user and compare that against the returned state

utkrishtsahu (approved):

PR #975, opened 2026-05-26
feat: CTE delegation and impersonation support

pmathew92:
### Changes

Adds delegation and impersonation support for Custom Token Exchange (CTE) per RFC 8693, allowing an actor (e.g. an AI agent or support rep) to act on behalf of a user.

- Add actor_token and actor_token_type to CustomTokenExchangeOptions
- Add ActorClaim type  to  UserProfile

### Testing

### Checklist


auth0/src/main/java/com/auth0/android/request/internal/UserProfileDeserializer.java
```diff
@@ -11,7 +12,9 @@
 import com.google.gson.reflect.TypeToken;
 
 import java.lang.reflect.Type;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
```
utkrishtsahu: import unused

auth0/src/main/java/com/auth0/android/result/UserProfile.kt
```diff
@@ -25,7 +25,13 @@ public class UserProfile(
     private val extraInfo: Map<String, Any>?,
     private val userMetadata: Map<String, Any>?,
     private val appMetadata: Map<String, Any>?,
-    public val givenName: String?
+    public val givenName: String?,
```
utkrishtsahu, line 28: Since UserProfiles constructor is public, adding a new parameter is a binary-breaking change for any Java callers that construct this directly (Kotlin handles it via the default value, but Java doesn't see default params). I know it's mostly constructed internally by the deserializer, but would it be worth adding @JvmOverloads here to generate the overloaded constructors for Java consumers? Or alternatively, a secondary constructor without actor for backward compat

auth0/src/main/java/com/auth0/android/authentication/request/CustomTokenExchangeOptions.kt
```diff
@@ -0,0 +1,13 @@
+package com.auth0.android.authentication.request
+
+/**
+ * Data class representing custom token exchange options.
+ * @param actorToken A token representing the acting party for delegation or impersonation scenarios. Used when one principal needs to act on behalf of another.
+ *  For example, an AI agent acting on behalf of a user.
+ * @param actorTokenType The type identifier for the actor token. Must be a URI under your organization's control, following the same
+ *  rules as subject_token_type.
+ */
+public data class CustomTokenExchangeOptions(
```
utkrishtsahu: should we validate that neither value is blank? e.g., require(actorToken.isNotBlank()) in an init block? Or are we okay leaving that to the server to reject?

utkrishtsahu (changes requested):

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -815,17 +816,27 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
...
     public fun customTokenExchange(
         subjectTokenType: String,
         subjectToken: String,
-        organization: String? = null
+        organization: String? = null,
+        customTokenExchangeOptions: CustomTokenExchangeOptions? = null
     ): AuthenticationRequest {
-        return tokenExchange(subjectTokenType, subjectToken, organization)
+        return tokenExchange(
+            subjectTokenType,
```
kishore7snehil, line 839: Adding `customTokenExchangeOptions` as a new default parameter to the existing public `customTokenExchange` method changes the JVM method signature. Kotlin default arguments generate synthetic `$default` methods, but the exact signature changes. Already-compiled Kotlin or Java callers using the old 3-argument method could encounter `NoSuchMethodError` after upgrading the SDK.

Similar to the UserProfile concern raised by @utkrishtsahu - consider retaining the original 3-parameter overload (delegating with `options = null`) and adding a separate 4-parameter overload, or annotating with `@JvmOverloads` to ensure the old Java-visible signatures are preserved.

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
```diff
@@ -1117,6 +1129,10 @@ public class AuthenticationAPIClient @VisibleForTesting(otherwise = VisibleForTe
             organization?.let {
                 set(ORGANIZATION_KEY, it)
             }
+            customTokenExchangeOptions?.let {
+                set(ACTOR_TOKEN_KEY, it.actorToken)
+                set(ACTOR_TOKEN_TYPE_KEY, it.actorTokenType)
+            }
         }.asDictionary()
```
kishore7snehil, line 1140: Per RFC 8693 and Auth0's implementation, when `actor_token` is present in the request, Auth0 will NOT issue a refresh token regardless of whether `offline_access` is in the scope. The response will have `refresh_token: null`.

Need to verify that `Credentials` and `CredentialsManager.saveCredentials()` handle null `refresh_token` gracefully in this flow. If any downstream code expects a non-null refresh_token or logs a warning when it's missing, this could cause unexpected behavior for developers using delegation.

Consider adding a test case that verifies the full flow works correctly when the token response contains no refresh_token (simulating the delegation scenario), and document this behavior in the method's KDoc.

auth0/src/main/java/com/auth0/android/request/internal/UserProfileDeserializer.java
```diff
@@ -44,10 +47,34 @@ public UserProfile deserialize(JsonElement json, Type typeOfT, JsonDeserializati
...
+            return null;
+        }
+
+        JsonObject actObject = actElement.getAsJsonObject();
+        String sub = context.deserialize(actObject.remove("sub"), String.class);
+        if (sub == null) {
+            return null;
+        }
+
+        ActorClaim nestedActor = deserializeActorClaim(actObject.remove("act"), context);
```
kishore7snehil, line 76: If the `act` claim is present but has no `sub` field, `deserializeActorClaim` returns `null` silently. While defensively correct, developers won't know that an `act` claim was received but couldn't be parsed. A log statement here would help with debugging delegation configuration issues (e.g., Action not setting `sub` correctly in `api.authentication.setActor()`).

kishore7snehil (changes requested):

auth0/src/main/java/com/auth0/android/result/ActorClaim.kt
```diff
@@ -0,0 +1,17 @@
+package com.auth0.android.result
+
+import java.io.Serializable
+
+/**
+ * Represents the `act` (actor) claim in an ID token, used in delegation and impersonation scenarios.
+ * See RFC 8693 Section 4.1 for the specification of the `act` claim.
+ *
+ * @param sub The unique identifier of the actor (required).
```
kishore7snehil, line 9: This references "RFC 8693 Section 4.1" but the `act` claim is defined in **Section 4.4** (Token Exchange Act Claim). Section 4.1 covers the token response structure.

kishore7snehil (changes requested):

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: This is already present. If no RT is returned, the Credentials won't have it. Since SDK doesn't save the token automatically it is upto the user to handle this scenario. We need to document this properly

auth0/src/main/java/com/auth0/android/authentication/request/CustomTokenExchangeOptions.kt
pmathew92: Will leave it to server to validate the same

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
pmathew92: Added @JvmOverloads to the method

auth0/src/main/java/com/auth0/android/request/internal/UserProfileDeserializer.java
pmathew92: Done

auth0/src/main/java/com/auth0/android/result/ActorClaim.kt
pmathew92: Corrected

auth0/src/main/java/com/auth0/android/result/UserProfile.kt
pmathew92: Done

auth0/src/main/java/com/auth0/android/authentication/request/CustomTokenExchangeOptions.kt
NandanPrabhu: blank as in empty @utkrishtsahu ?

auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt
kishore7snehil: Okay, it's worth adding a note in the doc for customTokenExchange that when `actor_token` is present, Auth0 suppresses the `refresh_token` in the response, so `Credentials.refreshToken` will be null.

kishore7snehil (dismissed):
LGTM!

kishore7snehil (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Change Log
 
+## [3.18.0](https://github.com/auth0/Auth0.Android/tree/3.18.0) (2026-05-26)
+[Full Changelog](https://github.com/auth0/Auth0.Android/compare/3.17.0...3.18.0)
+
+**Added**
+- feat : Add support to pass type parameter to the MyAccount list authentication method API [\#974](https://github.com/auth0/Auth0.Android/pull/974) ([pmathew92](https://github.com/pmathew92))
```
NandanPrabhu, line 7: do we need CHANGELOG.md changes?

.version
```diff
@@ -1 +1 @@
-3.17.0
\ No newline at end of file
+3.18.0
```
NandanPrabhu, line 1: these too. Looks like release changes are added in this. may be a rebase to main is required

pmathew92: Rebased

CHANGELOG.md
pmathew92: Rebased

utkrishtsahu (approved):

Issue #972, opened 2026-05-20
OAuthManagerKt.fromState passes null for DPoP, breaking token exchange after process death


hamed17n:
### Checklist

- [x] The issue can be reproduced in the [Auth0.Android sample app](https://github.com/auth0-samples/auth0-android-sample/tree/master/00-Login-Kt) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/Auth0.Android#readme), [Examples](https://github.com/auth0/Auth0.Android/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.Android/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0.android/auth0/latest/index.html) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/Auth0.Android/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

## Summary

When the Android host process is killed while Chrome Custom Tabs is open during a
DPoP-enabled Universal Login flow, the Auth0 SDK correctly persists `OAuthManagerState`
to `savedInstanceState`. On process restart, `AuthenticationActivity` deserialises the
state and calls `OAuthManagerKt.fromState`. However, `fromState` hardcodes `null` for
the `dPoP` constructor argument instead of passing `state.getDPoP()`. The reconstructed
`OAuthManager` also builds its internal `apiClient` via
`new AuthenticationAPIClient(account)` without calling `useDPoP(context)`. As a result,
the `/oauth/token` exchange after process restart is sent **without a DPoP proof header**,
and the server rejects it with HTTP 400 `invalid_request`.

### Reproduction

1. Initialise the SDK with DPoP enabled:
   `WebAuthProvider.useDPoP(context)` + authorise request that includes `dpop_jkt=…`
   in the redirect URL.
2. Start the login flow: `WebAuthProvider.login(auth0).useDPoP(context).start(activity, callback)`.
3. While Chrome Custom Tabs is open, kill the host process:
   `adb shell input keyevent KEYCODE_HOME && adb shell am kill <package>`
4. Reopen Chrome, complete the login in the browser.
[...]

PR #977, opened 2026-06-08
fix : Handling DPoP enabled WebAuth flow after process death

pmathew92:
### Changes

  Fixes a bug where the OAuth token exchange fails after process death during a
  DPoP-enabled Universal Login flow.

  When the Android host process is killed while Chrome Custom Tabs is open, the SDK
  persists `OAuthManagerState` to `savedInstanceState` and rebuilds the manager via
  `OAuthManager.fromState` on restart. Previously, the DPoP configuration was lost on
  this restore path:

  - `OAuthManager.toState()` did not persist the DPoP-enabled flag.
  - `OAuthManager.fromState()` reconstructed the manager without DPoP, and the restored
    `PKCE`'s `AuthenticationAPIClient` was never DPoP-enabled.

  As a result, the `/oauth/token` exchange after process restart was sent **without a
  DPoP proof header**, and the server rejected it with HTTP 400 `invalid_request`
  ("Authorization Code is bound to a DPoP key, but DPoP header is missing or invalid").

  This change persists a `dPoPEnabled` boolean in `OAuthManagerState` (the `DPoP`
  object holds a `Context` and cannot be serialized directly), and on restore re-creates
  the `DPoP` instance and re-enables DPoP on the restored API client so the token
  exchange carries the proof header.

  ### References

  - Original report: #972
  - Tracked downstream in `react-native-auth0` #1549

  ### Files changed

  - `OAuthManager.kt` — `toState` persists `dPoPEnabled`; `fromState` takes a `Context`,
    re-enables DPoP on the restored API client, and rebuilds the `DPoP` instance.
  - `OAuthManagerState.kt` — replaced the non-serializable `DPoP?` field with
    `dPoPEnabled: Boolean`.
  - `WebAuthProvider.kt` — `onRestoreInstanceState` accepts a `Context` and forwards it
    to `fromState`. PAR restore path is unchanged.
  - `AuthenticationActivity.kt` — passes the activity `Context` into
    `onRestoreInstanceState`.

  ### Tests

  - `OAuthManagerStateTest` — serialization of `dPoPEnabled` (true / default false /
    legacy JSON missing the field defaults to false, preserving back-compat with
    previously persisted state).
  - `WebAuthProviderTest` — added a behavioral test that simulates the process
    death → restore path and asserts DPoP is re-enabled on the restored manager.

### Testing

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Restores DPoP-enabled authentication after app/process restoration so resumed sign-ins continue safely.
  * Ensures OAuth manager is fully reconstructed on restore, preserving authentication configuration and pending results to prevent lost responses.
  * Improves activity restoration to correctly rehydrate auth state.

* **Tests**
  * Added regression tests covering DPoP state persistence and restoration to prevent future regressions.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
@@ -394,7 +394,7 @@ internal fun OAuthManager.Companion.fromState(
 ): OAuthManager {
     // Enable DPoP on the restored PKCE's AuthenticationAPIClient so that
     // the token exchange request includes the DPoP proof after process restore.
-    if (state.dPoPEnabled  && state.pkce != null) {
+    if (state.dPoPEnabled && state.pkce != null) {
         state.pkce.apiClient.useDPoP(context)
     }
     return OAuthManager(
@@ -403,7 +403,7 @@ internal fun OAuthManager.Companion.fromState(
         parameters = state.parameters,
         callback = callback,
         customAuthorizeUrl = state.customAuthorizeUrl,
-        dPoP = if (state.dPoPEnabled ) DPoP(context) else null
+        dPoP = if (state.dPoPEnabled) DPoP(context) else null
     ).apply {
         setHeaders(
             state.headers
--- a/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
+++ b/auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
@@ -29,6 +29,7 @@ internal class OAuthManager(
     @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal val dPoP: DPoP? = null
 ) : ResumableManager() {
+
     private val parameters: MutableMap<String, String>
     private val headers: MutableMap<String, String>
     private val ctOptions: CustomTabsOptions
@@ -211,7 +212,8 @@ internal class OAuthManager(
             auth0 = account,
             idTokenVerificationIssuer = idTokenVerificationIssuer,
             idTokenVerificationLeeway = idTokenVerificationLeeway,
-            customAuthorizeUrl = this.customAuthorizeUrl
+            customAuthorizeUrl = this.customAuthorizeUrl,
+            dPoPEnabled = dPoP != null
         )
     }
 
@@ -387,14 +389,21 @@ internal class OAuthManager(
 
 internal fun OAuthManager.Companion.fromState(
     state: OAuthManagerState,
-    callback: Callback<Credentials, AuthenticationException>
+    callback: Callback<Credentials, AuthenticationException>,
+    context: Context
 ): OAuthManager {
+    // Enable DPoP on the restored PKCE's AuthenticationAPIClient so that
+    // the token exchange request includes the DPoP proof after process restore.
+    if (state.dPoPEnabled  && state.pkce != null) {
+        state.pkce.apiClient.useDPoP(context)
+    }
     return OAuthManager(
         account = state.auth0,
         ctOptions = state.ctOptions,
         parameters = state.parameters,
         callback = callback,
-        customAuthorizeUrl = state.customAuthorizeUrl
+        customAuthorizeUrl = state.customAuthorizeUrl,
+        dPoP = if (state.dPoPEnabled ) DPoP(context) else null
     ).apply {
         setHeaders(
             state.headers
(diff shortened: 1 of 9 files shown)


pmathew92:
@coderabbit review

auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
```diff
@@ -387,14 +389,21 @@ internal class OAuthManager(
 
 internal fun OAuthManager.Companion.fromState(
     state: OAuthManagerState,
-    callback: Callback<Credentials, AuthenticationException>
+    callback: Callback<Credentials, AuthenticationException>,
+    context: Context
 ): OAuthManager {
+    // Enable DPoP on the restored PKCE's AuthenticationAPIClient so that
+    // the token exchange request includes the DPoP proof after process restore.
+    if (state.dPoPEnabled  && state.pkce != null) {
```
utkrishtsahu: Double space in dPoPEnabled  &&

auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
```diff
@@ -387,14 +389,21 @@ internal class OAuthManager(
...
+        state.pkce.apiClient.useDPoP(context)
+    }
     return OAuthManager(
         account = state.auth0,
         ctOptions = state.ctOptions,
         parameters = state.parameters,
         callback = callback,
-        customAuthorizeUrl = state.customAuthorizeUrl
+        customAuthorizeUrl = state.customAuthorizeUrl,
+        dPoP = if (state.dPoPEnabled ) DPoP(context) else null
```
utkrishtsahu: The real fix is the useDPoP(context) call above — this dPoP arg only gets read in startAuthentication(), which doesn't run on resume. Is it set here just to keep state consistent / for the test to assert on? Want to confirm it's intentional.

auth0/src/main/java/com/auth0/android/provider/OAuthManagerState.kt
```diff
@@ -37,7 +36,7 @@ internal data class OAuthManagerState(
         val idTokenVerificationLeeway: Int?,
         val idTokenVerificationIssuer: String?,
         val customAuthorizeUrl: String? = null,
-        val dPoP: DPoP? = null
+        val dPoPEnabled: Boolean
```
utkrishtsahu: Could we default this to = false? The outer class does, and since back-compat with legacy JSON is the point here, better to be explicit than rely on Gson zero-filling.

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
```diff
@@ -2958,6 +2959,27 @@ public class WebAuthProviderTest {
...
+        val bundle = Bundle()
+        WebAuthProvider.onSaveInstanceState(bundle)
+
+        // Simulate the host process being killed and recreated: the manager instance is gone,
+        // and the activity is recreated with the saved state.
+        WebAuthProvider.resetManagerInstance()
+        WebAuthProvider.onRestoreInstanceState(bundle, activity)
+
+        val restoredManager = WebAuthProvider.managerInstance as OAuthManager
+        assertThat(restoredManager.dPoP, `is`(notNullValue()))
```
utkrishtsahu, line 2990: This only asserts dPoP is non-null, which checks the constructor arg — not the actual fix. The test would still pass if pkce.apiClient.useDPoP(context) were removed. Could we assert DPoP is enabled on the restored PKCE's API client instead?

utkrishtsahu:
@pmathew92 The PR notes the PAR restore path (PARCodeManager.fromState) is left unchanged. If a DPoP + PAR Universal Login flow is supported in this SDK, wouldn't it hit the same process-death bug and not be covered by this fix? Could you confirm whether that combination is possible?

utkrishtsahu (changes requested):

auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
pmathew92: Yes, the dPoP arg is intentional — for state-consistency on toState(), round-trip + the test assertion.

pmathew92:
@utkrishtsahu  PAR flow doesn't support DPoP since in PAR the jwk exchange is done in the client backend. So the flow is not applicable

auth0/src/test/java/com/auth0/android/provider/WebAuthProviderTest.kt
pmathew92: Added test to assert on restored API client

auth0/src/main/java/com/auth0/android/provider/OAuthManagerState.kt
pmathew92: Done

auth0/src/main/java/com/auth0/android/provider/OAuthManager.kt
pmathew92: Fixed

utkrishtsahu (dismissed):

utkrishtsahu (approved):

PR #983, opened 2026-06-18
feat: enforce IPSIE session_expiry ceiling in credentials managers

utkrishtsahu:
### Changes

This change adds support for the IPSIE SL1 session_expiry claim in the credentials managers. When an upstream identity provider (e.g. an enterprise/OIDC connection) asserts a session_expiry claim on the ID token, the SDK now treats it as a hard session ceiling: an absolute lifetime, expressed in Unix seconds, beyond which the local session is no longer valid and cannot be extended by a refresh-token renewal. This is important for downstream/federated scenarios where the upstream IdP dictates a maximum session lifetime that Auth0-issued tokens must respect, independently of the access-token expiry.

**Behavior**

- The claim is read at login and persisted to storage (key com.auth0.session_expiry) so that a subsequent refresh whose ID token omits the claim does not silently drop the ceiling.
- On every getCredentials / hasValidCredentials call, if the ceiling has been reached the stored credentials are cleared and the call fails with CredentialsManagerException.SESSION_EXPIRED. The refresh token is never used to renew past the ceiling.
- A 30-second negative clock-skew leeway is applied — the session is treated as expired slightly before the wall-clock ceiling, never after.
- A missing claim means "no ceiling" → fully backward compatible.
- clearCredentials() removes the persisted ceiling so it does not leak past logout.

**Public API added**

- Credentials.sessionExpiresAt: Long? — nullable Unix-seconds ceiling for a credential set; null when the connection does not emit the claim.
- CredentialsManagerException.SESSION_EXPIRED (new Code.SESSION_EXPIRED + public exception constant) — surfaced when the ceiling is reached.

**Internal/affected classes (no breaking signature changes)**

- Jwt — added internal sessionExpiry: Long? parsing of the session_expiry payload claim.
- BaseCredentialsManager — added isSessionExpired(...), persistSessionExpiry(...), validateSessionExpiryAtCreation(...), the KEY_SESSION_EXPIRY storage key, and the 30s leeway constant.
- CredentialsManager and SecureCredentialsManager — persist the ceiling on save and enforce/clear on retrieval; remove the key on clearCredentials().

**Behavioral / upgrade note:** For a user whose connection asserts session_expiry, a getCredentials call that previously succeeded can now fail with SESSION_EXPIRED once the ceiling is reached. Callers should treat SESSION_EXPIRED as a prompt to re-authenticate. Documented in EXAMPLES.md (new "Upstream session expiry" section).

### References

SDK-9590

### Testing

Reviewers can run the existing unit suite — ./gradlew :auth0:testDebugUnitTest — which now includes coverage for the new behavior:

- JwtTest — parsing session_expiry as Long seconds, and absence of the claim.
- CredentialsTest — sessionExpiresAt returns the ceiling / null when absent.
- CredentialsManagerTest and SecureCredentialsManagerTest — persisting the ceiling at save, valid-before-ceiling retrieval, SESSION_EXPIRED (with credentials cleared) once the ceiling passes, refresh not extending past the ceiling, and clearCredentials() removing the persisted key.

**End-to-end**, the feature was manually verified on a real device (Pixel 7a, Android 16) against a live tenant. Because the platform does not yet emit session_expiry natively, the claim was injected via a Post-Login Action (api.idToken.setCustomClaim('session_expiry', Math.floor(Date.now()/1000) + N)); all behaviors above were confirmed on both CredentialsManager and SecureCredentialsManager. 

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added upstream `session_expiry` ceiling enforcement across credential, SSO, and API-credential flows, including persisted/pinned ceiling and clock-skew leeway.
  * Exposed `Credentials.sessionExpiresAt` and added `CredentialsManagerException.SESSION_EXPIRED`.
* **Bug Fixes**
  * Cached/renewed credentials are no longer served once the ceiling is reached; stored credentials and the persisted ceiling are cleared.
  * `hasValidCredentials` now returns false after the ceiling is breached.
* **Documentation**
  * Updated error-handling examples and added guidance on `session_expiry` ceiling behavior and leeway.
* **Tests**
  * Added/expanded coverage for ceiling parsing, pinning behavior, and failure scenarios.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2937,6 +2937,7 @@ When an enterprise connection (for example an OIDC or Okta connection) is config
 The credentials managers enforce this ceiling automatically:
 
 - The ceiling is read from the ID token at login and persisted, so it survives refreshes whose ID token does not re-emit the claim.
+- `saveCredentials` rejects an already-expired session up front: if the ID token is already past its ceiling at login, the save throws `CredentialsManagerException.SESSION_EXPIRED` and nothing is persisted.
 - On every `getCredentials` call, if the ceiling has been reached the stored credentials are cleared and the call fails with `CredentialsManagerException.SESSION_EXPIRED`. The refresh token is **never** used to renew a session past the ceiling.
 - A small negative clock-skew leeway (~30 seconds) is applied, so the session is treated as expired slightly *before* the wall-clock ceiling, never after.
 - Connections that do not emit the claim are unaffected — there is no ceiling and behavior is unchanged.
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2886,6 +2886,7 @@ In the event that something happened while trying to save or retrieve the creden
 - **DPoP key pair lost** — The DPoP key pair is no longer available in the Android KeyStore. The stored credentials are cleared and re-authentication is required.
 - **DPoP key pair mismatch** — The DPoP key pair exists but is different from the one used when the credentials were saved. The stored credentials are cleared and re-authentication is required.
 - **DPoP not configured** — The stored credentials are DPoP-bound but the `AuthenticationAPIClient` used by the credentials manager was not configured with `useDPoP(context)`. The developer needs to call `AuthenticationAPIClient(auth0).useDPoP(context)` and pass the configured client to the credentials manager.
+- **Session expired** — The session has reached the `session_expiry` ceiling asserted by the upstream identity provider. The stored credentials are cleared and re-authentication is required. See [Upstream session expiry](#upstream-session-expiry) below.
 
 You can access the `code` property of the `CredentialsManagerException` to understand why the operation with `CredentialsManager` has failed and the `message` property of the `CredentialsManagerException` would give you a description of the exception.
 
@@ -2919,8 +2920,34 @@ when(credentialsManagerException) {
         // Developer forgot to call useDPoP() on the AuthenticationAPIClient
         // passed to the credentials manager. Fix the client configuration.
     }
+
+    CredentialsManagerException.SESSION_EXPIRED -> {
+        // The upstream identity provider's session_expiry ceiling was reached.
+        // The stored credentials have already been cleared; prompt the user to
+        // re-authenticate.
+    }
     // ... similarly for other error codes
 }
+```
+
+### Upstream session expiry
+
+When an enterprise connection (for example an OIDC or Okta connection) is configured to assert a session lifetime, Auth0 includes a `session_expiry` claim in the ID token. This claim is an absolute ceiling — expressed in **Unix seconds** — on how long the local session may live, independently of the access-token expiry. It usually sits much further out than `expiresAt`, and it cannot be extended by a refresh-token renewal.
+
+The credentials managers enforce this ceiling automatically:
+
+- The ceiling is read from the ID token at login and persisted, so it survives refreshes whose ID token does not re-emit the claim.
+- On every `getCredentials` call, if the ceiling has been reached the stored credentials are cleared and the call fails with `CredentialsManagerException.SESSION_EXPIRED`. The refresh token is **never** used to renew a session past the ceiling.
+- A small negative clock-skew leeway (~30 seconds) is applied, so the session is treated as expired slightly *before* the wall-clock ceiling, never after.
+- Connections that do not emit the claim are unaffected — there is no ceiling and behavior is unchanged.
+
+> ⚠️ **Upgrade note:** For a user whose connection asserts `session_expiry`, a `getCredentials` call that previously succeeded can now fail with `SESSION_EXPIRED` once the ceiling is reached. Make sure your error handling treats `SESSION_EXPIRED` as a prompt to re-authenticate.
+
+You can read the ceiling for a given credential set from `Credentials.sessionExpiresAt` (a nullable `Long` of Unix seconds, `null` when the connection does not emit the claim):
+
+```kotlin
+val credentials = credentialsManager.awaitCredentials()
+val ceiling: Long? = credentials.sessionExpiresAt
 ```
 
 ## Passkeys
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -3063,7 +3063,23 @@ The credentials managers enforce this ceiling automatically:
 - A small negative clock-skew leeway (~30 seconds) is applied, so the session is treated as expired slightly *before* the wall-clock ceiling, never after.
 - Connections that do not emit the claim are unaffected — there is no ceiling and behavior is unchanged.
 
+> ⚠️ **The `session_expiry` value must be Unix seconds.** Per [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519), the claim is interpreted as seconds since the Unix epoch. A millisecond-magnitude value (e.g. `1700000000000`) resolves to a date ~50,000 years out and would **silently disable** the ceiling, so the SDK treats any implausibly large value (`>= 10_000_000_000`) as "no ceiling". The SDK also **fails open** on any malformed value — a non-numeric, zero, negative, or millisecond value is treated as "no ceiling" and the session proceeds without enforcement. When emitting the claim from an Action, always use seconds (divide a milliseconds timestamp by 1000).
+
 > ⚠️ **Upgrade note:** For a user whose connection asserts `session_expiry`, a `getCredentials` call that previously succeeded can now fail with `SESSION_EXPIRED` once the ceiling is reached. Make sure your error handling treats `SESSION_EXPIRED` as a prompt to re-authenticate.
+
+#### Emitting the claim
+
+The `session_expiry` claim is not emitted by default — it is set on your tenant by a [Post-Login Action](https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow) that adds it to the ID token, for example:
+
+```javascript
+exports.onExecutePostLogin = async (event, api) => {
+  // session_expiry must be expressed in Unix seconds
+  const sessionExpiry = Math.floor(Date.now() / 1000) + 8 * 60 * 60; // 8 hours from now
+  api.idToken.setCustomClaim('session_expiry', sessionExpiry);
+};
+```
+
+> 📝 A link to the canonical Auth0 `session_expiry` Action guide will be added here once it is published.
 
 You can read the ceiling for a given credential set from `Credentials.sessionExpiresAt` (a nullable `Long` of Unix seconds, `null` when the connection does not emit the claim):
 
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -349,14 +349,22 @@ public abstract class BaseCredentialsManager internal constructor(
     }
 
     /**
-     * Persists the `session_expiry` ceiling read from the given ID token, if present.
+     * Pins the `session_expiry` ceiling from the initial login and preserves it across refreshes.
      *
-     * To preserve the ceiling across refreshes (the refresh grant does not re-emit `session_expiry`),
-     * the stored value is only ever written, never cleared, when a fresh ID token omits the claim.
-     * Call from `saveCredentials`.
+     * The ceiling is read once and stamped onto the session at login: it is stored only when no value
+     * is already persisted. A `session_expiry` re-emitted on a later (refresh) grant is deliberately
+     * ignored, so the bound stays pinned to the initial-login value and a refresh can never extend the
+     * session past it. [clearCredentials] removes the stored value on logout, so the next login re-pins
+     * a fresh ceiling. Call from `saveCredentials`.
      */
     protected fun persistSessionExpiry(idToken: String?) {
-        sessionExpiryFromIdToken(idToken)?.let { storage.store(KEY_SESSION_EXPIRY, it) }
+        val incoming = sessionExpiryFromIdToken(idToken) ?: return
+        // A positive value is already pinned from the initial login -> keep it; ignore the claim
+        // re-emitted on this (refresh) grant. A null/non-positive stored value means nothing is pinned
+        // yet (mirrors the unset/migration guard in [isSessionExpired]), so stamp the ceiling now.
+        val pinned = storage.retrieveLong(KEY_SESSION_EXPIRY)
+        if (pinned != null && pinned > 0) return
+        storage.store(KEY_SESSION_EXPIRY, incoming)
     }
 
     /**
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -327,17 +327,19 @@ public abstract class BaseCredentialsManager internal constructor(
     /**
      * Checks whether the upstream-IdP session ceiling (`session_expiry`) has been reached.
      *
-     * The ceiling is resolved in order: (1) the live claim in [idToken]; (2) the value persisted at
-     * login under [KEY_SESSION_EXPIRY] (so a refresh whose ID token omits the claim does not silently
-     * drop the ceiling); (3) if neither is present there is no ceiling and the session is NOT expired
-     * — a missing value must fall through to existing behavior, never be treated as already-expired.
+     * The ceiling is resolved in order: (1) the value pinned at login under [KEY_SESSION_EXPIRY];
+     * (2) the live claim in [idToken], as a fallback only when nothing is pinned yet (migration of a
+     * session saved before this control existed); (3) if neither is present there is no ceiling and
+     * the session is NOT expired — a missing value must fall through to existing behavior, never be
+     * treated as already-expired. The pinned value is read first because the ceiling is fixed at the
+     * initial login: a refresh whose ID token re-emits a *later* `session_expiry` must never raise it.
      *
      * A small negative clock-skew leeway is applied so the session is treated as expired slightly
      * before the wall-clock ceiling, never after.
      */
     protected fun isSessionExpired(idToken: String?): Boolean {
-        val sessionExpiry = sessionExpiryFromIdToken(idToken)
-            ?: storage.retrieveLong(KEY_SESSION_EXPIRY)
+        val sessionExpiry = storage.retrieveLong(KEY_SESSION_EXPIRY)
+            ?: sessionExpiryFromIdToken(idToken)
             ?: return false
         // A non-positive ceiling is not a valid Unix timestamp; treat it as "no ceiling" rather than
         // already-expired (mirrors the guard in [willExpire] for unset/migration values).
@@ -346,6 +348,20 @@ public abstract class BaseCredentialsManager internal constructor(
         }
         val nowSeconds = currentTimeInMillis / 1000
         return nowSeconds + SESSION_EXPIRY_LEEWAY_SECONDS >= sessionExpiry
+    }
+
+    /**
+     * Stamps the pinned `session_expiry` ceiling (the value persisted at login under
+     * [KEY_SESSION_EXPIRY]) onto [credentials] so its public `sessionExpiresAt` reflects the value
+     * the SDK actually enforces, rather than re-decoding the live ID token — which would diverge
+     * after a refresh whose token omits or re-emits the claim. No-op when nothing is pinned, so the
+     * getter falls back to the token claim. Returns the same instance for call-site convenience.
+     */
+    protected fun stampPinnedSessionExpiry(credentials: Credentials): Credentials {
+        storage.retrieveLong(KEY_SESSION_EXPIRY)?.takeIf { it > 0 }?.let {
+            credentials.pinnedSessionExpiresAt = it
+        }
+        return credentials
     }
 
     /**
@@ -369,9 +385,13 @@ public abstract class BaseCredentialsManager internal constructor(
 
     /**
      * Validates, at session-creation time, that the given ID token is not already past its
-     * `session_expiry` ceiling (i.e. `session_expiry <= iat`). Throws [CredentialsManagerException]
-     * with code `SESSION_EXPIRED` when it is, so an already-expired session is never persisted.
-     * No-op when the token is absent or does not carry both `session_expiry` and `iat`.
+     * `session_expiry` ceiling. Throws [CredentialsManagerException] with code `SESSION_EXPIRED`
+     * when it is, so an already-expired session is never persisted. No-op when the token is absent
+     * or does not carry both `session_expiry` and `iat`.
+     *
+     * The same [SESSION_EXPIRY_LEEWAY_SECONDS] leeway used by [isSessionExpired] is applied here so
+     * the two checks agree: a ceiling within the leeway of `iat` is rejected up front rather than
+     * being persisted only to be treated as expired on the very next read.
      */
     @Throws(CredentialsManagerException::class)
     protected fun validateSessionExpiryAtCreation(idToken: String?) {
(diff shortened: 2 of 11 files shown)


auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -28,6 +28,19 @@ public abstract class BaseCredentialsManager internal constructor(
...
+         * survives a refresh whose ID token does not re-emit the claim. See [isSessionExpired].
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE)
+        internal const val KEY_SESSION_EXPIRY = "com.auth0.session_expiry"
+
+        /**
+         * Negative clock-skew leeway (seconds) applied when checking the `session_expiry` ceiling, so
+         * the session is treated as expired slightly *before* the wall-clock ceiling, never after.
+         */
+        private const val SESSION_EXPIRY_LEEWAY_SECONDS = 30L
```
pmathew92, line 43: Is 30s is the agreed leeway across ?

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,67 @@ public abstract class BaseCredentialsManager internal constructor(
...
+     * A small negative clock-skew leeway is applied so the session is treated as expired slightly
+     * before the wall-clock ceiling, never after.
+     */
+    protected fun isSessionExpired(idToken: String?): Boolean {
+        val sessionExpiry = sessionExpiryFromIdToken(idToken)
+            ?: storage.retrieveLong(KEY_SESSION_EXPIRY)
+            ?: return false
+        // A non-positive ceiling is not a valid Unix timestamp; treat it as "no ceiling" rather than
+        // already-expired (mirrors the guard in [willExpire] for unset/migration values).
+        if (sessionExpiry <= 0) {
```
pmathew92: The comment says non-positive but the check is against 0 also. Shouldn't this be `sessionExpiry < 0 `

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -68,13 +68,18 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         if (TextUtils.isEmpty(credentials.accessToken) && TextUtils.isEmpty(credentials.idToken)) {
             throw CredentialsManagerException.INVALID_CREDENTIALS
         }
+        // IPSIE session_expiry: reject a session already past its ceiling at creation time.
+        validateSessionExpiryAtCreation(credentials.idToken)
```
pmathew92, line 76: validateSessionExpiryAtCreation now throws an exception . Update the comment and signature of saveCredentials for the same

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -177,6 +177,8 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
         if (TextUtils.isEmpty(credentials.accessToken) && TextUtils.isEmpty(credentials.idToken)) {
             throw CredentialsManagerException.INVALID_CREDENTIALS
         }
+        // IPSIE session_expiry: reject a session already past its ceiling at creation time.
+        validateSessionExpiryAtCreation(credentials.idToken)
```
pmathew92, line 184: Same as https://github.com/auth0/Auth0.Android/pull/983/changes#r3433411514

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
```diff
@@ -775,9 +790,16 @@ public class SecureCredentialsManager @VisibleForTesting(otherwise = VisibleForT
...
-        return !(emptyCredentials || willExpire(
-            expiresAt, minTtl
-        ) && (canRefresh == null || !canRefresh))
+        if (emptyCredentials) {
+            return false
+        }
+        // IPSIE session_expiry: once the upstream-IdP ceiling passes, no valid credentials remain.
+        // The credentials blob is encrypted and cannot be decoded here, so the ceiling is read from
+        // the value persisted at login (KEY_SESSION_EXPIRY) via isSessionExpired(null).
+        if (isSessionExpired(null)) {
```
pmathew92, line 817: What is the intent here ? `isSessionExpired(null)` will always return false . So this check is completely redundant

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
utkrishtsahu: Yes — 30s aligns implementation ,the ~30s negative leeway. The leeway is intentionally negative so we treat the session as expired slightly before the wall-clock ceiling, never after.

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
utkrishtsahu: Intentional — "non-positive" means <= 0. We exclude 0 (epoch 1970) too, because it's not a valid live ceiling and storage.retrieveLong(KEY_SESSION_EXPIRY) returns 0 for the unset/migration case. Treating <= 0 as "no ceiling" mirrors the existing guard in willExpire, so we don't log out users who never had a ceiling persisted.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
utkrishtsahu: Done. saveCredentials here already had @Throws; updated the KDoc to document the SESSION_EXPIRED case as well.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
utkrishtsahu: Done. Added @Throws(CredentialsManagerException::class) and documented that saveCredentials can now throw SESSION_EXPIRED (when the ID token is already past its session_expiry ceiling at creation) and INVALID_CREDENTIALS.

auth0/src/main/java/com/auth0/android/authentication/storage/SecureCredentialsManager.kt
utkrishtsahu: It isn't always false. With a null idToken, isSessionExpired falls through to storage.retrieveLong(KEY_SESSION_EXPIRY) — the ceiling persisted at login. We pass null deliberately because the credentials blob is encrypted and can't be decoded in hasValidCredentials, so the persisted value is the only ceiling available here. Covered by the tests shouldFailWithSessionExpiredWithoutBiometricsOrRenewWhenStoredCeilingReached and ...WhenStoredCeilingFallsWithinLeeway, which assert this returns SESSION_EXPIRED from a stored ceiling.

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,67 @@ public abstract class BaseCredentialsManager internal constructor(
...
+    /**
+     * Persists the `session_expiry` ceiling read from the given ID token, if present.
+     *
+     * To preserve the ceiling across refreshes (the refresh grant does not re-emit `session_expiry`),
+     * the stored value is only ever written, never cleared, when a fresh ID token omits the claim.
+     * Call from `saveCredentials`.
+     */
+    protected fun persistSessionExpiry(idToken: String?) {
+        sessionExpiryFromIdToken(idToken)?.let { storage.store(KEY_SESSION_EXPIRY, it) }
+    }
```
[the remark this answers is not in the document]
utkrishtsahu: he realistic re-login path is already covered: clearCredentials() removes KEY_SESSION_EXPIRY on logout, so a subsequent fresh login that omits the claim doesn't inherit a stale ceiling. The write-only behavior is deliberate so that a refresh (whose ID token doesn't re-emit session_expiry) doesn't drop the ceiling mid-session. I'd prefer to keep the SL1 scope as-is here; happy to track a monotonic/earliest-wins refinement as a follow-up if you think the edge case (claimless fresh login without a preceding logout) is worth hardening.

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,75 @@ public abstract class BaseCredentialsManager internal constructor(
...
+     * with code `SESSION_EXPIRED` when it is, so an already-expired session is never persisted.
+     * No-op when the token is absent or does not carry both `session_expiry` and `iat`.
+     */
+    @Throws(CredentialsManagerException::class)
+    protected fun validateSessionExpiryAtCreation(idToken: String?) {
+        if (idToken.isNullOrBlank()) return
+        val jwt = runCatching { jwtDecoder.decode(idToken) }.getOrNull() ?: return
+        val sessionExpiry = jwt.sessionExpiry ?: return
+        val issuedAtSeconds = jwt.issuedAt?.time?.div(1000) ?: return
+        if (sessionExpiry <= issuedAtSeconds) {
```
kishore7snehil: This creation-time check uses a bare `sessionExpiry <= issuedAtSeconds`, but the read-time check at line 348 applies the 30s leeway. So a ceiling 1-30s past `iat` saves here, then gets rejected on the next read. Should we apply the same leeway in both places so they agree?

auth0/src/main/java/com/auth0/android/request/internal/Jwt.kt
```diff
@@ -53,6 +60,7 @@ internal class Jwt(rawToken: String) {
         authorizedParty = decodedPayload["azp"] as String?
         authenticationTime =
             (decodedPayload["auth_time"] as? Double)?.let { Date(it.toLong() * 1000) }
+        sessionExpiry = (decodedPayload["session_expiry"] as? Double)?.toLong()
```
kishore7snehil: `session_expiry` is customer-authored and expected in seconds, but there's no plausibility cap. A milliseconds value parses as a date ~50,000 years out and silently disables the ceiling - a fail-open with no signal. Could we reject implausibly large values (e.g. `>= 10_000_000_000`) and treat them as no ceiling? Minor: `as? Double` would drop a `Long`-typed value; `as? Number` parses both.

auth0/src/test/java/com/auth0/android/request/internal/JwtTest.kt
```diff
@@ -203,4 +204,53 @@ public class JwtTest {
...
+
+        assertThat(jwt.sessionExpiry, `is`(nullValue()))
+    }
+
+    @Test
+    public fun shouldTruncateFractionalSessionExpiryToLong() {
+        val jwt = Jwt(jwtWithPayload("""{"session_expiry":1700000000.75}"""))
+        assertThat(jwt, `is`(notNullValue()))
+
+        assertThat(jwt.sessionExpiry, `is`(1700000000L))
```
kishore7snehil, line 238: The one gap is an implausibly large (milliseconds) value - worth a test pinning that it's treated as no ceiling, since that's the silent fail-open mode for this control.

auth0/src/main/java/com/auth0/android/result/Credentials.kt
```diff
@@ -79,4 +79,18 @@ public data class Credentials(
...
+     * provider via the IPSIE `session_expiry` claim in the ID token, or `null` when the connection
+     * does not emit the claim.
+     *
+     * This is a session-level ceiling that is independent of [expiresAt] (the access-token expiry):
+     * it usually sits much further out and caps how long the local session may live, regardless of
+     * access-token renewals. A `null` value means there is no such ceiling.
+     *
+     * The value is decoded on demand from [idToken] and is not stored as a separate field.
+     */
+    public val sessionExpiresAt: Long?
```
kishore7snehil, line 102: This decodes the live `idToken` on each read, but the enforced ceiling is the one pinned at login and stored under `KEY_SESSION_EXPIRY`. After a refresh whose token omits the claim, this accessor returns null while the SDK is still enforcing a ceiling - so the public value diverges from what's actually enforced. Should it return the pinned value?

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,75 @@ public abstract class BaseCredentialsManager internal constructor(
...
+     * The ceiling is resolved in order: (1) the live claim in [idToken]; (2) the value persisted at
+     * login under [KEY_SESSION_EXPIRY] (so a refresh whose ID token omits the claim does not silently
+     * drop the ceiling); (3) if neither is present there is no ceiling and the session is NOT expired
+     * — a missing value must fall through to existing behavior, never be treated as already-expired.
+     *
+     * A small negative clock-skew leeway is applied so the session is treated as expired slightly
+     * before the wall-clock ceiling, never after.
+     */
+    protected fun isSessionExpired(idToken: String?): Boolean {
+        val sessionExpiry = sessionExpiryFromIdToken(idToken)
```
pmathew92: As per the spec, the expiry time we get during login is the final value and subsequent token refreshes shouldn't overwrite this initial value. This check , currently considers the id token which might contain a new value than the initially stored one thus giving a higher ceiling. So shouldn't the check be something like
```
val sessionExpiry = storage.retrieveLong(KEY_SESSION_EXPIRY)
      ?: sessionExpiryFromIdToken(idToken)
      ?: return false
```
@kishore7snehil  could you confirm this behaviour / condition ?

kishore7snehil (changes requested):

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
```diff
@@ -1812,6 +1813,20 @@ public class CredentialsManagerTest {
         Assert.assertFalse(manager.hasValidCredentials())
     }
 
+    @Test
```
pmathew92, line 1816: Add a test case for when the refreshed  ID token re-emits a later session_expiry and enforcement must still honor the pinned earlier value

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
utkrishtsahu: fixed. Both checks now use the same 30s leeway so they agree.

auth0/src/main/java/com/auth0/android/request/internal/Jwt.kt
utkrishtsahu: Done. Switched to as? Number and now treat values >= 10_000_000_000 (a ms value) as no ceiling instead of letting it silently disable enforcement.

auth0/src/test/java/com/auth0/android/request/internal/JwtTest.kt
utkrishtsahu: Added a test for this — pins that a milliseconds value parses as null.

auth0/src/main/java/com/auth0/android/result/Credentials.kt
utkrishtsahu: greed. The getter now returns the pinned value. Since Credentials doesn't have storage access, the managers stamp the pinned ceiling onto it before returning, and we fall back to the ID token only when nothing is pinned. Added tests for both cases.

auth0/src/test/java/com/auth0/android/authentication/storage/CredentialsManagerTest.kt
utkrishtsahu: Added a test for this — refreshed token carries a later expiry but enforcement still honors the earlier pinned value.

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
utkrishtsahu: Done, went with your suggestion. We now read the pinned value first and only fall back to the ID token claim, so a refresh can't bump the ceiling up.

EXAMPLES.md
```diff
@@ -2919,10 +2920,36 @@ when(credentialsManagerException) {
...
 ```
 
+### Upstream session expiry
+
+When an enterprise connection (for example an OIDC or Okta connection) is configured to assert a session lifetime, Auth0 includes a `session_expiry` claim in the ID token. This claim is an absolute ceiling — expressed in **Unix seconds** — on how long the local session may live, independently of the access-token expiry. It usually sits much further out than `expiresAt`, and it cannot be extended by a refresh-token renewal.
+
+The credentials managers enforce this ceiling automatically:
+
+- The ceiling is read from the ID token at login and persisted, so it survives refreshes whose ID token does not re-emit the claim.
+- On every `getCredentials` call, if the ceiling has been reached the stored credentials are cleared and the call fails with `CredentialsManagerException.SESSION_EXPIRED`. The refresh token is **never** used to renew a session past the ceiling.
```
kishore7snehil, line 3062: This section only talks about `getCredentials` failing once the ceiling is reached. But `saveCredentials` can also throw `SESSION_EXPIRED` - when the ID token is already past its ceiling at login. It's in the `saveCredentials` KDoc but not here, so someone reading just this guide would only handle it around `getCredentials` and get caught out when the save throws.

Could we add a line saying `saveCredentials` rejects an already-expired session up front too?

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -525,7 +551,7 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
                         fresh.scope
                     )
                     saveCredentials(credentials)
-                    callback.onSuccess(credentials)
+                    callback.onSuccess(stampPinnedSessionExpiry(credentials))
```
kishore7snehil, line 554: This stamping is what makes the returned credentials carry the pinned ceiling instead of the live id_token value - it's the actual fix for the `sessionExpiresAt` mismatch. But no test checks it end to end: a dropped `stampPinnedSessionExpiry(...)` here would still pass every test while the mismatch quietly came back.

Could we add a test that reads back a successful `getCredentials` and checks `sessionExpiresAt` is the pinned value?

kishore7snehil (changes requested):

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,95 @@ public abstract class BaseCredentialsManager internal constructor(
...
+     * treated as already-expired. The pinned value is read first because the ceiling is fixed at the
+     * initial login: a refresh whose ID token re-emits a *later* `session_expiry` must never raise it.
+     *
+     * A small negative clock-skew leeway is applied so the session is treated as expired slightly
+     * before the wall-clock ceiling, never after.
+     */
+    protected fun isSessionExpired(idToken: String?): Boolean {
+        val sessionExpiry = storage.retrieveLong(KEY_SESSION_EXPIRY)
+            ?: sessionExpiryFromIdToken(idToken)
+            ?: return false
```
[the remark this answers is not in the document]
utkrishtsahu: ixed. Non-positive values from both storage and the claim are now treated as "not pinned"/"no ceiling" via takeIf { it > 0 }, so a 0/negative stored sentinel falls through to the live claim instead of failing open.

EXAMPLES.md
utkrishtsahu: Added a bullet — saveCredentials rejects an already-expired session up front and throws SESSION_EXPIRED, so it's now documented in the guide and not just the KDoc.

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
utkrishtsahu: Added shouldReturnCredentialsCarryingPinnedSessionExpiresAt — it reads back a successful getCredentials and asserts sessionExpiresAt equals the pinned value, so dropping the stampPinnedSessionExpiry(...) call would now fail a test.

EXAMPLES.md
```diff
@@ -3040,10 +3041,37 @@ when(credentialsManagerException) {
...
+        // The stored credentials have already been cleared; prompt the user to
+        // re-authenticate.
+    }
     // ... similarly for other error codes
 }
 ```
 
+### Upstream session expiry
+
+When an enterprise connection (for example an OIDC or Okta connection) is configured to assert a session lifetime, Auth0 includes a `session_expiry` claim in the ID token. This claim is an absolute ceiling — expressed in **Unix seconds** — on how long the local session may live, independently of the access-token expiry. It usually sits much further out than `expiresAt`, and it cannot be extended by a refresh-token renewal.
```
kishore7snehil, line 3056: Could we add a short warning that the value must be Unix seconds, that an implausibly large (millisecond-magnitude) value is treated as "no ceiling", and that any malformed value fails open? 

It's a doc-only addition since the code already does the right thing. The other SDKs call this out prominently - see [auth0-server-python RetrievingData.md](https://github.com/auth0/auth0-server-python/blob/f29e406/examples/RetrievingData.md#session-expiry-from-the-upstream-idp) (dedicated warning: "a milliseconds value will silently disable enforcement... always divide by 1000") and [auth0-server-js EXAMPLES.md](https://github.com/auth0/auth0-auth-js/blob/cef72915050abe34a921f8fa1566774c76f94511/packages/auth0-server-js/EXAMPLES.md#session-expiry-from-upstream-idp-ipsie-session_expiry).

EXAMPLES.md
```diff
@@ -3040,10 +3041,37 @@ when(credentialsManagerException) {
...
+
+The credentials managers enforce this ceiling automatically:
+
+- The ceiling is read from the ID token at login and persisted, so it survives refreshes whose ID token does not re-emit the claim.
```
kishore7snehil, line 3060: The section explains what the SDK does once the claim is present, but not how a developer actually emits it. On a tenant the `session_expiry` claim is set by a Post-Login Action (the PR description uses `api.idToken.setCustomClaim('session_expiry', ...)` for testing) — a developer reading this doc has no pointer to that setup. server-python handles it with a short "Emitting the claim" note plus a placeholder for the canonical Auth0 Action guide: [RetrievingData.md](https://github.com/auth0/auth0-server-python/blob/f29e406/examples/RetrievingData.md#session-expiry-from-the-upstream-idp) ("will be adding the link to the session_expiry Action guide once published"). 

Could we add the same here?

kishore7snehil (changes requested):

EXAMPLES.md
utkrishtsahu: Added a warning callout — the value must be Unix seconds, an implausibly large (millisecond) value is treated as "no ceiling," and any malformed value fails open. Called out the divide-by-1000 guidance to match the other SDKs.

EXAMPLES.md
utkrishtsahu: Added an "Emitting the claim" subsection with the Post-Login Action snippet (api.idToken.setCustomClaim('session_expiry', ...) in seconds) and a placeholder for the canonical Action guide link, same as server-python.

kishore7snehil (approved):
LGTM!

auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
```diff
@@ -302,6 +315,93 @@ public abstract class BaseCredentialsManager internal constructor(
...
+     * session past it. [clearCredentials] removes the stored value on logout, so the next login re-pins
+     * a fresh ceiling. Call from `saveCredentials`.
+     */
+    protected fun persistSessionExpiry(idToken: String?) {
+        val incoming = sessionExpiryFromIdToken(idToken) ?: return
+        // A positive value is already pinned from the initial login -> keep it; ignore the claim
+        // re-emitted on this (refresh) grant. A null/non-positive stored value means nothing is pinned
+        // yet (mirrors the unset/migration guard in [isSessionExpired]), so stamp the ceiling now.
+        val pinned = storage.retrieveLong(KEY_SESSION_EXPIRY)
+        if (pinned != null && pinned > 0) return
```
pmathew92, line 380: Shouldn't this check be before the 
``` val incoming = sessionExpiryFromIdToken(idToken) ?: return```
It makes sense to decode the new session_expiry from the Id token, if there is no pinned value already present

pmathew92 (approved):

pmathew92 (approved):

PR #997, opened 2026-06-30
Guard against all uncaught exceptions in the serial executor block

pmathew92:
## Summary
Ports the serial-executor exception guard (#970) onto the `v4_development` branch as its own feature.

- Adds `runCatchingOnExecutor` helper to `BaseCredentialsManager`, wrapping each `serialExecutor.execute { }` body in a `catch (Throwable) -> onFailure(UNKNOWN_ERROR)` safety net so unexpected errors surface as a failure callback instead of being swallowed on the executor thread.
- Wraps all 3 `CredentialsManager` and all 3 `SecureCredentialsManager` execute blocks; removes the now-redundant trailing `catch (RuntimeException)` blocks while keeping the domain-specific catches (`AuthenticationException`, `CredentialsManagerException`, `IncompatibleDeviceException`, `CryptoException`).
- Brings `SecureCredentialsManager.continueGetApiCredentials` to parity with the other two blocks by mapping a save failure to `STORE_FAILED` (rather than letting it fall through to the new `UNKNOWN_ERROR` net).

This is the v4 port of the change merged to `main` in #970; reconciled onto v4's restructured credentials managers and test imports (`org.mockito.kotlin.*`).

## Test plan
- [x] `./gradlew :auth0:compileDebugKotlin`
- [x] `./gradlew :auth0:testDebugUnitTest` — full suite green
- [x] 5 new tests in `CredentialsManagerTest` (98 total), 4 new in `SecureCredentialsManagerTest` (131 total), 0 failures
- [x] Verified wrap/catch parity against #970 (3+3 execute blocks wrapped, STORE_FAILED 3=3, domain catches match)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential retrieval so unexpected errors are now handled consistently and returned as a standard error response instead of causing unhandled failures.
  * Applied the same safer error handling across sign-in, SSO, and API credential renewal flows.
  * Preserved existing behavior for authentication-specific errors while making runtime failures more predictable.

* **Tests**
  * Added coverage for unexpected storage and renewal failures to verify they surface the expected error response.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
@@ -129,48 +129,39 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
         callback: Callback<SSOCredentials, CredentialsManagerException>
     ) {
         serialExecutor.execute {
-            val refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN)
-            if (refreshToken.isNullOrEmpty()) {
-                callback.onFailure(CredentialsManagerException.NO_REFRESH_TOKEN)
-                return@execute
-            }
-
-            val tokenType = storage.retrieveString(KEY_TOKEN_TYPE)
-            validateDPoPState(tokenType)?.let { dpopError ->
-                callback.onFailure(dpopError)
-                return@execute
-            }
-
-            val request = authenticationClient.ssoExchange(refreshToken)
-            try {
-                if (parameters.isNotEmpty()) {
-                    request.addParameters(parameters)
+            runCatchingOnExecutor(callback) {
+                val refreshToken = storage.retrieveString(KEY_REFRESH_TOKEN)
+                if (refreshToken.isNullOrEmpty()) {
+                    callback.onFailure(CredentialsManagerException.NO_REFRESH_TOKEN)
+                    return@execute
                 }
-                val sessionTransferCredentials = request.execute()
-                saveSsoCredentials(sessionTransferCredentials)
-                callback.onSuccess(sessionTransferCredentials)
-            } catch (error: AuthenticationException) {
-                val exception = when {
-                    error.isNetworkError -> CredentialsManagerException.Code.NO_NETWORK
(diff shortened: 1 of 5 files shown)


NandanPrabhu:
@codeRabbitai full review

auth0/src/main/java/com/auth0/android/authentication/storage/CredentialsManager.kt
```diff
@@ -602,97 +581,99 @@ public class CredentialsManager @VisibleForTesting(otherwise = VisibleForTesting
...
-            val tokenType = apiCredentialType ?: storage.retrieveString(KEY_TOKEN_TYPE)
-            validateDPoPState(tokenType)?.let { dpopError ->
-                callback.onFailure(dpopError)
-                return@execute
-            }
+                val tokenType = apiCredentialType ?: storage.retrieveString(KEY_TOKEN_TYPE)
+                validateDPoPState(tokenType)?.let { dpopError ->
+                    callback.onFailure(dpopError)
+                    return@execute
+                }
```
utkrishtsahu, line 618: Wrapping the whole block means the fast-path callback.onSuccess(...) now runs inside the catch (Throwable). If a consumer's onSuccess throws, it gets caught and turned into a second onFailure(UNKNOWN_ERROR) — a double callback. On awaitCredentials, that's a resume after resume → IllegalStateException: Already resumed. Might be cleaner to call onSuccess outside the guarded block. Minor/edge-case, and it's inherited from #970

pmathew92: If the passed `onSuccess` throws , thats something the consumer should handle IMO. Most of the scenario, the onSuccess involves passing the fetched credentials to another API call which would have its own error catching mechanism. We can add this if a usecase on the same arises . Let me know your thoughts

utkrishtsahu: Makes sense — agreed the consumer should handle their own onSuccess. Fine to leave as-is since it matches #970. 👍

utkrishtsahu (approved):

PR #998, opened 2026-07-01
fix: MFAClient getAuthenticators filtering based on Authenticator.type field

utkrishtsahu:
### Changes

This PR contains two related MFA fixes.

#### 1. `getAuthenticators` — filter on `Authenticator.type` (exact match)

Updates how `MfaApiClient.getAuthenticators(factorsAllowed)` filters the authenticators returned by `GET /mfa/authenticators`.

**What changed**

- **Filtering logic** — Authenticators are now filtered by comparing each `Authenticator.type` value directly against the `factorsAllowed` list using exact (case-sensitive) equality. Previously the SDK derived an "effective type" by inspecting `oob_channel` and `authenticator_type` (with aliasing, e.g. `sms`/`phone`, `otp`/`totp`, and treating `oob` as a wildcard). That derivation and its helpers have been removed.
- **Methods removed** — the private helpers `matchesFactorType(...)` and `getEffectiveType(...)` in `MfaApiClient`.
- **No endpoint or public method signature changes** — `getAuthenticators(factorsAllowed: List<String>)` keeps the same signature; only its internal filtering behavior changed.
- **Documentation** — `MfaApiClient` KDoc and `EXAMPLES.md` updated so the `factorsAllowed` examples use the same factor values as the `type` field (e.g. `phone`, `email`, `otp`, `recovery-code`) instead of channel-level values like `sms`/`oob`.

**Why it matters**

The `factorsAllowed` values now map one-to-one to the `Authenticator.type` field, so filtering is predictable and matches the values a caller gets from `MfaRequirements` (`MfaFactor.type`). This is a **behavioral change**: callers passing channel-level or wildcard values such as `sms` or `oob` will no longer match — they should pass the resolved `type` value (e.g. `phone`). Confirmed against live FGS that a TOTP factor has `type: "totp"` (`authenticator_type: "otp"`), so callers must pass `totp` to surface it.

#### 2. `TotpEnrollmentChallenge` — correct field mapping for `/mfa/associate`

`TotpEnrollmentChallenge` is shared by two endpoints with different response shapes, but was modeled only on the My Account `/authentication-methods` shape. Used against the ROPG `/mfa/associate` endpoint (which `MfaApiClient.enroll(Otp)` calls), this caused:

- `id` and `authSession` declared non-null but populated as `null` by Gson (null-safety bypass / latent NPE hazard),
- `manualInputCode` always `null` (the endpoint returns the key as `secret`, not `manual_input_code`),
- `secret` and `recovery_codes` silently dropped.

**What changed**

- Made `id` and `authSession` nullable (`manualInputCode` was already nullable) and added `authenticatorType`, `secret`, `recoveryCodes`, so a single class deserializes both endpoints correctly. This matches Auth0.swift's `OTPMFAEnrollmentChallenge` (the only cross-SDK peer that implements `/mfa/associate`).
- **My Account flow unaffected** — its fields (`id`/`authSession`/`manualInputCode`/`barcodeUri`) still map identically; the change only loosens nullability and adds fields. No runtime break for existing users; recompiling My Account consumers may need a null check on `id`/`authSession` (the safe direction).
- **Documentation** — `EXAMPLES.md` MFA `enroll(Otp)` snippet corrected to read `secret` (not `manualInputCode`) and now shows `recoveryCodes`.

**Usage**

```kotlin
// factorsAllowed values map directly to Authenticator.type
val factorTypes = requirements?.challenge?.map { it.type } ?: emptyList()

mfaClient
    .getAuthenticators(factorsAllowed = factorTypes) // e.g. ["otp", "phone", "email", "recovery-code"]
    .start(object : Callback<List<Authenticator>, MfaListAuthenticatorsException> {
        override fun onSuccess(result: List<Authenticator>) {
            // Only authenticators whose `type` matches one of factorsAllowed are returned
        }
        override fun onFailure(error: MfaListAuthenticatorsException) { }
    })

// Enrolling TOTP via /mfa/associate
mfaClient
    .enroll(MfaEnrollmentType.Otp)
    .start(object : Callback<EnrollmentChallenge, MfaEnrollmentException> {
        override fun onSuccess(enrollment: EnrollmentChallenge) {
            if (enrollment is TotpEnrollmentChallenge) {
                val secret = enrollment.secret          // manual-entry key
                val barcodeUri = enrollment.barcodeUri  // otpauth:// URI for QR
                val recoveryCodes = enrollment.recoveryCodes
            }
        }
        override fun onFailure(error: MfaEnrollmentException) { }
    })
```

### Testing

- **getAuthenticators** unit tests in `MfaApiClientTest` exercise exact-type matching: fixtures include a top-level `type` field and assert only authenticators whose `type` is in `factorsAllowed` are returned.
- **TOTP enroll** test updated to use the real `/mfa/associate` response shape, asserting `secret`/`recoveryCodes`/`authenticatorType` populate.
- `MyAccountAPIClientTest` re-run to confirm the shared-class change does not affect the My Account TOTP/push flows.
- Verified end-to-end on a physical device (Pixel 7a) against live FGS: login → enroll OTP → verify OTP → verify recovery code, all succeeded; `secret`/`recovery_codes` populate correctly.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MFA authenticator filtering to use the exact factor type you request, resulting in more consistent authenticator lists and smoother enrollment.
  * Updated TOTP enrollment handling to correctly capture and expose the latest response fields (including secret and recovery codes), improving setup details.
* **Documentation**
  * Refreshed MFA Kotlin/Java examples to use the correct factor names and updated TOTP enrollment fields to match current responses.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/mfa/MfaApiClient.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/mfa/MfaApiClient.kt
@@ -124,16 +124,19 @@ public class MfaApiClient @VisibleForTesting(otherwise = VisibleForTesting.PRIVA
      * ## Usage
      *
      * ```kotlin
-     * mfaClient.getAuthenticators(listOf("otp", "oob"))
+     * mfaClient.getAuthenticators(listOf("otp", "phone"))
      *     .start(object : Callback<List<Authenticator>, MfaListAuthenticatorsException> {
      *         override fun onSuccess(result: List<Authenticator>) {
-     *             // Only OTP and OOB authenticators returned
+     *             // Only authenticators whose `type` is "otp" or "phone" are returned
      *         }
      *         override fun onFailure(error: MfaListAuthenticatorsException) { }
      *     })
      * ```
      *
-     * @param factorsAllowed Array of factor types to filter the authenticators (e.g., `["otp", "oob", "recovery-code"]`).
+     * Filtering matches each authenticator's [Authenticator.type] field against the
+     * provided values using exact (case-sensitive) equality.
+     *
+     * @param factorsAllowed Array of factor types to filter the authenticators (e.g., `["otp", "phone", "email", "recovery-code"]`).
      *                       Must contain at least one factor type.
      * @return a request to configure and start that will yield a list of [Authenticator]
      *
@@ -307,11 +310,11 @@ public class MfaApiClient @VisibleForTesting(otherwise = VisibleForTesting.PRIVA
      * transparently by the SDK.
      *
      * **Filtering:**
-     * Authenticators are filtered by their effective type:
-     * - OOB authenticators: matched by their channel ("sms" or "email")
-     * - Other authenticators: matched by their type ("otp", "recovery-code", etc.)
+     * An authenticator is included when [Authenticator.type] exactly matches one of the
+     * provided [factorsAllowed] values (case-sensitive equality). This mirrors the
+     * filtering behavior of the Auth0.swift SDK.
      *
-     * @param factorsAllowed List of factor types to include (e.g., ["sms", "email", "otp"])
+     * @param factorsAllowed List of factor types to include (e.g., ["phone", "email", "otp"])
      * @return A JsonAdapter that produces a filtered list of authenticators
      */
     private fun createFilteringAuthenticatorsAdapter(factorsAllowed: List<String>): JsonAdapter<List<Authenticator>> {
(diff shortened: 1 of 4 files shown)


auth0/src/main/java/com/auth0/android/authentication/mfa/MfaApiClient.kt
```diff
@@ -124,16 +124,19 @@ public class MfaApiClient @VisibleForTesting(otherwise = VisibleForTesting.PRIVA
      * ## Usage
      *
      * ```kotlin
-     * mfaClient.getAuthenticators(listOf("otp", "oob"))
+     * mfaClient.getAuthenticators(listOf("otp", "phone"))
```
NandanPrabhu, line 127: @utkrishtsahu plz check output of getauthenticators once I guess its totp and not otp thats returned. I might be wrong but please check

NandanPrabhu (dismissed):

pmathew92 (dismissed):

auth0/src/main/java/com/auth0/android/result/EnrollmentChallenge.kt
```diff
@@ -55,15 +55,36 @@ public data class OobEnrollmentChallenge(
...
+ * authenticator app; only one is populated depending on the endpoint. The [barcodeUri]
+ * (`otpauth://` URI) is always present and embeds the same secret.
+ */
 public data class TotpEnrollmentChallenge(
     @SerializedName("id")
-    override val id: String,
+    override val id: String? = null,
     @SerializedName("auth_session")
-    override val authSession: String,
+    override val authSession: String? = null,
```
pmathew92, line 94: Document these nullable property change

NandanPrabhu (dismissed):

auth0/src/main/java/com/auth0/android/authentication/mfa/MfaApiClient.kt
```diff
@@ -124,16 +124,19 @@ public class MfaApiClient @VisibleForTesting(otherwise = VisibleForTesting.PRIVA
...
      *         override fun onFailure(error: MfaListAuthenticatorsException) { }
      *     })
      * ```
      *
-     * @param factorsAllowed Array of factor types to filter the authenticators (e.g., `["otp", "oob", "recovery-code"]`).
+     * Filtering matches each authenticator's [Authenticator.type] field against the
+     * provided values using exact (case-sensitive) equality.
+     *
+     * @param factorsAllowed Array of factor types to filter the authenticators (e.g., `["otp", "phone", "email", "recovery-code"]`).
      *                       Must contain at least one factor type.
```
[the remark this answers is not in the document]
utkrishtsahu: his filters on the high-level type field, not authenticator_type. Confirmed against a live GET /mfa/authenticators response from an FGS tenant: an email authenticator returns "type": "email", "authenticator_type": "oob", "oob_channel": "email". The SDK matches Authenticator.type, so phone/email are the correct filter values — oob (the low-level authenticator_type) would match nothing. These are also the same values returned by MfaRequirements.MfaFactor.type, so callers can pass them through directly.

auth0/src/main/java/com/auth0/android/result/EnrollmentChallenge.kt
utkrishtsahu: added.

pmathew92 (approved):

PR #1031, opened 2026-07-23
fix: guard against fatal double-resume crash in CredentialsManager

pmathew92:
## Why

Customers report a fatal, uncaught `IllegalStateException: "Already resumed"` originating in the credentials managers 

### Root cause

An executor block can invoke a terminal callback (`onSuccess`) and then have a `Throwable` escape *after* it — e.g. in `continueGetCredentials`, `callback.onSuccess(...)` fires, and the refresh-branch inner `catch` only handles `AuthenticationException`, so any other `Throwable` propagates up to `runCatchingOnExecutor`, which then calls `callback.onFailure(...)`. That is **two terminal callbacks** for one request.

Through the `suspendCancellableCoroutine` bridges (`awaitCredentials`/`awaitSsoCredentials`/`awaitApiCredentials`), the first callback resumes the continuation and the second resumes it again. A second resume throws `"Already resumed"` on the single-threaded serial executor, where nothing catches it → process death.

## Fix (defense in depth)

**Layer A — root cause (`BaseCredentialsManager`)**
`runCatchingOnExecutor` now wraps the callback in a new `SingleShotCallback` (backed by `AtomicBoolean`), so any terminal callback after the first is silently dropped — upholding the fire-once contract of `Callback` even when a block both invokes the callback and later throws. Fatal errors (`VirtualMachineError`/`ThreadDeath`/`LinkageError`) are still rethrown. All call sites shadow the block's callback parameter so existing inner `callback.xxx` calls route through the guard with no body edits.

**Layer B — bridge safety net (`CredentialsManager`, `SecureCredentialsManager`)**
Every coroutine bridge now guards its resume with `if (continuation.isActive)`, so even a double callback reaching a bridge directly can never resume the continuation twice.

## Tests

Added regression tests to both `CredentialsManagerTest` and `SecureCredentialsManagerTest`:

- `shouldNotInvokeCallbackTwiceWhenConsumerThrowsAfterSuccessOnGetCredentials` — covers **Layer A** via the callback API.
- `shouldNotResumeContinuationTwiceWhenCallbackFiresTwiceOnAwaitCredentials` — covers **Layer B** via the coroutine/`await` API.

Each test was verified to **fail** with the respective guard removed (reproducing `IllegalStateException: Already resumed`) and **pass** with the fix.

### DIFF
--- a/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
+++ b/auth0/src/main/java/com/auth0/android/authentication/storage/BaseCredentialsManager.kt
@@ -11,7 +11,7 @@ import com.auth0.android.result.Credentials
 import com.auth0.android.result.SSOCredentials
 import com.auth0.android.result.UserProfile
 import com.auth0.android.util.Clock
-import java.util.*
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Base class meant to abstract common logic across Credentials Manager implementations.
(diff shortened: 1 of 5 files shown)


utkrishtsahu (approved):
LGTM,no remarks.

